PR #2, opened 2019-07-17
feat: added utils for running probot apps in gcf

orthros:
Work in progress. Tests are still failing.

Needs License header (if only there was a bot for this! 🤔)


orthros:
cc @bcoe

packages/gcf-utils/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "name": "gcf-utils",
+  "version": "0.1.0",
+  "description": "An extension for running Probot in Google Cloud Functions",
+  "scripts": {
+    "compile": "tsc -p .",
+    "prepare": "npm run compile",
+    "pretest": "npm run compile",
+    "test": "cross-env LOG_LEVEL=fatal mocha --recursive ./build/test",
+    "fix": "gts check",
+    "check": "gts fix"
```
bcoe: we switched this to `lint` from `check` in the top level.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -0,0 +1,113 @@
...
+import { Storage } from '@google-cloud/storage';
+import * as KMS from '@google-cloud/kms';
+import { readFileSync } from 'fs';
+import * as express from 'express';
+
+
+export class GCFBootstrapper {
+  probot?: Probot;
+
+  async loadProbot(appFn: ApplicationFunction): Promise<Probot> {
```
bcoe: how are you picturing applications will bootstrap, does it make sense to expose this as a bin, so that ultimately we just having something like:

```json
{
  "scripts": {
    "prod": "cf-probot ./build/my-bot.js"
  }
}
```

orthros: I would imagine it would be something similar to the following:

``` typescript
import { GCFBootstrapper } from '../gcf-utils';
import * as appFn from './';

bootstrap = new GCFBootstrapper();
module.exports = bootstrap.gcf(appFn);
```

With that being the "main" that is loaded as a Cloud Function.

It is fairly boilerplate, but I think holds in convention with other Probot apps && will allow us to write our "core" Probot apps cross-functional if need be.

What do you think?

bcoe: this seems reasonable to me, I think we can probably just add a `bin/run.js` for our bots, which the boiler plate you shared, and then make `npm run` call this 👍

PR #16, opened 2019-08-01
test: adds integration test for gcf-utils and a local development opt…

orthros:
…ion for conventional-commit-lint


packages/conventional-commit-lint/src/app.ts
```diff
@@ -18,4 +18,9 @@ import { GCFBootstrapper } from 'gcf-utils';
 import appFn from './conventional-commit-lint';
 
 const bootstrap = new GCFBootstrapper();
-module.exports.conventional_commit_lint = bootstrap.gcf(appFn);
+async function run() {
+    const opts = await bootstrap.getProbotConfig();
+    console.info(opts.cert);
```
bcoe: we can probably revert this bit, will be curious to see if the underlying bug crops up again.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -93,6 +93,7 @@ export class GCFBootstrapper {
       // Do the thing
       if (name) {
         try {
+          console.log(request.body);
```
bcoe: can probably nuke this console.log.

packages/conventional-commit-lint/package.json
```diff
@@ -30,23 +31,28 @@
     "@commitlint/config-conventional": "^8.0.0",
     "@commitlint/lint": "^8.0.0",
     "@octokit/rest": "^16.28.3",
-    "gcf-utils": "^0.4.1",
-    "probot": "^9.2.15"
+    "gcf-utils": "^0.4.0",
+    "probot": "^9.2.20"
   },
   "devDependencies": {
+    "@types/body-parser": "^1.17.0",
```
bcoe: I think this should be in `dependencies` not `devDependencies`.

orthros:
No problems on the nits. I just wanted to get exactly what we had (no matter how rough) into a branch so we can get more eyes on it if need be.

packages/conventional-commit-lint/package.json
orthros: I left this in the devDependencies as (in theory) one should only be using body-parser when running locally.

bcoe: oh, fair my bad; `@types` should usually be dev dependencies.

packages/gcf-utils/package.json
```diff
@@ -6,7 +6,8 @@
     "compile": "tsc -p .",
     "prepare": "npm run compile",
     "pretest": "npm run compile",
-    "test": "cross-env LOG_LEVEL=fatal mocha --recursive ./build/test",
+    "test": "cross-env LOG_LEVEL=fatal mocha ./build/test",
+    "test:integration": "npm run pretest && cross-env LOG_LEVEL=fatal mocha ./build/test/integration",
```
JustinBeckwith: Can we use `system-test` to be more consistent with other modules?

packages/gcf-utils/package.json
```diff
@@ -53,4 +54,4 @@
     "package.json",
     "LICENSE"
   ]
-}
+}
```
JustinBeckwith: Let's make the angry red squiggle go away

packages/conventional-commit-lint/src/local.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { GCFBootstrapper } from 'gcf-utils';
+import appFn from './conventional-commit-lint';
+import express from 'express';
+import { config } from 'dotenv';
+import { resolve } from 'path';
+import * as bodyParser from 'body-parser';
+
+const out = config({ path: resolve(__dirname, '../../.env') });
+console.log(out);
```
JustinBeckwith: Will this env config have keys in it that shouldn't be output to logs?

packages/gcf-utils/src/bin/genkey.ts
```diff
@@ -25,127 +25,133 @@ import { Base64 } from 'js-base64';
...
 const botname: string = argv.bot as string;
 const webhookSecret: string = argv.secret as string;
 const id: number = argv.id as number;
 
-
 if (!project) {
-    console.error('Project name is required');
-    yargs.showHelp();
-    process.exit(1);
+  console.error('Project name is required');
```
JustinBeckwith: Pretty sure there's a way to make `yargs` do this kind of validation for us, right @bcoe ?

packages/gcf-utils/src/bin/genkey.ts
```diff
@@ -25,127 +25,133 @@ import { Base64 } from 'js-base64';
...
+    const kmsclient = new KMS.KeyManagementServiceClient(opts);
+
+    const name = kmsclient.cryptoKeyPath(project, location, keyring, botname);
+
+    const plaintext = Base64.encode(JSON.stringify(blob));
+    const [kmsresult] = await kmsclient.encrypt({ name, plaintext });
+    encblob = kmsresult.ciphertext;
+  } catch (e) {
+    console.error('Got an error encrypting the blob');
+    console.error(e);
```
JustinBeckwith: Instead of this, can we just `throw e`?  @bcoe

packages/gcf-utils/test/integration/gcf-utils-integration.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  describe('loadProbot', () => {
+    let bootstrapper: GCFBootstrapper;
+
+    beforeEach(async () => {
+      bootstrapper = new GCFBootstrapper();
+      config({ path: resolve(__dirname, '../../.env') });
+    });
+
+    afterEach(() => {});
```
JustinBeckwith: Don't think we need this

packages/gcf-utils/test/integration/gcf-utils-integration.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    it('is called properly', async () => {
+      const pb = await bootstrapper.loadProbot((app: Application) => {
+        app.on('foo', async context => {
+          console.log('We are called!');
+        });
+      });
+      //console.log(pb);
```
JustinBeckwith: lets remove that

packages/conventional-commit-lint/src/local.ts
orthros: The .env in this case just has environment variables like GCP Project name, Bucket Name, Function Name etc.

```
DRIFT_PRO_BUCKET=YOUR_BUCKET_NAME
GCF_SHORT_FUNCTION_NAME=YOUR_FUNCTION_NAME
PROJECT_ID=YOUR_PROJECT_NAME
KEY_LOCATION=global
KEY_RING=YOUR_KEY_NAME
```

packages/gcf-utils/package.json
orthros: I actually don't see an angry red squiggle anywhere 😬

packages/gcf-utils/src/bin/genkey.ts
bcoe: @orthros in this case, I tend to agree, you could probably just drop the `try`, `catch`, and let the exception bubble to the terminal.

Where I will sometimes capture errors, is for user facing CLI applications, where I potentially want to obfuscate the internal implementation (do they even need to know it's Node.js). In this case, I'd just let it throw.

bcoe:
@orthros bother you to rebase and we can merge this, if there's still some worthwhile testing code you'd like to keep?

packages/gcf-utils/src/bin/genkey.ts
orthros: Done 👍

PR #18, opened 2019-08-01
feat(gcf-utils): Utility to add/upload probot secrets.

orthros:
Added new bin: `genkey`, which will build and encrypt a set of
configuration options and upload them to a GCS bucket.

### DIFF
--- a/packages/gcf-utils/src/bin/genkey.ts
+++ b/packages/gcf-utils/src/bin/genkey.ts
@@ -17,55 +17,84 @@
 
 import fs from 'fs';
 import path from 'path';
-import commander from 'commander';
+import * as yargs from 'yargs';
+import { Argv } from 'yargs';
 import * as KMS from '@google-cloud/kms';
 import { Storage, StorageOptions } from '@google-cloud/storage';
 import { Base64 } from 'js-base64';
 import { Options } from 'probot';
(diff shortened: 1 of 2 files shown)


packages/gcf-utils/src/bin/genkey.ts
```diff
@@ -0,0 +1,122 @@
...
+import fs from 'fs';
+import path from 'path';
+import commander from 'commander';
+import * as KMS from '@google-cloud/kms';
+import { Storage, StorageOptions } from '@google-cloud/storage';
+import { Base64 } from 'js-base64';
+import { Options } from 'probot';
+import * as tmp from 'tmp';
+
+commander
```
bcoe: I believe in developers making their own design decisions ... but since I wrote yargs, there's a pretty good chance I eventually swap it in for commander :stuck_out_tongue:

orthros: WHOOPS! 😅 Maybe I'll refactor tomorrow morning

bcoe: @orthros I mean, one less repo to generate bug reports for me from; don't feel you have to.

orthros: Too late!

PR #25, opened 2019-08-11
feat: add label-sync bot

### LINKED ISSUE #52: New Bot: Label sync
Provide a common set of labels (name and color) that can be enforced in all repositories across an organization.

JustinBeckwith:
Fixes #52


packages/label-sync/src/app.ts
```diff
@@ -0,0 +1,142 @@
...
+    await reconcileLabels(owner, repo, context.github);
+  });
+
+  // On startup, synchronize labels for all relevant repos
+  const reposResponse = await request<RepoResponse>({
+    url: reposUrl
+  });
+  await Promise.all(reposResponse.data.repos.map(r => {
+    const [owner, repo] = r.repo.split('/');
+    return reconcileLabels(owner, repo, {} as any);
```
JustinBeckwith: When this bot wakes up for the first time, I'd like to iterate over all repos and run this little reconciler thingy.  After that, we'll just look at net new repos.  When someone modifies the `labels.json`, my assumption is that this would get re-deployed, and restart the server, running this thingy again.  Here's the only problem.  I don't seem to have access to the `context.github` object with the preloaded `octokit` instance like I have on event hooks.  

Do y'all have any ideas on the "right" way to do this?
@orthros @bcoe

packages/label-sync/src/label-sync.ts
```diff
@@ -0,0 +1,130 @@
...
+    language: string;
+    repo: string;
+  }[]
+}
+
+export = (app: Application) => {
+  // Sync labels when a new repo is created
+  app.on('schedule.repository', async context => {
+    // this needs logic to list all the repos we
+    // have label sync installed on:
```
bcoe: `schedule.repository` will be called on whatever schedule we wish, with an authenticated `context.github` instance. We should update the bot logic to pull a list of repos that need to have their labels fixed (we probably want to do this on a fairly infrequent schedule, e.g., daily).

bcoe:
@JustinBeckwith is this ready to go out the door, once linting passes?

packages/label-sync/src/label-sync.ts
```diff
@@ -0,0 +1,137 @@
...
+interface Repos {
+  repos: [{
+    language: string,
+    repo: string
+  }]
+}
+
+export = (app: Application) => {
+  const events = [
+    'repository.created',
```
bcoe: this seems reasonable to me 👍 

So if I edit labels they immediately go back to where they started.

The one use-case we don't cover, is how we periodically sync labels as new ones are added.

JustinBeckwith: We don't really need periodic :) Below we trigger a full on update whenever there are pushes to the repo-automation bots repo (with a todo to make it more scoped).

packages/label-sync/src/label-sync.ts
```diff
@@ -0,0 +1,176 @@
...
+    'repository.transferred',
+    'label.edited',
+    'label.deleted',
+  ];
+
+  events.forEach(evt =>
+    app.on(evt, async c => {
```
bcoe: you can provide an array to `events` rather than calling `app.on` multiple times I believe.

JustinBeckwith:
aiight @bcoe & @chingor13 - this is actually ready now :) Happy to work with y'all to get it deployed next week.

PR #29, opened 2019-08-14
feat: add release-please bot

### LINKED ISSUE #54: New Bot: Propose/start a release
When there are changes to master, propose a new release.

chingor13:
Fixes #54


packages/release-please/src/release-please.ts
```diff
@@ -0,0 +1,47 @@
...
+    const repoUrl = context.payload.repository.full_name;
+    const branch = context.payload.ref.replace('refs/heads/', '')
+    if (branch != PRIMARY_BRANCH) {
+      app.log.info(`Not on primary branch (${PRIMARY_BRANCH}): ${branch}`);
+      return;
+    }
+    const packageName = context.payload.repository.name;
+    const token = 'FIXME'; // somehow get the token
+
+    const rp = new ReleasePR({
```
bcoe: I split this out into a few classes now:

https://github.com/googleapis/release-please/blob/master/src/bin/release-please.ts#L80

we'll probably want to do some sort of similar factory pattern.

packages/release-please/README.md
```diff
@@ -0,0 +1,43 @@
+# release-please
```
JustinBeckwith: If we are going to keep the README, let's make sure it's relevant :)

packages/release-please/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "release-please",
```
JustinBeckwith: That's the name of the actual `release-please` npm module, I think.  We should make this `release-please-bot` or something like that to avoid issues.

packages/release-please/src/release-please.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Application } from 'probot';
+import { GitHubAPI } from 'probot/lib/github';
+import { ReleaseType } from 'release-please/build/src/release-pr';
```
JustinBeckwith: Reaching down into non-index exported classes like this is generally something we want to avoid.  We should make sure any methods or classes we want to use are exported from the root @bcoe

packages/release-please/src/release-please.ts
```diff
@@ -0,0 +1,97 @@
...
+import { JavaYoshi } from 'release-please/build/src/releasers/java-yoshi';
+
+interface ConfigurationOptions {
+  strategy?: string;
+  primaryBranch: string;
+  releaseLabels: string[];
+}
+
+const DEFAULT_API_URL = 'https://api.github.com';
+const WELL_KNOWN_CONFIGURATION_FILE = '.github/release-please.json';
```
JustinBeckwith: I notice that here we're using `json`, but in the blunderbuss bot we're using `yaml`.  Can we pick one and be consistent?

packages/release-please/src/release-please.ts
```diff
@@ -0,0 +1,97 @@
...
+    );
+
+    if (branch !== configuration.primaryBranch) {
+      app.log.info(
+        `Not on primary branch (${configuration.primaryBranch}): ${branch}`
+      );
+      return;
+    }
+
+    const rp = new JavaYoshi({
```
JustinBeckwith: It would be nice if `release-please` exported something generic that we could use here for *all* the languages we support @bcoe

packages/release-please/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "gts/tsconfig-google",
+  "compilerOptions": {
+    "esModuleInterop": true,
```
JustinBeckwith: It's weird to me that we turned this on for our bots, but don't use it with the rest of our modules.  Can we consider dropping this?

packages/release-please/src/release-please.ts
chingor13: We just created a factory in release-please 2.7.0 exactly to support this. This PR is still a WIP.

packages/release-please/src/release-please.ts
chingor13: Filed https://github.com/googleapis/release-please/issues/249 and added a TODO once the types are exposed.

packages/release-please/tsconfig.json
chingor13: Per the offline discussion, this seems to be necessary as one of our dependencies requires it? @bcoe

packages/release-please/src/release-please.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const DEFAULT_API_URL = 'https://api.github.com';
+const WELL_KNOWN_CONFIGURATION_FILE = 'release-please.yml';
+const DEFAULT_CONFIGURATION: ConfigurationOptions = {
+  primaryBranch: 'master',
+  releaseLabels: ['autorelease: pending', 'type: process'],
+};
+
+function releaseTypeFromRepoLanguage(language: string | null): ReleaseType {
+  switch (language) {
```
bcoe: I would be tempted to `switch(language.toLowerCase())` rather than the multiple switches.

packages/release-please/src/release-please.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    const releaseType = configuration.releaseType
+      ? configuration.releaseType
+      : releaseTypeFromRepoLanguage(context.payload.repository.language);
+    const buildOptions: BuildOptions = {
+      packageName: repoName,
+      repoUrl,
+      apiUrl: DEFAULT_API_URL,
+      octokitAPIs: {
+        octokit: context.github,
```
bcoe: I'm itching to see if this works as expected once installed.

packages/release-please/tsconfig.json
bcoe: let's check how much trouble it creates removing the setting, and if it's a big refactor, create a tracking ticket to stop using the interop?

I agreed with @JustinBeckwith's reasoning, but might be worth fixing in all bots in a future PR.

chingor13: Filed https://github.com/googleapis/repo-automation-bots/issues/62 to track esModuleInterop

packages/release-please/src/release-please.ts
chingor13: done.

PR #33, opened 2019-08-15
fix: match copyright line in comments only

### LINKED ISSUE #32: License header checker should only look in headers
Example: https://github.com/googleapis/gcs-resumable-upload/pull/256

This file has a string that matches the copyright header regex and is resulting in an incorrect failure.

chingor13:
Fixes #32


packages/header-checker-lint/__snapshots__/header-checker-lint.js
```diff
@@ -1,19 +1,3 @@
-/**
```
chingor13: Looks like snap-shot-it strips this out.

bcoe: Yeah, I'd noticed this, any thoughts? is there a heuristic way we could skip `__snapshot__` files, while still catching files that should have headers; it's been helpful to me catching a few legacy files that were simply missing headers.

chingor13: I'm looking into the configuration files in #34

bcoe:
@chingor13 mind resolving the conflicts when you have a moment?

PR #34, opened 2019-08-15
feat: add header-checker-lint options file

chingor13:
The bot now reads a `.bots/header-checker-lint.json` file for configuration. If available, it will set the override the default configuration with values from this json file. If not available, it will use the default configuration.

### DIFF
--- a/packages/header-checker-lint/README.md
+++ b/packages/header-checker-lint/README.md
@@ -15,10 +15,10 @@ with the following options:
 
 | Name | Description | Type | Default |
 |----- | ----------- | ---- | ------- |
-| `allowed_copyright_holders` | List of allowed copyright holders | `string[]` | `[]` |
-| `allowed_license` | The allowed license. This is **not** currently validated. | "Apache-2.0", "MIT | any |
-| `source_file_extensions` | List of file extensions that will be treated as source files | `string[]` | `["ts", "js", "java"]` |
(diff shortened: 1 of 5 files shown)


packages/header-checker-lint/README.md
```diff
@@ -1,9 +1,28 @@
...
 
-## Setup
+## Installation
+
+You can install this bot [from GitHub][github-app-link].
+
+### Configuration
+
+To configure the bot, you can create a configuration file:
+`.bots/header-checker-lint.json`. The contents of this file define a JSON object
```
bcoe: clever, I like the `.bots/` folder, and then each config just maps to a bot?

packages/header-checker-lint/src/header-checker-lint.ts
```diff
@@ -39,15 +40,72 @@ interface LicenseHeader {
...
+  }
+
+  static async fromGitHub(
+    path: string,
+    owner: string,
+    repo: string,
+    ref: string,
+    github: GitHubAPI
+  ): Promise<Configuration> {
+    return github.repos
```
bcoe: no need for the promise chain here, in an async function, you can just:

```
try {
await response = github.repos.getContents();
} catch () {}
```

chingor13: done

packages/header-checker-lint/README.md
chingor13: Yeah, this layout is totally up for discussion, though.

PR #42, opened 2019-08-20
fix: loosen copyright regex

bcoe:
allow `// Copyright`, allow `Google LLC.`, this addresses a bunch of alerts I've seen.

### DIFF
--- a/packages/header-checker-lint/src/header-checker-lint.ts
+++ b/packages/header-checker-lint/src/header-checker-lint.ts
@@ -114,7 +114,7 @@ class Configuration {
   }
 }
 
-const COPYRIGHT_REGEX = new RegExp('\\s*[\\*#]\\s*Copyright (\\d{4}) (.*)$');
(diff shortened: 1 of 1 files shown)


packages/header-checker-lint/src/header-checker-lint.ts
```diff
@@ -114,7 +114,7 @@ class Configuration {
   }
 }
 
-const COPYRIGHT_REGEX = new RegExp('\\s*[\\*#]\\s*Copyright (\\d{4}) (.*)$');
+const COPYRIGHT_REGEX = /\s*[*#/]\/? \s*Copyright (\d{4}) ([\w\s]+).?$/
```
chingor13: The last dot should probably be escaped and perhaps the `[*#/]\/?` should become `(\*|#|\/\/)`.

We should probably add some helper unit tests for the license detection so we don't need to create a bunch of fixtures for testing header values.

bcoe: i was thinking the same thing, figured I'd forgo for this one patch, but it would be nice to be able to write unit tests for specific headers.

PR #47, opened 2019-08-23
fix: blunderbuss ignores missing configs

kurtisvg:


packages/blunderbuss/test/blunderbuss.ts
```diff
@@ -271,7 +271,11 @@ describe('Blunderbuss', () => {
         .get('/repos/testOwner/testRepo/contents/.github/blunderbuss.yml')
         .reply(200, { content: config });
 
-      await probot.receive({ name: 'issues.labeled', payload, id: 'abc123' });
+      await probot.receive({
```
bcoe: I might be tempted to add a unit tests that returns a 404 rather than 200 for fetching the content of a config file, I think this would allow us to confirm that missing config behaves as expected.

kurtisvg: This was how I originally had the tests set up, but it wasn't working. I spent a little more time on it, and it looks like ocotkit's `repos.getContents()` makes a second attempt at a different endpoint (for some reason).

I ended up just adding a second stub and the tests work correctly now.

PR #76, opened 2019-09-17
fix(build): add subcommand to get proxyurl

orthros:


orthros:
@chingor13 I believe this will fix the build

Issue #87, opened 2019-09-19
feat: Switch cloudbuild to use sh files and add tests.


orthros:
Our `cloudbuild.yaml` uses a lot of complex bash. Let's pull the bash steps out into various scripts that can be tested and verified.

- [x] pull cloudbuild steps into scripts
- [ ] test the scripts

JustinBeckwith:
I believe this is done now :)

PR #88, opened 2019-09-19
feat(build): split build steps into scripts/

orthros:
Makes build scripts more testable with `shellcheck`

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -35,57 +35,18 @@ steps:
       - "-c"
       - |
         # Deploy a new version
-        gcloud beta run deploy \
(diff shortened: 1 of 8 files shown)


orthros:
Fixes #87

bcoe:
If we can get tests passing 👍 looks like a great refactor to me.

orthros:
I'll try re-merging from master. Strange that they are all failing; I did not change any of the js files :thinking:

orthros:
Strangely enough this is a problem with master. I'll investigate

orthros:
This will be fixed when #89 is merged

PR #89, opened 2019-09-19
fix: pinned dependency to octokit/rest to 16.28.3

orthros:
This fixes problems introduced with the type system in latest releases
of octokit


orthros:
summoning @bcoe

orthros:
🎉

Issue #91, opened 2019-09-20
improvement: stop using types from octokit


bcoe:
We keep bumping into types incompatibilities between Octokit and probot; Let's look into instead using the types exporte by probot.

related: https://github.com/googleapis/repo-automation-bots/pull/89

CC: @orthros

orthros:
Disclaimer: I am not a full time Node/Typescript aficionado, so this may not be the right answer.

The only way I can see getting this to work is to have Probot declare Octokit as a `peerDependency` and force the upstream consumer (our library) to provide the dependency.

Does this track with your understanding?

bcoe:
@orthros I'm suspicious that probot might have the octokit types buried somewhere in what it ships, haven't dug yet though.

PR #113, opened 2019-10-10
refactor: we can trust that @octokit/rest exists

bcoe:
`probot` is guaranteed to pull `@octokit` into our dependency graph, so there's no need to explicitly install it.

This should prevent the two octokit versions from getting out of sync.

fixes https://github.com/googleapis/repo-automation-bots/issues/91

### DIFF
--- a/packages/blunderbuss/package.json
+++ b/packages/blunderbuss/package.json
@@ -27,7 +27,6 @@
     "lint": "gts check"
   },
   "dependencies": {
-    "@octokit/rest": "16.28.3",
     "gcf-utils": "^1.0.0",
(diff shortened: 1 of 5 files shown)


orthros:
So LGTM, but the kokoro builds failing on TypeScript compilation again. 

```
src/header-checker-lint.ts:82:23 - error TS2339: Property 'content' does not exist on type 'ReposGetContentsResponse'.
  Property 'content' does not exist on type 'ReposGetContentsResponseItem[]'.

82         response.data.content,
                         ~~~~~~~
```

This looks like the type incompatibilities issue again?

PR #114, opened 2019-10-11
feat: bot for running kokoro for trusted, but external, contributors and bots

crwilcox:


packages/trusted-contribution-test-runner/README.md
```diff
@@ -0,0 +1,43 @@
+# header-checker-lint
```
crwilcox: update this readme

README.md
```diff
@@ -33,18 +34,40 @@ GitHub application using the probot server:
 
 1. `cd packages/your-bot`.
 1. `npm start`.
-1. visit:  http://localhost:3000
+1. visit:  http://localhost:3000 and install.
+
+### Granting the Development Application permissions and events
+
+1. By default there will be no permissions. Visit
```
bcoe: thanks for fleshing this out further :+1:

packages/trusted-contribution/src/trusted-contribution.ts
```diff
@@ -0,0 +1,54 @@
...
+  | 'success'
+  | 'failure'
+  | 'neutral'
+  | 'cancelled'
+  | 'timed_out'
+  | 'action_required'
+  | undefined;
+
+// TODO: add this to a configuration file to be extended per repository.
+const TRUSTED_CONTRIBUTORS = ['renovatebot', 'release-please'];
```
bcoe: we could potentially pull this into config we load from the repo, but this seems great so far.

packages/trusted-contribution/src/trusted-contribution.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export = (app: Application) => {
+  app.on('pull_request', async context => {
+    const PR_AUTHOR = context.payload.pull_request.user.login;
+
+    // TODO: add additional verification that only dependency version changes occurred.
+    if (isTrustedContribution(PR_AUTHOR)) {
+      const issuesAddLabelsParams = context.repo({
+        issue_number: context.payload.pull_request.number,
+        labels: ['kokoro:force-run'],
```
bcoe: will force run work without the PR having been approved? I'm guessing yes, but haven't tried.

packages/trusted-contribution/src/trusted-contribution.ts
crwilcox: That is my plan also. I wanted to get this in before expanding since it was my first bot and I know, like this, it will provide a good amount of value.

crwilcox: https://github.com/googleapis/repo-automation-bots/issues/117 to track.

packages/trusted-contribution/src/trusted-contribution.ts
crwilcox: I believe this works yes. I have definitely done this for contributions from external folks since I usually want to make sure tests pass before approving.

packages/trusted-contribution/src/trusted-contribution.ts
```diff
@@ -0,0 +1,54 @@
...
+  | undefined;
+
+// TODO: add this to a configuration file to be extended per repository.
+const TRUSTED_CONTRIBUTORS = ['renovatebot', 'release-please'];
+function isTrustedContribution(author: string): boolean {
+  return TRUSTED_CONTRIBUTORS.includes(author);
+}
+
+export = (app: Application) => {
+  app.on('pull_request', async context => {
```
JustinBeckwith: Is there a way to re-trigger this when the branch is updated?

chingor13: The context payload has an action with opened, reopened, synchronized, etc. In this case it's running on any of those types.

JustinBeckwith: Rad

chingor13: Actually, we probably want to limit this - if kokoro removes the label, will it trigger an event and cause this bot to re-add the label?

PR #118, opened 2019-10-14
fix(trusted contribution): limit bot to trigger on creation and edit

crwilcox:
Acting on feedback left on PR #114

### DIFF
--- a/packages/trusted-contribution/src/trusted-contribution.ts
+++ b/packages/trusted-contribution/src/trusted-contribution.ts
@@ -38,7 +38,7 @@ function isTrustedContribution(author: string): boolean {
 }
 
 export = (app: Application) => {
-  app.on('pull_request', async context => {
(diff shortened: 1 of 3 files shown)


packages/trusted-contribution/src/trusted-contribution.ts
```diff
@@ -38,7 +38,7 @@ function isTrustedContribution(author: string): boolean {
 }
 
 export = (app: Application) => {
-  app.on('pull_request', async context => {
+  app.on(['pull_request.opened', 'pull_request.edited'], async context => {
```
chingor13: I think we want `pull_request.synchronized` too which is when code is pushed to the tracking branch.

crwilcox: Looking at these docs it is sort of hard to tell: https://developer.github.com/v3/activity/events/types/#pullrequestevent

But I can add it here.

PR #125, opened 2019-10-17
fix(trusted-contribution): use kokoro:run instead of kokoro:force-run

tbpg:
It's possible there are jobs that need `force`, but the the repos I've seen have all worked with `kokoro:run`. Can switch back if this doesn't work as expected.


bcoe:
@tbpg I didn't know this was a feature :smile: unfortunately, I think we might need to also update the labels in [sloth](https://github.com/googleapis/sloth) because I don't believe we populate this label by default, which might cause issues.

@JustinBeckwith :point_up: can you confirm or deny this.

tbpg:
Good call. I sent https://github.com/googleapis/sloth/pull/364 to add the `kokoro:run` label to Sloth.

bcoe:
@tbpg awesome, thank you for the contribution. I'm going to try out the new tag on one repo, and once I see it working, I'll merge this.

PR #134, opened 2019-10-28
feat: initial check-in of buildcop bot

bcoe:
Stubbed out bot for build-cop.

## Todo

- [ ] add information in top-level README linking to bot.
- [ ] add a section on the Yoshi bot site.
- [ ] actually make it do something (with tests and such, talk @bcoe about the framework).
- [ ] test with pubsub (talk to @orthros about this).


bcoe:
Let's use @sofisl generator bot for this.

PR #135, opened 2019-10-28
feat(serverless-scheduler-proxy): add PubSub Support

### LINKED ISSUE #98: feature: PubSub triggers
Similarly to Cron Support; there are events from other services (e.g. Kokoro) whose job completion may trigger the invocation of a bot.

### Bots
* A bot that wants to subscribe to PubSub triggers needs to include a `topic` file in their source folder
* A bot that wishes to subscribe to PubSub triggers needs to subscribe to the `topic.repository` event in the appFn:
```typescript
 app.on('topic.repository', async context => {});
```

### Infrastructure
* The build process will inspect all bots && if a bot has a `topic` file and no corresponding PubSub topic it will
  * Create a PubSub topic whose name is the bot name
  * Create a Push Subscription for the topic and point it at the Cloud Run `serverless-scheduler-proxy` instance.  NOTE: This will use a custom ServiceAccount that has permissions to invoke the CloudRun instance.

### serverless-scheduler-proxy
* Move the cron processing endpoint to `/cron` (was `/`)
* Add a new endpoint `/pubsub` which handles incoming PubSub push messages of form:
```go
// PubSubMessage is the payload of a Pub/Sub event.
type PubSubMessage struct {
        Message struct {
                Data []byte `json:"data,omitempty"`
                ID   string `json:"id"`
        } `json:"message"`
        Subscription string `json:"subscription"`
}
```
* This handler parses the name of the bot from the `Subscription` field, and creates a new HTTP request to the bot/Cloud Function whose name is in the subscription. This request will be signed using the signing key of the bot in the `x-hub-signature` header, and the `x-github-event` header will be `topic.repository`

orthros:
There are some cases where an external source needs to trigger a Bot to
interact with GitHub (such as Kokoro).

This moves the "v0" URL (which used to support cron) to "v0/cron" and
adds a new URL handler at "v0/pubsub" to handle incoming PubSub
messages.

Fixes #98

### DIFF
--- a/serverless-scheduler-proxy/main.go
+++ b/serverless-scheduler-proxy/main.go
@@ -68,7 +68,8 @@ func main() {
 	}
 
 	mux := http.NewServeMux()
-	mux.Handle("/v0", botCronProxy(cfg))
+	mux.Handle("/v0/cron", botCronProxy(cfg))
+	mux.Handle("/v0/pubsub", botPubSubProxy(cfg))
 
 	addr := ":" + port
 	log.Printf("starting to listen on %s", addr)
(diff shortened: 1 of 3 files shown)


serverless-scheduler-proxy/main.go
```diff
@@ -88,45 +89,70 @@ func main() {
...
+			json.Unmarshal(bodyBytes, &pay)
+			return pay.Name, pay.Location
 		}
+		rewriteBotURL(c, parser, req)
+	}
+}
 
-		var pay reqPayload
-		json.Unmarshal(bodyBytes, &pay)
+func rewriteBotPubSubURL(c botConfig) func(*http.Request) {
```
bcoe: we're going to be able to use this for both `pubsub` and `cron`? I noticed they share a helper.

serverless-scheduler-proxy/main.go
```diff
@@ -88,45 +89,70 @@ func main() {
...
+	return func(req *http.Request) {
+		req.Header.Add("x-github-event", "pubsub.message")
+		parser := func(bodyBytes []byte) (string, string) {
+			var pay PubSubMessage
+			json.Unmarshal(bodyBytes, &pay)
+			log.Printf("handling pubsub message for subscription: %v\n", pay.Subscription)
+
```
bcoe: I"m pretty sure we need to find a way to insert an `installation` identifier, perhaps this can be set as the job is dispatched? what are you thinking.

orthros:
100% agree on the tests (tis one of the reasons why this is a draft). 

I'm going to look into the Serverless proxy for probot that I based this on; I'm not 100% sure we need the installation ID, but I'm not sure. We might need to schedule some time to go over why you think that and why I dont

bcoe:
Colin and I have been having some offline conversations about how to best deal with the installation ID.

bcoe:
I'm feeling like we might just be able to piggyback off the existing cron functionality, can't we wire up a pubsub queue to just call the same redirecting proxy the scheduler calls?

orthros:
That’s exactly what my (yet unpublished) PR does!

bcoe:
rebased in: https://github.com/googleapis/repo-automation-bots/pull/185

PR #137, opened 2019-10-31
feat: support for GitHub releases and Ruby

bcoe:
This adds support to our bot for creating GitHub releases, if `handleGHRelease` is set to `true`.

This is exciting to me because it allows me to start adding release-please to non-autorelease repos, e.g., [yargs](https://github.com/yargs/yargs/pull/1462).

---

This also updates the version of `release-please` used by by the bot, which has additional functionality added for @TheRoyalTnetennba; Ruby will be using the `Ruby` release type for the bot, not `RubyYoshi`.

### DIFF
--- a/packages/release-please/src/release-please.ts
+++ b/packages/release-please/src/release-please.ts
@@ -60,8 +60,7 @@ function releaseTypeFromRepoLanguage(language: string | null): ReleaseType {
   }
 }
 
-// creates or updates the evergreen release-please
-// release PR.
+// creates or updates the evergreen release-please release PR.
 function createReleasePR(
   releaseType: ReleaseType,
   packageName: string,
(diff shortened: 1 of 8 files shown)


packages/release-please/src/release-please.ts
```diff
@@ -20,6 +20,10 @@ import { Application } from 'probot';
 // See https://github.com/googleapis/release-please/issues/249
 import { ReleaseType, BuildOptions } from 'release-please/build/src/release-pr';
 import { ReleasePRFactory } from 'release-please/build/src/release-pr-factory';
+import {
+  GitHubRelease,
+  GitHubReleaseOptions,
+} from 'release-please/build/src/github-release';
```
JustinBeckwith: Should probably export these types from the index, so consumers are resilient to files moving around.

bcoe: good call.

PR #139, opened 2019-11-05
chore: update to latest version of release-please

bcoe:
Updates to latest version of release please, with support for additional version file in Java repo.

I've also fixed coverage, c8 should only be called at the top level.

### DIFF
--- a/packages/label-sync/package.json
+++ b/packages/label-sync/package.json
@@ -19,7 +19,7 @@
     "start": "probot run ./build/src/label-sync.js",
     "prepare": "npm run compile",
     "pretest": "npm run compile",
-    "test": "cross-env LOG_LEVEL=fatal c8 mocha --timeout=15000 --recursive ./build/test",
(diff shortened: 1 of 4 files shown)


packages/label-sync/package.json
```diff
@@ -19,7 +19,7 @@
     "start": "probot run ./build/src/label-sync.js",
     "prepare": "npm run compile",
     "pretest": "npm run compile",
-    "test": "cross-env LOG_LEVEL=fatal c8 mocha --timeout=15000 --recursive ./build/test",
+    "test": "cross-env LOG_LEVEL=fatal mocha --timeout=15000 --recursive ./build/test",
```
JustinBeckwith: Interesting :) Why? I like having the coverage visible to me while working on a bot.

bcoe: you could add a `coverage` script that has `c8`? It's fine to have the `c8` dependency, we just want the bin to not reset the coverage output, which happens when `c8 foo` is called.

PR #143, opened 2019-11-06
feat: add script for generating new bots

sofisl:
Fix: Add data normalization, fix failing tests, add tests for new data validation, create default location for bot


sofisl:
Feel free to ignore the changes to blunderbuss, I'm not sure how to leave them out of the PR

bcoe:
A bit of meta-comment, I would make the pull request description include what the bot does, e.g., "to simplify the creation of new bots, this introduces a tool that generates a skeleton GitHub application using templates".

lerna-debug.log
```diff
@@ -0,0 +1,1248 @@
+0 silly argv {
+0 silly argv   _: [ 'bootstrap' ],
+0 silly argv   hoist: true,
+0 silly argv   lernaVersion: '3.17.0',
```
bcoe: you should remove this from the repository with `git rm lerna-debug.log`. I would also add this as a line to `.gitignore`, so that we don't accidentally add it in the future.

packages/blunderbuss/src/blunderbuss.ts
```diff
@@ -54,15 +54,15 @@ export = (app: Application) => {
       'pull_request.labeled',
     ],
     async context => {
-      const config = (await context.config(
-        CONFIGURATION_FILE_PATH,
-        {}
-      )) as Configuration;
-      const issue = context.issue() as Issue;
+      // const config = (await context.config(
+      //   CONFIGURATION_FILE_PATH,
```
bcoe: we should revert any of the `blunderbus` changes that were on this branch.

bcoe: `git checkout origin/master [filename]`

:point_up: you can just do something like this and checkout the specific files from a specific branch.

packages/label-sync/package-lock.json
```diff
@@ -34,130 +34,6 @@
 				"lazy-ass": "1.6.0"
 			}
 		},
-		"@bcoe/v8-coverage": {
-			"version": "0.2.3",
-			"resolved": "https://registry.npmjs.org/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz",
```
bcoe: as with the blunderbuss changes, you shouldn't need to make any updates to these files, I think you can use the:

`git checkout origin/master [filename]`

Approach to revert any files that changed in a folder other than `googleapis/generate-bot`.

sofisl:
@bcoe - Addressed the comments above, thanks so much for the in-depth review!

packages/generate-bot/templates/src/app.ts
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { GCFBootstrapper } from 'gcf-utils';
+import appFn from './{{programName}}';
+
+const bootstrap = new GCFBootstrapper();
+module.exports.{ { programName } } = bootstrap.gcf(appFn);
```
bcoe: this should be `{{programName}}` or else the template string won't be replaced :+1:

packages/generate-bot/main.js
```diff
@@ -0,0 +1,124 @@
...
+
+const { prompt } = require("enquirer");
+const Handlebars = require("handlebars");
+const fs = require("fs");
+const path = require("path");
+const process = require("process");
+
+exports.checkValidity = function(testString) {
+  let isValid = true;
+  const relativePath = path.resolve("../repo-automation-bots/packages");
```
bcoe: I believe this can just be:

```
path.resolve('./packages')
```

packages/generate-bot/test/generate-bot.js
```diff
@@ -0,0 +1,145 @@
...
+  contentString = contentString || null;
+  files.forEach(function(file) {
+    const readName = path.join(dirNameRead, file);
+    if (fs.statSync(readName).isDirectory()) {
+      readAllFiles(readName, contentString);
+    } else {
+      contentString += fs.readFileSync(readName);
+    }
+  });
+  console.log(contentString);
```
bcoe: let's get rid of this `console.log`, because it makes tests hard to debug (it's causing a lot of output).

packages/generate-bot/test/generate-bot.js
```diff
@@ -0,0 +1,145 @@
...
+
+  it("checks that the file content carries over", async () => {
+    GenerateBot.creatingBotFiles("./templates", {
+      programName: "helloWorld",
+      description: "says hi",
+      fileLocation: "./helloWorld"
+    });
+
+    const contentString = "Start of snapshot: ";
+    return snapshot(readAllFiles("./helloWorld", contentString));
```
bcoe: try replacing this with:

```
return snapshot(readAllFiles("./helloWorld", contentString).replace('\r\n', '\n'));
```

Windows has `\r\n` on the end of each line rather than `\n`, which might be the source of one of your failing windows tests.

packages/generate-bot/test/generate-bot.js
```diff
@@ -0,0 +1,145 @@
...
+      !integerTest &&
+      !overWritingTest &&
+      programNameToLower.programName === "passbutmakelowercase" &&
+      regex.test(fileLocationDefault.fileLocation)
+    ) {
+      validityWorkingWell = true;
+    } else {
+      validityWorkingWell = false;
+    }
+    expect(validityWorkingWell).to.be.true;
```
bcoe: something is breaking in windows, but it's hard to tell what exactly, because of all our assertions being combined, I would do this as:

```js
expect(nullTest).to.be.true;
expect(integerTest).to.be.true;
... etc.
```

packages/generate-bot/templates/test/Fixtures/config/valid-config.yml
```diff
@@ -0,0 +1 @@
+randomBoolean: true
```
bcoe: this folder should be lowercase, `fixtures`, rather than `Fixtures` (I noticed the generated tests fail due to the uppercase letter).

packages/generate-bot/templates/test/{{programName}}.ts
```diff
@@ -0,0 +1,97 @@
...
+    it('responds to issues', async () => {
+      const payload = require(resolve(
+        fixturesPath,
+        './events/issue_opened'
+      ));
+
+      const requests = nock('https://api.github.com')
+        .get('/repos/testOwner/testRepo/contents/.github/{{programName}}.yml')
+        .reply(200, { content: config.toString('base64') })
+        .log(console.log)
```
bcoe: let's remove this `.log(console.log)` :+1:  (it's creating a tiny bit of noise in logs).

bcoe:
@sofisl I ran the generator to create the pull request here:

https://github.com/googleapis/repo-automation-bots/pull/158

First off, incredible work, it worked great!

Two very small nits jumped out at me:

1. the `.gitkeep` file doesn't need any content, it can just be an empty file.
2. we should add `#` and `##` headings to the titles in the README.md (this will make them headints).

PR #144, opened 2019-11-07
fix: address issues with loading labels

bcoe:
@sofisl and I tested this against an actual repo, we found a couple bugs in the process:

1. requesting `https://github.com/googleapis/repo-automation-bots/blob/master/packages/label-sync/src/labels.json'` did not work (returned a 404 I believe); we switched to using the API.
2. `Promise.all()` created a race condition where deletion and creation could potentially happen in conjunction; I switched the logic to being synchronous for the time being (I think for how many labels we have, and how frequently this will run, this is a reasonable decision for the time being).

### DIFF
--- a/packages/label-sync/src/label-sync.ts
+++ b/packages/label-sync/src/label-sync.ts
@@ -41,19 +41,20 @@ interface Repos {
 // from the local copy.  We are using the `PushEvent` to detect the change,
 // meaning the file running in cloud will be older than the one on master.
 let labelsCache: Labels;
-
-async function getLabels() {
+async function getLabels(github: GitHubAPI) {
   if (!labelsCache) {
-    await refreshLabels();
+    await refreshLabels(github);
   }
   return labelsCache;
 }
 
-async function refreshLabels() {
-  const url =
-    'https://github.com/googleapis/repo-automation-bots/blob/master/packages/label-sync/src/labels.json';
-  const res = await request<Labels>({ url });
-  labelsCache = res.data;
+async function refreshLabels(github: GitHubAPI) {
+  const data = (await github.repos.getContents({
+    owner: 'googleapis',
+    repo: 'repo-automation-bots',
+    path: 'packages/label-sync/src/labels.json'
+  })).data as { content?: string };
+  labelsCache = JSON.parse(Buffer.from(data.content as string, 'base64').toString('utf8'))
 }
 
 export = (app: Application) => {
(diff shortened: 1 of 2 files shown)


packages/label-sync/src/label-sync.ts
```diff
@@ -41,19 +41,20 @@ interface Repos {
...
   return labelsCache;
 }
 
-async function refreshLabels() {
-  const url =
-    'https://github.com/googleapis/repo-automation-bots/blob/master/packages/label-sync/src/labels.json';
-  const res = await request<Labels>({ url });
-  labelsCache = res.data;
+async function refreshLabels(github: GitHubAPI) {
+  const data = (await github.repos.getContents({
```
JustinBeckwith: Oooops, my mistake.  I think you can fix this just by updating the url to https://raw.githubusercontent.com/googleapis/repo-automation-bots/master/packages/label-sync/src/labels.json

Shouldn't need to use the API and thread all this through for a simple GET (you know, when you use the right url 🤦‍♂ )

packages/label-sync/src/label-sync.ts
```diff
@@ -94,15 +97,15 @@ export = (app: Application) => {
...
+  const newLabels = await getLabels(github);
   const res = await github.issues.listLabelsForRepo({
     owner,
     repo,
     per_page: 100,
   });
   const oldLabels = res.data;
   const promises = new Array<Promise<unknown>>();
-  newLabels.labels.forEach(l => {
+  for (const l of newLabels.labels) {
```
JustinBeckwith: I want to poke at this just a little bit:

> Promise.all() created a race condition where deletion and creation could potentially happen in conjunction; I switched the logic to being synchronous for the time being (I think for how many labels we have, and how frequently this will run, this is a reasonable decision for the time being).

The labels that are being deleted are always different from the ones being created or update.  That is to say - the deletes should never interfere with the create/updates because they can never be on the same resources.  I've been running this script for over a year (albeit manually) with no issues.  Did you experience a problem here, or did you notice a potential for problems?

packages/label-sync/src/label-sync.ts
bcoe: One reason I'd advocate using the authenticated API instance we have available, is that it means that the request will fall under the quota of an authenticated client, vs., an unauthenticated request (which I believe has a significantly lower request limit).

Might not matter for the throughput of this bot, but worth considering.

packages/label-sync/src/label-sync.ts
bcoe: I was seeing a lot of exceptions running the bot locally, I believe something long the lines of `VALIDATION_ERROR` (_actual logs aren't handy, but I can find them tomorrow).

My hunch was that:

* since there's no guarantee of the ordering of `Promise.all`, perhaps requests were being queued that were somehow contradictory with each other?
* or, perhaps this is more likely, the errors were cropping up when too many operations where happening concurrently?

In general, I'm not a huge proponent of an unbounded `Promise.all`, just because there's nothing that limits the amount of concurrency. An alternative refactor might be to figure out a way to kick off and await `N` promises at  time.

```js
let count  = 0
const work = []
for (const l of labels) {
  work.push(l)
  if ((i % 4) === 0) await Promise.all(work)
  count++
}
await Promise.all(work)
```

I can test more tomorrow with whatever approach you land on.

PR #157, opened 2019-11-22
fix: update refresh label logic

sofisl:
@bcoe and I applied the following fixes after testing the bot:

1. lint
2. Made sure the bot checked for mismatching descriptions on preexisting labels, and made sure the descriptions were short enough so that the bot would update the descriptions on the labels
3. Made sure the bot didn't try deleting labels if there were no labels (i.e., new repo)
4. When a new repo is created, the bot creates labels and updates the ones that come prepackaged w/i github. This means that the bot triggers itself because it is also triggered when a label is edited. This can cause issues because it tries to create the same label more than once. We've updated the logic so that it doesn't log errors with the code 'already_created.'


packages/label-sync/src/label-sync.ts
```diff
@@ -73,12 +77,14 @@ export = (app: Application) => {
     const { owner, repo } = context.repo();
     // TODO: Limit this to pushes that edit `labels.json`
     if (
-      owner === 'googleapis' &&
+      owner === 'sofisl' &&
```
bcoe: looks like you left in some testing data here, this should be `googleapis`.

packages/label-sync/src/label-sync.ts
```diff
@@ -41,19 +42,22 @@ interface Repos {
...
 }
 
-async function refreshLabels() {
-  const url =
-    'https://github.com/googleapis/repo-automation-bots/blob/master/packages/label-sync/src/labels.json';
-  const res = await request<Labels>({ url });
-  labelsCache = res.data;
+async function refreshLabels(github: GitHubAPI) {
+  const data = (await github.repos.getContents({
+    owner: 'sofisl',
```
bcoe: this should be `googleapis` too.

packages/label-sync/src/label-sync.ts
```diff
@@ -73,12 +77,14 @@ export = (app: Application) => {
...
     if (
-      owner === 'googleapis' &&
+      owner === 'sofisl' &&
       repo === 'repo-automation-bots' &&
       context.payload.ref === 'refs/heads/master'
     ) {
-      await refreshLabels();
+      await refreshLabels(context.github);
       // TODO: Use the GitHub installations API to retreive this list
+      //   (also, our QA indicates there's a good chance this request isn't
```
bcoe: I think we can get rid of thee comments, seems like your QA showed us things were working.

packages/label-sync/src/label-sync.ts
```diff
@@ -156,23 +163,24 @@ async function reconcileLabels(github: GitHubAPI, owner: string, repo: string) {
...
-          console.log(`Deleted '${l.name}' from ${owner}/${repo}`);
-        })
-        .catch(e => {
-          console.error(`Error deleting label ${l.name} in ${owner}/${repo}`);
-          console.error(e.stack);
-        });
-      promises.push(p);
+  if (!Object.keys(oldLabels).length) {
+    for (const l of oldLabels) {
+      console.log(l.name);
```
bcoe: let's get rid of this label, but perhaps add something more descriptive right above `await github.issues`.

```
console.info(`deleting label ${l.name}`)
```

packages/label-sync/package.json
```diff
@@ -25,9 +25,11 @@
   },
   "dependencies": {
     "@octokit/rest": "^16.34.0",
+    "@types/uuid": "^3.4.6",
     "gaxios": "^2.1.0",
     "gcf-utils": "1.1.0",
-    "probot": "9.6.4"
+    "probot": "9.6.4",
+    "uuid": "^3.3.3"
```
bcoe: we didn't end up using `uuid`, so let's get rid of this.

bcoe:
```
npm uninstall uuid
npm uninstall @types/uuid
```

packages/label-sync/src/labels.json
```diff
@@ -62,7 +62,7 @@
     },
     {
       "name": "priority: p0",
-      "description": "Highest priority. Critical issue. P0 implies highest priority (e.g. blocking imminent release). P0 issues should be monitored on a daily basis.",
```
bcoe: GitHub had a limit on the length of label description @JustinBeckwith; awesome debugging @sofisl.

I believe this was the error that we were seeing when we tried to create labels in parallel with `Promise.all`, I still think for the time being this approach is reasonable however, it's helped us debug a couple issues more easily to run in series.

packages/label-sync/src/label-sync.ts
sofisl: great catch!

packages/label-sync/src/label-sync.ts
sofisl: Just deleted line 165, as it logs at the end if the label was deleted or not!

packages/label-sync/src/label-sync.ts
```diff
@@ -17,6 +17,7 @@
 import { Application } from 'probot';
 import { request } from 'gaxios';
 import { GitHubAPI } from 'probot/lib/github';
+import e = require('express');
```
bcoe: I don't think we need this import.

packages/label-sync/src/label-sync.ts
```diff
@@ -140,12 +142,14 @@ async function reconcileLabels(github: GitHubAPI, owner: string, repo: string) {
           name: l.name,
         })
         .catch(e => {
-          console.error(`Error creating label ${l.name} in ${owner}/${repo}`);
-          console.error(e.stack);
+          //ignores errors that are caused by two requests kicking off at the same time
+          if (e.errors[0].code !== 'already_exists') {
```
bcoe: love this 😄

bcoe:
@sofisl thanks for thoroughly testing this :tada:

PR #166, opened 2019-11-27
feat: scheduled tasks now execute authenticated for each repository

bcoe:
This PR reworks our scheduled tasks logic, such that it will now execute the task in the context of each GitHub repository (based on our central list in sloth).

If the upstream task sets an installation ID, this should also result in the tasks being appropriately authenticated.


jadekler:
LGTM but I don't know anything about node.

PR #167, opened 2019-11-27
feat: improvements to cron functionality

bcoe:
* fixed a couple small bugs with testing cron job (it was not running inside work hours).
* add support for crons that run on a single repository.
* parallelize crons, so that it doesn't take 50 years to run on all our repos.
* in production express receives a Buffer not a string or object, we now sniff this behavior out.


packages/gcf-utils/test/fixtures/repos.json
```diff
@@ -3,6 +3,7 @@
   "repos": [
     { "language": "cpp", "repo": "googleapis/cpp-cmakefiles"},
     { "language": "cpp", "repo": "googleapis/google-cloud-cpp"},
-    { "language": "cpp", "repo": "googleapis/google-cloud-cpp-bigquery"}
+    { "language": "cpp", "repo": "googleapis/google-cloud-cpp-bigquery"},
```
bcoe: added a fourth repo, so that we test the edge-case of the parallelization logic.

Issue #178, opened 2019-12-11
New Bot: Build cop


tbpg:
Manage issues for failing unit tests.

- [x] Pub/Sub support in the proxy.
- [x] A bot to consume the Pub/Sub events, parse test output, and manage issues.
- [ ] Docs: how to publish, what format, etc.

cc @sofisl @bcoe @orthros @dzlier-gcp @fhinkel

tbpg:
#237 added support for the Python flavor of xUnit output.

PR #179, opened 2019-12-12
feat: introduce initial version of buildcop-bot

tbpg:
We still need to add PubSub integration. That will change the payload and setup slightly (I left a few TODOs in a few of the places we'll need to update). I defined a `PubSubPayload` and `PubSubContext`, but we might want to move them to a shared location.

The bot is rather spammy right now. But, all state is stored on the issues themselves (rather than, say, Firestore). The complexity comes from the fact a single build can run N jobs (e.g. for different language versions). If a test fails in one of those N, we should leave the issue open.

This implementation of the bot works by including a build ID in the issue body and all status update comments. Then, before closing an issue when it passes in a build, the issue body and comments are parsed to see if the test has already failed in the same build. If not, the bot closes the issue. If another job in the same build fails _after_ that, the issue is reopened. Theoretically, repos could configure the number of jobs, but that seems brittle.

There are no configuration options right now. We might want one in the future to be able to configure the issue title.

cc @dzlier-gcp @fhinkel (won't let me add as reviewers).

Updates #178.


packages/buildcop/README.md
```diff
@@ -0,0 +1,21 @@
+buildcop
```
bcoe: nit, but we should probably update the upstream template to make this title case:

```
# buildcop
```

tbpg: Updated this readme.

PR #180, opened 2019-12-16
fix: improve documentation on using genkey

sofisl:
genkey tool did not work with just npm i, had to add npm link in order to get command to work. I think it's worth noting in the readme documentation.


packages/gcf-utils/src/bin/README.md
```diff
@@ -47,6 +47,7 @@ Application you wish to use with your bot.
  export BUCKET=<NAME OF BUCKET TO STORE KEYS>
  export BOT=<NAME OF YOUR BOT>
  ```
+4. Install dependencies and link using npm i and npm link
 
 5. Run the `genkey` tool
```
orthros: nit: Numbers need to increment
4 -> 5
5 -> 6

packages/gcf-utils/src/bin/README.md
```diff
@@ -47,6 +47,7 @@ Application you wish to use with your bot.
  export BUCKET=<NAME OF BUCKET TO STORE KEYS>
  export BOT=<NAME OF YOUR BOT>
  ```
+4. Install dependencies and link using npm i and npm link
```
bcoe: also a nit, I trend to put commands in back ticks, `npm i` and `npm link`, so that they jump out as commends.

packages/gcf-utils/src/bin/README.md
sofisl: ah, yes, great catch! Thanks!

packages/gcf-utils/src/bin/README.md
sofisl: sounds good!

sofisl:
@bcoe - submitted changes!

PR #184, opened 2019-12-19
feat: add bot for npm publication through wombat

bcoe:
Bot for publishing to npm from cloud function, using wombat dressing room.


packages/publish/test/publish.ts
```diff
@@ -0,0 +1,142 @@
+/**
```
JustinBeckwith: Did you mean to check in the tarball above?

packages/publish/src/publish.ts
```diff
@@ -0,0 +1,210 @@
...
+  return `/tmp/${uuid.v4()}`;
+};
+
+handler.getPublicationSecrets = async (
+  app: Application,
+  config: Configuration
+): Promise<Secret> => {
+  const secretId = config.secretId || 'publish';
+  const project = config.project || process.env.PROJECT_ID;
+  app.log.info(`looking secret for ${project}`);
```
tbpg: "looking secret"?

packages/publish/test/fixtures/config/valid-config.yml
```diff
@@ -0,0 +1 @@
+randomBoolean: true
```
tbpg: Update?

packages/publish/test/publish.ts
```diff
@@ -0,0 +1,142 @@
...
+import snapshot from 'snap-shot-it';
+import nock from 'nock';
+import * as fs from 'fs';
+import { expect } from 'chai';
+
+nock.disableNetConnect();
+
+const fixturesPath = resolve(__dirname, '../../test/fixtures');
+
+interface Secret {
```
tbpg: Export `Secret` and `PublishConfig` to avoid having to duplicate the interface declarations?

packages/publish/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "dotenv": "^8.0.0",
+    "gts": "^1.0.0",
+    "mocha": "^6.1.4",
+    "nock": "^11.0.0",
+    "smee-client": "^1.1.0",
+    "snap-shot-it": "^7.8.0",
+    "typescript": "^3.5.3"
+  },
+  "engines": {
+    "node": ">= 10.13.0"
```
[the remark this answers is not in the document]
bcoe: I'm currently skipping Node 8 tests in this bot, within the test; `fs.promises` let me clean up a lot of logic to use `async`/`await`, and I figured that we're dropping Node 8 in the pretty near future any ways (and I control the runtime that the bot gets deployed to).

packages/publish/src/publish.ts
```diff
@@ -0,0 +1,210 @@
...
+function handler(app: Application) {
+  app.on('release.released', async context => {
+    const repoName = context.payload.repository.name;
+    const remoteConfiguration: Configuration | null = (await context.config(
+      CONFIGURATION_FILE_PATH
+    )) as Configuration | null;
+
+    // Skip publication unless it's explicitly enabled.
+    if (!remoteConfiguration) {
+      app.log.info(`publish not configured for (${repoName})`);
```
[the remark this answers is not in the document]
bcoe: :+1: rather than doing so in this PR, perhaps we could create a tracking issue to override the default logger for all our bots; it would be nice to add a bit more context in the logs in general.

packages/publish/src/publish.ts
```diff
@@ -0,0 +1,210 @@
...
+    if (context.payload.release.tarball_url) {
+      // Create a temporary directory to unpack release tarball to:
+      const unpackPath = handler.unpackPath();
+      app.log.info(`creating tmp directory ${unpackPath}`);
+      await fs.mkdir(unpackPath, {
+        recursive: true,
+      });
+
+      // Unpack the tarball to the directory we just created:
+      await fetch(context.payload.release.tarball_url).then(res => {
```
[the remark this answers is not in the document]
bcoe: :+1: yeah, I figured the extenuating circumstance of using `tar`, which isn't particularly suited to a promise API, made it an acceptable evil ... I wanted to use this library since it's the one used by `npm`, and I trust it.

packages/publish/src/publish.ts
```diff
@@ -0,0 +1,210 @@
...
+      if (files.length === 1 && files[0].isDirectory()) {
+        pkgPath = `${pkgPath}/${files[0].name}`;
+      }
+
+      const secret: Secret = await handler.getPublicationSecrets(
+        app,
+        remoteConfiguration
+      );
+      if (secret && secret.payload && secret.payload.data) {
+        const publishConfig = handler.publishConfigFromSecret(secret as Secret);
```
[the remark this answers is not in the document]
bcoe: I think this was hold over from before some refactoring, removed the cast.

packages/publish/test/fixtures/config/valid-config.yml
bcoe: :+1:  updated to reflect actual config format.

packages/publish/test/publish.ts
bcoe: Right now, the way our bots are designed, they expect a single export which is the bot's handler; If I export the interfaces I get the following compile error:

```bash
: An export assignment cannot be used in a module with other exported elements
```

We could potentially pull the fixtures into a new file, if we find they're being updated frequently and the tests are getting more complex ... but I figured this is probably reasonable for now?

PR #195, opened 2019-12-23
fix: slight changes to payload format, based on testing

tbpg:
I also renamed the app to `buildcop` and adjusted exports so I could use `BuildCopPayload` from tests. Let me know if there is a better way to do that -- wasn't sure the correct way.

Updates #178.

### DIFF
--- a/packages/buildcop/src/buildcop.ts
+++ b/packages/buildcop/src/buildcop.ts
@@ -57,18 +57,14 @@ interface PubSubContext {
   github: GitHubAPI;
   log: LoggerWithTarget;
   payload: {
-    message: {
-      data: string; // base64 encoded value with {payload: BuildCopPayload}.
-    };
-  };
+    payload: BuildCopPayload
+  }
 }
 
 export function buildcop(app: Application) {
   app.on('pubsub.message', async (context: PubSubContext) => {
-    const data = JSON.parse(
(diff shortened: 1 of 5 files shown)


bcoe:
@tbpg I will have broken all of your tests, because of the refactoring I did to match [this pr](https://github.com/googleapis/repo-automation-bots/pull/196).

packages/buildcop/src/buildcop.ts
```diff
@@ -44,28 +44,33 @@ interface TestFailure {
...
+export function buildcop(app: Application) {
   app.on('pubsub.message', async (context: PubSubContext) => {
-    const owner = context.payload.repoOwner;
-    const repo = context.payload.repoName;
-    const buildID = context.payload.buildID || '[TODO: set buildID]';
-    const buildURL = context.payload.buildURL || '[TODO: set buildURL]';
-    const xml = context.payload.xunitXML;
+    const payload: BuildCopPayload = context.payload.payload as BuildCopPayload;
+
+    const owner = payload.repoOwner;
```
bcoe: rather than setting this in your payload, you could use the approach here:

https://github.com/googleapis/repo-automation-bots/blob/master/packages/failurechecker/src/failurechecker.ts#L43

This gets populated magically by `gcf-utils`.

PR #196, opened 2019-12-23
feat: make gcf-utils work with pubsub

bcoe:
@tbpg this adds support for the `pubsub.message` message type, a couple notes:

1. it takes care of parsing `body.message.data` for you, so in your bot you should just assume you've gotten the parsed JavaScript object represented by `data`.
2. you must set a top level key `repo` in the message you dispatch, it uses this to decide what repo the cloud function is running for.

### DIFF
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -36,6 +36,7 @@ interface Repos {
 
 interface Scheduled {
   repo?: string;
+  message?: {[key: string]: string};
 }
 
 export class GCFBootstrapper {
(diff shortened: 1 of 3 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -135,14 +136,21 @@ export class GCFBootstrapper {
     };
   }
 
-  private async handleScheduled(id: string, req: express.Request) {
-    const body = (Buffer.isBuffer(req.body)
+  private async handleScheduled(id: string, req: express.Request, eventName: string) {
```
tbpg: Consider renaming this function since it now handles scheduled and PubSub events.

Issue #202, opened 2019-12-30
update license headers in bot generation


bcoe:
We've been moving our template towards license headers with a `//` prefix rather than `/*`, see: #184, we should update our bot generation script accordingly.

JustinBeckwith:
Looks to have been fixed in #209

Issue #177, opened 2019-12-09
Allow bots to have hyphens in their names when created with generate-bot


crwilcox:
Today, if you run the generate-bot and attempt to name a package with a hyphen ('-') it will fail.

```
✔ What is the name of the program? · trusted-contributions
✔ What is the description of the program? · blah
✔ This package will be saved in /packages/yourProgramName unless you specify another location and directory name here relative to /home/crwilcox/workspace/repo-automation-bots :  · /package/trusted-contributions

You used an invalid character, like a hyphen or an integer. Please try again.
```

While not necessary, it is a convention we have used to this point. It would be great if we could extend the generator to support this.

bcoe:
👍 I'd love this functionality too, we will need to make sure to dig in to @orthros' deploy script to make sure we don't break any of the logic for keynames, etc.

sofisl:
Thanks for mentioning Chris! Even though @orthos's script handles hyphens, I can't export the function to cloud functions with a hyphen.

Would it help if we allowed hyphens, but just had the bot change the hyphens to underscores? Or, open to any other suggestions you might have!

tmatsuo:
I think it's possible now.

PR #209, opened 2020-01-03
fix: modified snapshots, added tests, fixed tests

sofisl:
Fixes:

Feature request #177 by allowing user to input hyphens and changing to underscores

Issue #202: updated license headers

Issue #199 by updating dependency and modifying test so that it does not get angry over updated dependencies

Also adds the chai library and an example of using it to the test of the actual bot in order to show its functionality

### DIFF
--- a/packages/generate-bot/main.js
+++ b/packages/generate-bot/main.js
@@ -22,11 +22,21 @@ const process = require("process");
 exports.checkValidity = function(testString) {
   let isValid = true;
   const relativePath = path.resolve("./packages");
-  const invalidName = /[0-9]/;
-  const string = JSON.stringify(testString);
-  if (invalidName.test(string)) {
+  const validName = /[^-A-Za-z_]+/;
+  let string = JSON.stringify(testString);
+  string = string
+    .replace('{"programName":', "")
+    .replace(',"description":', "")
+    .replace(',"fileLocation":', "")
+    .replace(/"/g, "")
+    .replace(/}$/, "");
+
+  console.log(string);
+  if (validName.test(string)) {
     isValid = false;
-    console.log("You used an integer. Please try again.");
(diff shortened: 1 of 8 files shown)


bcoe:
@sofisl looks like there's a chance you might need to run `npm run fix` inside your project?

packages/generate-bot/test/generate-bot.js
```diff
@@ -146,4 +134,16 @@ describe("user input", () => {
     const regexLinux = new RegExp("packages/pass");
     expect(regexLinux.test(fileLocation)).to.be.true;
   });
+
+  it("checks that hyphens are turned into underscores", () => {
+    const hyphenTest = {
+      programName: "pass-but-change",
+      description: "pass",
```
bcoe: it would probably also be good to make sure that we allow someone to provide the name `foo_bar` for a bit, i.e., that we allow underscores.

sofisl: Thanks for the review @bcoe! Implemented the logic we discussed for module.exports['program-name'] to allow for hyphens. I'll also submit a PR to Colin's script to encourage the use of hyphens as well!

packages/generate-bot/main.js
```diff
@@ -23,13 +22,11 @@ const process = require("process");
 exports.checkValidity = function(testString) {
   let isValid = true;
   const relativePath = path.resolve("./packages");
-  const invalidName = /[-0-9]/;
+  const invalidName = /[0-9]/;
```
bcoe: my concern with this is that it might include `@`, `ツ`, etc., instead of `invalidName`, I would have, `validName`:

```
const validName = /[-a-z_]+/;
```

And just reverse the logic, check for validity rather than that something isn't valid.

PR #213, opened 2020-01-07
feature: first stab at merge-on-green 

sofisl:
This PR:

(1) responds to PRs, identifies the owner, repo and branch for a specific PR
(2) identifies required status checks and/or config files with required status checks 
(3) posts a status check that passes if there are any required status checks or config files, or fails if there aren't


packages/mergeOnGreen/README.md
```diff
@@ -0,0 +1,21 @@
+mergeOnGreen
+
+Instructions are provided in [googleapis/repo-automation-bots](https://github.com/googleapis/repo-automation-bots/blob/master/README.md) for deploying and testing your bots.
+
+This bot uses nock for mocking requests to GitHub, and snap-shot-it for capturing responses; This allows updates to the API surface to be treated as a visual diff, rather than tediously asserting against each field.
+
+Running tests:
```
bcoe: I'm being very pedantic, but it would be nice to go back and make it so we generate a slightly nicer README out of the gate, an `h1` header is represented as `# My Header`, an `h2` as `## My Second Level Header`, I think we could make this look a bit better with these updates.

packages/mergeOnGreen/README.md
```diff
@@ -0,0 +1,21 @@
+mergeOnGreen
```
bcoe: In JavaScript land, the convention would be to clal this `merge-on-green`, rather than the camel case `mergeOnGreen` :+1:

packages/mergeOnGreen/src/index.d.ts
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+// define types for a few modules used by probot that do not have their
+// own definitions published. Before taking this step, folks should first
+// check whether type bindings are already published.
+declare module "promise-events" {
+  class EventEmitter {
+  }
+}
```
bcoe: we should update our templates to have a newline at the end of the file (which gets rid of this red mark at the end of the file); I know this drives @JustinBeckwith bonkers :laughing: 

As a goal, I'd like us to update our linters across the whole organization to catch this, so that it's not something that comes up in code review.

packages/mergeOnGreen/src/mergeOnGreen.ts
```diff
@@ -0,0 +1,118 @@
...
+    ],
+    async context => {
+      const config = (await context.config(
+        CONFIGURATION_FILE_PATH,
+        {}
+      )) as Configuration;
+
+      const { owner, repo } = context.repo();
+      
+      const data = await getBranchProtection(context.github, owner, repo, context.payload.pull_request.head.repo.default_branch);
```
bcoe: I'd probably call this something like `branchProtection`, vs., `data`.

packages/mergeOnGreen/src/mergeOnGreen.ts
```diff
@@ -0,0 +1,118 @@
...
+         output: {
+           title: 'You have no required status checks',
+           summary: 'Add required checks or make changes to the merge-on-green config file'
+         }
+       })
+     } 
+
+     await context.github.checks.create(checkParams);
+
+      // if (context.payload.check_run !== null) {
```
bcoe: get rid of these comments for now?

packages/mergeOnGreen/src/mergeOnGreen.ts
```diff
@@ -0,0 +1,118 @@
...
+      const commitParams = context.repo({
+        pull_number: context.payload.pull_request.number,
+        per_page: 100,
+      });
+
+      // Response object has a typed response.data, which has definitions that
+      // can be found here: https://unpkg.com/@octokit/rest@16.28.3/index.d.ts
+      let commitsResponse: Response<PullsListCommitsResponseItem[]>;
+      try {
+        commitsResponse = await context.github.pulls.listCommits(commitParams);
```
bcoe: eventually we'll want to be able to list commits I think, so we can check the status of the most recent commit and make sure all its checks are passing? for now though, i think we only need to call `getBranchProtection`.

packages/mergeOnGreen/src/mergeOnGreen.ts
```diff
@@ -0,0 +1,118 @@
...
+      const commits = commitsResponse.data;
+
+      let checkParams = context.repo({
+        name: 'merge-on-green-readiness',
+        conclusion: 'success' as Conclusion,
+        head_sha: commits[commits.length - 1].sha
+     })
+
+
+     if(!data && !config.required_status_checks) {
```
bcoe: I think this logic would be more along the lines of:

1. if there are no required status checks, enforce that branch protection is enabled, and perhaps a minimum of `N` (call it three :laughing:, I pulled that out of a hat) required checks are enabled.
2. if there are required status checks in the config, instead enforce that branch protection is enabled, and all of the required status checks are enabled on the repo.

packages/myTestCopy/test/myTestCopy.ts
```diff
@@ -0,0 +1,142 @@
...
+        './pull_request_synchronize'
+      ));
+      // create a history that has one valid commit, and one invalid commit:
+      const invalidCommits = require(resolve(fixturesPath, './invalid_commit'));
+      invalidCommits.push.apply(
+        invalidCommits,
+        require(resolve(fixturesPath, './valid_commit'))
+      );
+
+      const requests = nock('https://api.github.com')
```
bcoe: we can probably trim down this test quite a bit to only enforce the branch protection checks.

I would just pull the start of the commit checks onto a branch, so that you still have them as a starting point for future work.

packages/mergeOnGreen/src/mergeOnGreen.ts
sofisl: So, I think I need to call the commits in order to reference it to create a check-run; a check run object needs a head_sha, which I was just getting as the last commit. Let me know if you were commenting on something else/if I didn't understand the comment!

packages/myTestCopy/test/myTestCopy.ts
sofisl: ah, I didn't mean to commit myTestCopy to this branch - will remove. Thanks!

PR #218, opened 2020-01-10
feature: allow bots to have hyphens in google cloud functions

sofisl:
Related to Issue #177: after doing some digging with @bcoe, we found out that you can [publish a Google Cloud function](https://screenshot.googleplex.com/jJbAtUqiijw.png) with a hyphen. I'm recommending we make the change in this PR to provide naming consistency in the bots we create and deploy to GCF. In generate-bot, I've changed the logic so that module.exports can export a function with a hyphen, like so:

```
import` { GCFBootstrapper } from 'gcf-utils';
import appFn from './{{programName}}';

module.exports['{{program-name}}']= bootstrap.gcf(appFn);
```
`

which should work.

### DIFF
--- a/scripts/publish-functions.sh
+++ b/scripts/publish-functions.sh
@@ -43,9 +43,6 @@ for f in *; do
         (
         cd "$f" || exit 1
         echo "$f"
-        # Javascript does not allow function names with '-' so we need to
-        # replace the directory name (which might have '-') with "_"
-        functionname=${f//-/_}
         echo "About to publish function $functionname"
 
         gcloud functions deploy "$functionname" --trigger-http \
--- a/scripts/function-1.js
+++ b/scripts/function-1.js
@@ -1,0 +1,4 @@
+exports["hello-World"] = (req, res) => {
+    let message = req.query.message || req.body.message || 'Hello World!';
+    res.status(200).send(message);
+  };
--- a/scripts/function-1.js
+++ b/scripts/function-1.js
@@ -1,4 +1,0 @@
-exports["hello-World"] = (req, res) => {
-    let message = req.query.message || req.body.message || 'Hello World!';
-    res.status(200).send(message);
-  };


orthros:
Thanks for doing this! 

Just for my own clarification: this change (which removes the replacement of '-' with '_' in folder names) will make it so that future bots will be able to export their functions in the manner you provided above?

As far as I can tell, this would break other bots like `conventional-commit-lint` as there is now a discrepancy between the folder name and the exported function name, correct?

scripts/publish-functions.sh
```diff
@@ -43,9 +43,6 @@ for f in *; do
         (
         cd "$f" || exit 1
         echo "$f"
-        # Javascript does not allow function names with '-' so we need to
-        # replace the directory name (which might have '-') with "_"
-        functionname=${f//-/_}
```
orthros: Without `functionname` set, this will cause the `gcloud functions deploy "$functionname"...` to fail a few lines down, perhaps replace with `functionname=${f}`?

bcoe:
@sofisl @orthros it sounds like we'd want to move these legacy functions to a folder that uses `_` before we land this?

orthros:
@bcoe Either that or we move everything at once (fixing the function names in the process)

The problem with that is it would change the URL that the bots can be found at, and would then necessitate someone to fix the GitHub Application to point to the correct URL

bcoe:
@sofisl I think we're going to need to figure out a migration strategy for the culprit bots before we land this, or we're going to break quite a few things.

bcoe:
looking at the amount of work it would take to migrate all of our existing bots, and the benefit we'd get, I think we might do better @sofisl to move the logic for replacing `-` with `_` into `generate-bot`, rather than removing it from @orthros' scripts (at least for the time being)...

bcoe:
closing with tracking issue: https://github.com/googleapis/repo-automation-bots/issues/242

PR #220, opened 2020-01-10
fix: made fixtures folder name lowercase

sofisl:


bcoe:
looks like this work must have already been landed in another branch.

PR #223, opened 2020-01-13
fix: ensure header check ignores deleted files

### LINKED ISSUE #222: header-lint-check: Failing for a deleted file
See https://github.com/googleapis/google-auth-library-nodejs/pull/867/checks?check_run_id=387631481

When a file (`samples/iap.json`) was removed, the header lint check reports the file as having a missing header.

bshaffer:
fixes #222 - Ensures the header lint checker skips deleted files


feywind:
This looks fine to me, at a glance. The only thing that sticks out is that the URLs in the test fixtures are for something else (?) but I don't guess that matters much for the sake of that test.

bshaffer:
@feywind I made the URLs in the test fixtures `chingor13/google-auth-library-java` to match the other test fixture URLs!

PR #226, opened 2020-01-13
feature: first stab at merge-on-green

sofisl:
First stab at merge-on-green, using separate branch so as to not confuse w/ dev env

This PR:

(1) responds to PRs, identifies the owner, repo and branch for a specific PR
(2) identifies required status checks and/or config files with required status checks
(3) posts a status check that passes if there are any required status checks or config files, or fails if there aren't (also fails if there are less than 3 required status checks, or if status checks do not match up with config file)


packages/merge-on-green/README.md
```diff
@@ -0,0 +1,21 @@
+# merge-on-green
```
bcoe: this is looking good, we should make sure to upstream these template changes to your generator.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -0,0 +1,169 @@
...
+          }
+        });
+      }
+    }
+
+    await context.github.checks.create(checkParams);
+  });
+};
+
+//TODO: fail if config and no branch protection, ask user to enforce branch protection
```
bcoe: I think we can remove these TODO comments now?

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -0,0 +1,89 @@
...
+    );
+
+    it('gets branch information when a pull request is opened, responds with a status check', async () => {
+      const payload = require(resolve(
+        fixturesPath,
+        'events',
+        'pull_request_opened'
+      ));
+
+      // const scope = nockGetBranchProtection();
```
bcoe: we should probably get rid of this comment.

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -0,0 +1,89 @@
...
+      const scope = nock('https://api.github.com')
+        .log(console.log)
+        .get('/repos/testOwner/testRepo/contents/.github/merge-on-green.yml')
+        .reply(200, { content: config.toString('base64') })
+        .get('/repos/testOwner/testRepo/branches/master/protection')
+        .reply(200)
+        .get('/repos/testOwner/testRepo/pulls/6/commits?per_page=100')
+        .reply(200, commits)
+        .post('/repos/testOwner/testRepo/check-runs', body => {
+          snapshot(body);
```
bcoe: we should check in the snapshot we get from this call.

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -0,0 +1,89 @@
...
+      ));
+
+      // const scope = nockGetBranchProtection();
+
+      const scope = nock('https://api.github.com')
+        .log(console.log)
+        .get('/repos/testOwner/testRepo/contents/.github/merge-on-green.yml')
+        .reply(200, { content: config.toString('base64') })
+        .get('/repos/testOwner/testRepo/branches/master/protection')
+        .reply(200)
```
bcoe: It's probably worth adding a test for the bad case too.

packages/merge-on-green/README.md
sofisl: ah right, will do

PR #237, opened 2020-01-24
feat(buildcop): add support for Python

tbpg:
Python has a slightly different flavor of xUnit output. So, I updated
the parsing to match (testsuites and package name handling).

I switched the XML parsing to be compact. It makes iterating over the
results a bit cleaner to not have to worry about 'elements'. There is no
loss of accuracy for this purpose. The not-compact version only seems
needed if you need stable parsing and re-printing.

The Python test results also come in many different log files. So, I
changed the logic for checking when a test passed to be an explicit pass
rather than just not-fail. A possible downside of this is that if a
failing test is renamed, the corresponding issue will not be
automatically closed/renamed.


packages/buildcop/src/buildcop.ts
```diff
@@ -278,40 +291,52 @@ buildcop.formatFailure = (failure: TestFailure): string => {
...
+  const obj = xmljs.xml2js(xml, { compact: true }) as xmljs.ElementCompact;
+  const failures: TestCase[] = [];
+  const passes: TestCase[] = [];
+  // Python doesn't always have a top-level testsuites element.
+  let testsuites = obj['testsuite'];
+  if (testsuites === undefined) {
+    testsuites = obj['testsuites']['testsuite'];
+  }
+  // If there is only one test suite, put it into an array to make it iterable.
+  if (!isIterable(testsuites)) {
```
alexander-fenster: you can check `Array.isArray(testsuites)`, I'm sure there cannot be a real `Iterable` there.

tbpg: Perfect. Switched. Thanks!

packages/buildcop/src/buildcop.ts
```diff
@@ -278,40 +291,44 @@ buildcop.formatFailure = (failure: TestFailure): string => {
...
-  const obj = xmljs.xml2js(xml);
-  const failures: TestFailure[] = [];
-  for (const suites of obj.elements) {
-    if (suites.name !== 'testsuites') {
-      continue;
+buildcop.findTestResults = (xml: string): TestResults => {
+  const obj = xmljs.xml2js(xml, { compact: true }) as xmljs.ElementCompact;
+  const failures: TestCase[] = [];
+  const passes: TestCase[] = [];
+  // Python doesn't always have a top-level testsuites element.
```
bcoe: I'll be curious to see if we bump into similar edge-cases with Node.js.

tbpg: Me too. I imagine we will for all of the languages, which is why I included comments/tests for both languages.

tbpg:
Enabling for all of them (where desired) is the goal!

PR #247, opened 2020-01-28
feature: auto-labeling bot

sofisl:
This auto-labeling bot:

1) Pulls a storage file from GCP that matches repo names to GH labels, 
2) checks if the repo is in the list of repos
3) checks if that label already exists on the repo
4) checks if that label already exists on the issue, 
5) creates & applies a label


packages/auto-label/src/auto-label.ts
```diff
@@ -0,0 +1,256 @@
...
+const colorsData = JSON.parse(
+  fs.readFileSync(path.resolve(__dirname, '../../src/colors.json'), 'utf8')
+);
+
+interface JSONData {
+  num_total_prs: number;
+  num_open_p1s: number;
+  num_open_questions: number;
+  num_open_p0s: number;
+  language: string;
```
bcoe: In the future, you could be sneaky and only define the fields you care about in the interface.

packages/auto-label/src/auto-label.ts
```diff
@@ -0,0 +1,256 @@
...
+  srcFileName: string,
+  destFileName: string
+) {
+  const { Storage } = require('@google-cloud/storage');
+  const storage = new Storage();
+  const destFilename = path.resolve(destFileName);
+
+  const options = {
+    // The path to which the file should be downloaded, e.g. "./file.txt"
+    destination: destFilename,
```
bcoe: you can actually just download into memory like so:

```js
   // for mono-repos we populate a list of all apis and products,
    // since each repo might include multiple products:
    console.log(`populating all api labels for ${repoPath}`);
    const apis = await storage
      .bucket('devrel-prod-settings')
      .file('apis.json')
      .download();
    return JSON.parse(apis[0].toString()) as GetApiLabelsResponse;
```

:point_up: I had to crack open the reference docs, wasn't super well documented.

packages/auto-label/src/auto-label.ts
```diff
@@ -0,0 +1,256 @@
...
+    );
+    const githubLabel = objectInJsonArray.github_label;
+    const alreadyExists = await handler.checkExistingLabels(
+      context.github,
+      owner,
+      repo,
+      githubLabel
+    );
+
+    if (alreadyExists == null || alreadyExists === undefined) {
```
bcoe: I'd probably make this `===`, `null`, I try to avoid using `==` because JavaScript is too loose with it, it would work for `null`, `undefined`, false, 0, ''.

packages/auto-label/README.md
```diff
@@ -0,0 +1,21 @@
+auto-label
+
```
bcoe: we should get your PR landed that makes the README slightly prettier :laughing:, I think it just has a linting issue:

https://github.com/googleapis/repo-automation-bots/pull/230

packages/auto-label/src/app.ts
```diff
@@ -0,0 +1,24 @@
+<<<<<<< HEAD
+// Copyright 2020 Google LLC
+=======
+// Copyright 2019 Google LLC
+>>>>>>> d03795813fa2793da69d5429c734f08fb94e3e24
```
bcoe: looks like a merge conflict ended up in here and needs to be removed.

packages/auto-label/src/auto-label.ts
```diff
@@ -0,0 +1,264 @@
...
+};
+
+function handler(app: Application) {
+  app.on(['issues.opened', 'issues.reopened'], async context => {
+    const owner = context.payload.repository.owner.login;
+    const repo = context.payload.repository.name;
+    const issueId = context.payload.issue.number;
+
+    const jsonData = await handler.callStorage(
+<<<<<<< HEAD
```
bcoe: another little merge conflict, you can just manually delete these lines.

PR #251, opened 2020-01-29
fix: labels from first invocation were being applied to future calls

bcoe:
caching was interfering with our ability to return labels based on the context of the repo firing a webhook, this lead to the label for the first repo invoking the cloud function to be applied to subsequent calls.

### DIFF
--- a/packages/label-sync/src/label-sync.ts
+++ b/packages/label-sync/src/label-sync.ts
@@ -47,7 +47,7 @@ async function getLabels(github: GitHubAPI, repoPath: string): Promise<Labels> {
     await refreshLabels(github, repoPath);
   }
   const labels = {
-    labels: labelsCache.labels.slice(0)
+    labels: labelsCache.labels.slice(0),
   } as Labels;
   const apiLabelsRes = await handler.getApiLabels(repoPath);
   apiLabelsRes.apis.forEach(api => {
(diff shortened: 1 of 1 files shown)


packages/label-sync/src/label-sync.ts
```diff
@@ -42,11 +42,25 @@ interface Repos {
...
 // meaning the file running in cloud will be older than the one on master.
 let labelsCache: Labels;
-async function getLabels(github: GitHubAPI, repoPath: string) {
+async function getLabels(github: GitHubAPI, repoPath: string): Promise<Labels> {
   if (!labelsCache) {
     await refreshLabels(github, repoPath);
   }
-  return labelsCache;
+  const labels = {
+    labels: labelsCache.labels.slice(0),
```
JustinBeckwith: What's the purpose of a `labels.slice(0)`?

packages/label-sync/src/label-sync.ts
```diff
@@ -42,11 +42,25 @@ interface Repos {
...
+  } as Labels;
+  const apiLabelsRes = await handler.getApiLabels(repoPath);
```
JustinBeckwith: Little worried here. Isn't this essentially completely bypassing the cache?  Will this end up calling out to GCS constantly?

packages/label-sync/src/label-sync.ts
bcoe: this creates a copy of the `labels` so that we don't end up populating additional labels in the cache each time we access labels.

creating a cache means that the labels are cached between invocations of cloud functions.

packages/label-sync/src/label-sync.ts
bcoe: It's only called once per function invocation, which only occurs rarely, e.g., when a new repo is created, or a label is deleted.

I think we're going to need to get rid of the `push` functionality any ways, because it takes like an hour to backfill all labels, and the maximum execution time of a cloud function is 9 minutes.

PR #261, opened 2020-02-04
build: introducing GitHub actions for unit tests

bcoe:
To divide up load, and to make it easier to update the testing matrix, experimenting with GitHub actions for CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,15 @@ jobs:
           node-version: 12
       - run: npm install
       - run: npm run lint
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm install
+      - run: npm run docs-test
   coverage:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,38 @@
+on: [pull_request]
+name: ci
+jobs:
+  test:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
+on: [pull_request]
```
JustinBeckwith: I will admit to not fully understanding the difference between the `push` and `pull_request` events.  Do you know which one (or both?) we should be using?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm install
+      - run: npm test
+      - run: npm i codecov
+      - run: ./node_modules/.bin/codecov
+      - uses: codecov/codecov-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
JustinBeckwith: How are we managing these today?  I set it manually on the `gts` repo, but I'm hoping that isn't the answer as we try to fan this out :)

.github/workflows/ci.yml
bcoe: I think [pull_request](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#pull-request-event-pull_request) is what we want. If I'm reading properly it will fire on updates to a PR, but not no pushes to branches that have no associated PR.

JustinBeckwith: Do we want these tests to run on master after we merge?  I would imagine we want to keep those, right?  Was chatting with @chingor13 and he made the great point we need that for codecov to have something to anchor off of when deciding delta.

PR #263, opened 2020-02-04
build: codecov was finding the wrong JSON files

bcoe:
codecov was uploading a bunch of the wrong files, let's be more specific.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
       - run: npm install
       - run: npm test
       - run: npm i codecov
-      - run: ./node_modules/.bin/codecov
+      - run: ./node_modules/.bin/c8 report --reporter=text-lcov | ./node_modules/.bin/codecov --pipe
       - uses: codecov/codecov-action@v1
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,5 +41,4 @@ jobs:
           node-version: 12
       - run: npm install
       - run: npm test
-      - run: npm i codecov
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -42,7 +42,4 @@ jobs:
       - run: npm install
       - run: npm test
       - run: npm i codecov
```
JustinBeckwith: Wait a second - why do you need to install this here?  Instead, should we make sure it's in our package.jsons?

feywind: It seems like it'd be okay since it's the repo bot that's trying to run codecov, and not the individual repo `package.json`. I guess the place that'd be problematic is if we took the output of this process and ran other things that ended up depending on it.

JustinBeckwith: So technically, the `codecov` command line tool could have breaking changes from major version to major version.  I don't think it's likely, but it could happen.  I'd like to think we should either a.) include the dependency in package.json or b.) specify the version in the `npm i` call here.

If we go with `b`, may make sense to use `npx` too :)

.github/workflows/ci.yml
```diff
@@ -41,8 +41,4 @@ jobs:
           node-version: 12
       - run: npm install
       - run: npm test
-      - run: npm i codecov
-      - run: ./node_modules/.bin/codecov
-      - uses: codecov/codecov-action@v1
-        with:
-          token: ${{ secrets.CODECOV_TOKEN }}
+      - run: ./node_modules/.bin/c8 report --reporter=text-lcov | npx codecov@3 -t ${{ secrets.CODECOV_TOKEN }} --pipe
```
JustinBeckwith: I'm surprised you don't need to explicitly put the `with:` clause in here.  But if it works, 🤷‍♂

bcoe: It seems like we have access to the secrets set for the repo within the action. I think `with` is meant more for external actions, which specify the variables they expect using [inputs](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/metadata-syntax-for-github-actions#inputs).

PR #267, opened 2020-02-06
fix: ignore case when linting commit messages

### LINKED ISSUE #59: conventional-commit-bot: linting should be case insensitive

bcoe:
The Conventional Commit specification indicates:

"Any casing may be used, but it’s best to be consistent."

fixes #59


stephenplusplus:
Is there a way to force consistency?

bcoe:
talked about his offline, we'd rather err on the side of letting linting pass (as long as it works with the tooling) rather than putting any roadblocks in front of folks.

PR #269, opened 2020-02-06
build: push does not seem to fire on renovate PRs

bcoe:
let's see if `pull_request` is closer to what we were looking for, and if so make these the required checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,9 @@
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+      - *
 name: ci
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-on: [push]
+on: [push, pull_request]
```
JustinBeckwith: 😆 is this going to cause CI to fire twice on PRs though, like it was doing earlier?

bcoe: Yeah, which crumby behavior do we want? Feels like going back to a world where renovate PRs don't fire is pretty suboptimal.

PR #289, opened 2020-02-16
build: use package based CI

JustinBeckwith:
This is an experiment, and more a discussion starter than anything.  When working in this repository, I've run into the following problems:
- I end up with all kinds of `lerna` errors when running `npm install`
- Running `npm install` or `npm test` from the root takes a very long time, and it's increasing at a linear rate
- I frequently get compiler errors related to the `Application` class when I touched no such thing
- When something does fail in CI, I have to go digging through a ton of logs to find the problem

I look at these all as mono-repo struggles, and I wanted to try a different approach. 
- We run CI on a package by package basis, with no uber jobs
- We drop any explicit dependencies on `probot` or `@octokit/rest`, and instead rely on `gcf-utils` brining them in.  Yes I know this bucks conventional wisdom. 
- We ignore `package-lock.json`, as it was generating gobs of noise for no gain
- I introduced a simple script that lets you run commands in every package directory `npm run run`

I could be wrong about all of these things :) Also, I concede that I have no idea how releases work in this repository. 

Any thoughts on ways to make this better?


packages/failurechecker/package.json
```diff
@@ -27,8 +27,7 @@
   },
   "dependencies": {
     "gcf-utils": "1.6.1",
-    "moment": "^2.24.0",
-    "@octokit/rest": "16.41.2"
```
bcoe: if we're going to remove `@octokit/rest`, it should be added as a dependency to `gcf-utils`.

package-lock.json
```diff
@@ -1,7239 +0,0 @@
-{
-	"name": "repo-automation-bots",
-	"requires": true,
-	"lockfileVersion": 1,
-	"dependencies": {
```
bcoe: I'm open to getting rid of `package-lock.json` files, but I think in practice, given that we're deploying apps they would be a good idea (if we weren't seeing drift in TypeScript definitions so frequently).

.kokoro/continuous/node10/system-test.cfg
```diff
@@ -1,7 +0,0 @@
-# Download resources for system tests (service account key, etc.)
-gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/google-cloud-nodejs"
-
-env_vars: {
```
bcoe: If we're going to get rid of our standard system tests, we need to put an exception in sloth.

packages/auto-label/package.json
```diff
@@ -26,31 +26,27 @@
     "lint": "gts check"
   },
   "dependencies": {
-    "gcf-utils": "1.6.1",
-    "probot": "9.9.4",
-    "@octokit/rest": "16.41.2"
```
bcoe: if we were going to do this, I we should add coverage to each of the modules.

scripts/scripts.js
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Runs an arbitrary command in each directory in the `packages` dir.
+ * Example:
+ *    $ npm run run -- npm test
+ */
+async function main() {
+  const files = await fs.readdir('./packages', { withFileTypes: true });
+  const dirs = files.filter(f => f.isDirectory());
+  const cmd = process.argv.slice(2).join(' ');
```
bcoe: I wonder if we could use this same approach to continue having top level tests and coverage?

packages/failurechecker/package.json
JustinBeckwith: Oh I like that idea.  I thought it already was!

JustinBeckwith: It already was :)

package-lock.json
JustinBeckwith: In general, I would agree. But it's causing __pain__.

.kokoro/continuous/node10/system-test.cfg
JustinBeckwith: +1 - Should be an easy change in sloth when we're ready

JustinBeckwith:
So here's a thing.  I could get a top level `c8` to work, and it reported total end to end coverage.  Then I added `c8` to each individual package, but that seemed to cause the top level coverage to not report.  

Not sure how to get around that :) Right now I have it working for individual level packages.

bcoe:
@JustinBeckwith the problem is that every individual invocation of `c8` will reset the output directory, so you will only get reports for each individual library.

If you get things mostly working, I can take a look and fix the problem; we can make it so that `c8` accepts an environment variable that prevents it from deleting the output directory perhaps? and the top level test runner could set that when invoking the individual test suites.

cloudbuild.yaml
```diff
@@ -73,11 +73,13 @@ steps:
...
+      - "$PROJECT_ID"
+      - "$_BUCKET"
+      - "$_KEY_LOCATION"
+      - "$_KEY_RING"
       - "$_FUNCTION_REGION"
 # Each function can take ~2 minutes to deploy.
-timeout: 2400s
+# This means at 5 functions, we will run out
+# at the 10 minute mark. Increase to 20.
+timeout: 1200s
```
bcoe: just a bad merge?

JustinBeckwith: Crud, yes.  Apologies!  No idea why this even changed.

PR #298, opened 2020-02-19
fix: add client ID and secret to env vars

tbpg:
These seem to be required to run a bot locally. I got errors without them.


README.md
```diff
@@ -73,6 +73,8 @@ at https://github.com/settings/apps/{YOUR_APP}:
    created and downloaded at the bottom of the page.
 * `PRIVATE_KEY`: private key for application. This is the client secret near the top, `Client secret: a1bc23def4..........5678gh`
 * `WEBHOOK_SECRET`: secret key set in GitHub developer settings. Edit this to a known value in the settings page.
+* `GITHUB_CLIENT_ID`: client ID from the top of the page.
```
bcoe: I think these should only be needed if you're not pointing to a private key file, perhaps add a note that you can use these as an alternate to the key file?

tbpg: Good call. Done.

PR #303, opened 2020-02-19
fix(trusted-contribution): add gcf-merge-on-green[bot] as default trusted contributor

chingor13:
When merge-on-green bot updates the branch from master, it needs to retrigger ci


chingor13:
We probably should be looking at the last person to commit to the branch. In your case, yoshi-automation opened the PR, but merge-on-green updated it

PR #313, opened 2020-02-21
feat(Build Cop): add buildcop.go binary, remove buildcop.sh

tbpg:
gcloud runs into shell argument length limits. So, we need a binary.

I made a prototype using Python, but ran into dependency issues.

This Go binary is cross-compiled so it can work on the main 3 platforms.
I added a Makefile to make it easy to compile+upload.


packages/buildcop/buildcop.go
```diff
@@ -0,0 +1,208 @@
...
+
+	client, err := pubsub.NewClient(ctx, projectID, option.WithCredentialsFile(saPath))
+	if err != nil {
+		log.Printf("Unable to connect to Pub/Sub: %v", err)
+		return false
+	}
+	topic := client.Topic(topicID)
+
+	if repo == "" {
+		var ok bool
```
codyoss: nit: consider different name or don't use named return to avoid shadow and having to redeclare var name

packages/buildcop/buildcop.go
```diff
@@ -0,0 +1,208 @@
...
+		if err := filepath.Walk(artifactsDir, processLog(ctx, repo, installationID, topic)); err != nil {
+			log.Printf("Error publishing logs: %v", err)
+			return false
+		}
+	}
+	return true
+}
+
+// detectRepo tries to detect the repo from the environment.
+func detectRepo() (repo string, ok bool) {
```
codyoss: I don't think "" has special meaning for these, do you need to return a bool as well?

packages/buildcop/buildcop.go
tbpg: Removed the pre-declaration. Want the named return for documentation purposes.

packages/buildcop/buildcop.go
tbpg: Removed the bool.

PR #326, opened 2020-02-25
fix: add logic to check if PR has been merged

sofisl:
Added logic to see if the PR has been merged before trying to merge

### DIFF
--- a/packages/merge-on-green/src/merge-logic.ts
+++ b/packages/merge-on-green/src/merge-logic.ts
@@ -509,6 +509,8 @@ mergeOnGreen.checkPRMerged = async function checkPRMerged(
     });
     return true;
   } catch (err) {
+    console.log(err);
+    console.log('PR was not merged');
     return false;
   }
 };
--- a/packages/merge-on-green/src/merge-logic.ts
+++ b/packages/merge-on-green/src/merge-logic.ts
@@ -105,6 +105,9 @@ mergeOnGreen.getPR = async function getPR(
       repo,
       pull_number: pr,
     });
+    console.log(
+      `github's determination of whether the ${owner}/${repo} is mergeable: ${data.data.mergeable}`
+    );
     return data.data;
   } catch (err) {
     return { title: '', body: '' };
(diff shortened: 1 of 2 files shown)


packages/merge-on-green/src/merge-logic.ts
```diff
@@ -495,6 +495,24 @@ mergeOnGreen.commentOnPR = async function commentOnPR(
...
+  github: GitHubAPI
+): Promise<boolean> {
+  try {
+    await github.pulls.checkIfMerged({
+      owner,
+      repo,
+      pull_number: pr,
+    });
+    return true;
+  } catch (err) {
```
bcoe: the fact that you're getting a `405` error seems odd to me, I would expect this to return a `true` or `false`, is the response of this function documented somewhere?

packages/merge-on-green/src/merge-logic.ts
```diff
@@ -495,6 +495,26 @@ mergeOnGreen.commentOnPR = async function commentOnPR(
...
 };
 
+mergeOnGreen.checkPRMerged = async function checkPRMerged(
+  owner: string,
+  repo: string,
+  pr: number,
+  github: GitHubAPI
+): Promise<boolean> {
+  try {
+    await github.pulls.checkIfMerged({
```
tbpg: Does this function raise an error if the PR is not merged? Seems like it should return a boolean-type value? Docs aren't great for this: https://octokit.github.io/rest.js/v17#pulls-check-if-merged.

packages/merge-on-green/src/merge-logic.ts
```diff
@@ -504,6 +524,11 @@ export async function mergeOnGreen(
   github: GitHubAPI
 ): Promise<boolean | undefined> {
   console.info(`${owner}/${repo} checking merge on green PR status`);
+  const isMerged = await mergeOnGreen.checkPRMerged(owner, repo, pr, github);
+  if (isMerged) {
+    console.log('This PR has already been merged');
```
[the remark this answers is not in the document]
sofisl: Good catch, thanks!

packages/merge-on-green/src/merge-logic.ts
sofisl: Yeah, it's a bit strange.... Here's the best version of the documentation I could find:
https://developer.github.com/v3/pulls/#get-if-a-pull-request-has-been-merged

I've tested it myself as well and that seems to be the functionality. Ben recommended we print out the error just so we're aware and not just assuming that it means the PR has not been merged. I agree though, it's a strange method :)

tbpg: That's surprising! But, sounds good!

PR #331, opened 2020-02-25
chore: slight logging improvements

bcoe:
some small tweaks to logging to make debugging easier.

### DIFF
--- a/packages/merge-on-green/src/merge-logic.ts
+++ b/packages/merge-on-green/src/merge-logic.ts
@@ -106,7 +106,7 @@ mergeOnGreen.getPR = async function getPR(
       pull_number: pr,
     });
     console.log(
-      `github's determination of whether the ${owner}/${repo} is mergeable: ${data.data.mergeable}`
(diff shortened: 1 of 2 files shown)


packages/merge-on-green/src/merge-logic.ts
```diff
@@ -492,7 +492,7 @@ mergeOnGreen.commentOnPR = async function commentOnPR(
       owner,
       repo,
       issue_number: pr,
-      body: `Your ${owner}/${repo} PR ${pr} was not mergeable because either one of your required status checks failed, or one of your required reviews was not approved.`,
```
sofisl: I think it might be helpful to leave in which owner/repo/pr the messages are referring to where possible, unless there's a reason not to?

bcoe: my thinking was that the comment shows up on the given PR, on the given repo/organization, so thought it might feel redundant?

sofisl: you are totally right!

PR #332, opened 2020-02-25
fix: add native protection config

sofisl:
Adding option for repos to use their native branch protection without having to list out the protection


packages/merge-on-green/src/merge-logic.ts
```diff
@@ -214,6 +229,17 @@ mergeOnGreen.getRequiredChecks = async function getRequiredChecks(
           console.log(
             `Your language's required checks were overridden because of the repo ${owner}/${repo}`
           );
+          if (
+            isOverriden.requiredStatusChecks[0] === 'Native branch protection'
```
bcoe: rather than making this a string, part of me would rather this be an additional parameter on overrides, something like:

```
repoOverrides: [
  {
    repo: 'googleapis/blerg',
    useBranchProtectionRules: true
]
```

sofisl: I like it!!

packages/merge-on-green/src/merge-logic.ts
```diff
@@ -52,6 +52,7 @@ interface RequiredChecksByLanguage {
 interface RepoOverrides {
   repo: string;
   requiredStatusChecks: string[];
+  useBranchProtectionRules: boolean;
```
bcoe: I would make this optional, since it might not be populated:

`useBranchProtectionRules?: boolean`.

Issue #276, opened 2020-02-11
python-docs-samples missing API labels


JustinBeckwith:
Was trying to do some triage over here:
https://github.com/GoogleCloudPlatform/python-docs-samples/labels

Noticed that none of the `api: ` labels are getting automatically created for that repository.  Not sure why :)

JustinBeckwith:
Turns out this was a me problem.  I hadn't enabled the bot on the repository 🤦‍♂ I need a bot to help manage all these bots!

PR #336, opened 2020-02-26
chore: clean up comments in label-sync

JustinBeckwith:
I was trying to debug #276, and was struggling to follow along with the code.  Made a few quality of life improvements, but mostly added some docs to help make it clear what was happening.


packages/label-sync/src/label-sync.ts
```diff
@@ -38,18 +38,37 @@ interface Repos {
...
 let labelsCache: Labels;
 async function getLabels(github: GitHubAPI, repoPath: string): Promise<Labels> {
   if (!labelsCache) {
-    await refreshLabels(github, repoPath);
+    await refreshLabels(github);
   }
   const labels = {
     labels: labelsCache.labels.slice(0),
   } as Labels;
-  const apiLabelsRes = await handler.getApiLabels(repoPath);
```
bcoe: this method is on `handler` so that it can easily be mocked in tests, setting up 75 nock calls is a pain in the butt for simple methods like this.

bcoe:
👋 can we close this as a bit stale? the main blocker for landing is the `.handler` being used in tests as a way of mocking behavior.

Issue #312, opened 2020-02-21
Merge on Green: add the PR number in the merge commit


chingor13:
When you normally merge a PR, the default merge commit made appends the pull request number in the message. This makes it easier to click to the pull request when browsing the commit history.

When the bot merges, it would be nice if the merge commit was `<PR_TITLE> (#<PR_NUMBER>)`.

PR #342, opened 2020-02-26
feat: add update branch logic that checks 'mergeable' state

### LINKED ISSUE #309: Merge-on-green bug: see if we can check if a branch needs updating before updating
Instead of attempting to merge the branch before updating it, we should check to see if the most recent commit on master is in the commits of your branch, else it's not up to date.

sofisl:
Fixes #309  🦕
Also, sneaking in a quick fix to #312 also :)

### DIFF
--- a/packages/merge-on-green/src/merge-logic.ts
+++ b/packages/merge-on-green/src/merge-logic.ts
@@ -32,11 +32,6 @@ interface Reviews {
   state: string;
 }
 
-interface ReviewsRequested {
-  users: [];
-  teams: [];
-}
-
 interface PullRequest {
   title: string;
   body: string;
@@ -476,7 +471,7 @@ mergeOnGreen.merge = async function merge(
       owner,
       repo,
       pull_number: pr,
-      commit_title: prInfo.title,
+      commit_title: `${prInfo.title} (#${pr})`,
       commit_message: prInfo.body,
       merge_method: 'squash',
     })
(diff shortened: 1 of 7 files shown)


packages/merge-on-green/src/merge-logic.ts
```diff
@@ -106,12 +104,9 @@ mergeOnGreen.getPR = async function getPR(
       repo,
       pull_number: pr,
     });
-    console.log(
-      `github's determination of whether the ${owner}/${repo}/${pr} is mergeable: ${data.data.mergeable} ${data.data.mergeable_state}`
-    );
     return data.data;
   } catch (err) {
-    return { title: '', body: '' };
+    return { title: '', body: '', mergeable: false, mergeable_state: '' };
```
bcoe: I like this step, where if you couldn't get the data you indicate it's not `mergeable`.

packages/merge-on-green/src/merge-logic.ts
```diff
@@ -559,11 +555,16 @@ export async function mergeOnGreen(
...
   }
 
-  const [checkReview, checkStatus] = await Promise.all([
+  const [checkReview, checkStatus, prInfo] = await Promise.all([
     mergeOnGreen.checkReviews(owner, repo, pr, github),
     mergeOnGreen.statusesForRef(owner, repo, pr, labelName, github),
+    mergeOnGreen.getPR(owner, repo, pr, github),
   ]);
 
+  const failedMesssage = `Your PR was not mergeable because either one of your required status checks failed, or one of your required reviews was not approved. See required reviews for your repo here: https://github.com/googleapis/sloth/blob/master/required-checks.json`;
```
bcoe: great idea adding this context :+1:

PR #345, opened 2020-02-26
fix: parallelize merge on green logic, so that we work our way through PRs faster

sofisl:

### DIFF
--- a/packages/merge-on-green/src/merge-on-green.ts
+++ b/packages/merge-on-green/src/merge-on-green.ts
@@ -20,6 +20,7 @@ const TABLE = 'mog-prs';
 const datastore = new Datastore();
 const MAX_TEST_TIME = 1000 * 60 * 60 * 6; // 2 hr.
 const MERGE_ON_GREEN_LABEL = 'automerge';
+const WORKER_SIZE = 4;
 
 interface WatchPR {
   number: number;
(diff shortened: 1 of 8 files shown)


packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -33,24 +33,57 @@ handler.listPRs = async function listPRs(): Promise<WatchPR[]> {
   const query = datastore.createQuery(TABLE).order('created');
   const [prs] = await datastore.runQuery(query);
   const result: WatchPR[] = [];
-  for (const pr of prs) {
-    const created = new Date(pr.created).getTime();
+  for (let x = 0; x < prs.length; x += 5) {
+    if (prs[x + 1] === undefined) {
```
bcoe: I think you can get most of this same behavior with a few less lines of code by doing something like this:

```js
const WORK_SIZE = 4;
while (prs.length) {
  const work = prs.splice(0, WORK_SIZE);
}
```

:point_up:  and I would do this in the handler method `  app.on(['schedule.repository'], async context => {` rather than in the listing of work.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -80,38 +81,39 @@ handler.addPR = async function addPR(wp: WatchPR, url: string) {
...
+              wp.repo,
+              wp.number,
+              MERGE_ON_GREEN_LABEL,
+              wp.state,
+              context.github
+            );
+            if (remove || wp.state === 'stop') {
+              handler.removePR(wp.url);
+            }
+          } catch (err) {
```
bcoe: `finally`, instead of `catch`, and will run for both.

PR #365, opened 2020-03-02
fix: make sure sync-repo-settings is not included in deploy

sofisl:


bcoe:
we will remove this exception as soon as we have the settings sync bot deployed.

PR #367, opened 2020-03-03
build: ignore kokoro and release for synth

JustinBeckwith:


synth.py
```diff
@@ -7,12 +7,16 @@
...
   '.prettierignore',
   '.prettierrc',
-  '.nycrc',
+  'codecov.yaml',
   'README.md',
-  '.github/release-please.yml',
-  '.kokoro/',
-  '.github/workflows/',
-  'codecov.yaml'
+  'renovate.json'
```
chingor13: We don't want our standard renovate config? Or are there special configs in the templated versions that don't apply here?

synth.py
```diff
@@ -7,12 +7,16 @@
 s.copy(templates, excludes=[
   '.eslintignore',
   '.eslintrc.yml',
+  '.github/release-please.yml',
+  '.github/workflows/',
+  '.github/publish.yml',
+  '.kokoro/*.*',
+  '.kokoro/**/*.*',
```
bcoe: did some testing, this can just be the glob `.kokoro/**`.

synth.py
JustinBeckwith: We are supporting node.js *and* Go in this repo, so it has some customizations.  We're keeping the file, just not auto-updating.

synth.py
JustinBeckwith: Great catch!  I can't believe I missed that.

PR #368, opened 2020-03-03
chore: add p3 label

kolea2:
I'm looking for suggestions for the description, and if we need to update anything else. PTAL :)

### DIFF
--- a/packages/label-sync/src/labels.json
+++ b/packages/label-sync/src/labels.json
@@ -57,17 +57,17 @@
     },
     {
       "name": "priority: p3",
-      "description": "",
+      "description": "Desirable enhancement or fix. May not be include in next release.",
       "color": "ffffc7"
     },
     {
       "name": "priority: p2",
-      "description": "Lowest priority. Fix may not be included in next release.",
+      "description": "Moderately-important priority. Fix may not be included in next release.",
       "color": "fef2c0"
     },
     {
       "name": "priority: p1",
-      "description": "Medium priority. Will be fixed prior to next release.",
+      "description": "Important issue which blocks shipping the next release. Will be fixed prior to next release.",
       "color": "ffa03e"
     },
     {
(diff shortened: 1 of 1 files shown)


packages/label-sync/src/labels.json
```diff
@@ -55,6 +55,11 @@
       "description": "Request for information or clarification. Not an issue.",
       "color": "c5def5"
     },
+    {
+      "name": "priority: p3",
+      "description": "",
```
crwilcox: I may suggest that we remove the 'highest, medium, lowest' from others, and update descriptions a bit. For this one, perhaps "Desirable enhancement or fix. May not be include in next release"

packages/label-sync/src/labels.json
```diff
@@ -55,6 +55,11 @@
...
+      "color": "ffffc7"
+    },
     {
       "name": "priority: p2",
       "description": "Lowest priority. Fix may not be included in next release.",
```
crwilcox: P2 Current: Lowest priority. Fix may not be included in next release.
P2 Suggest: Moderately-important Issue. Fix may not be included in next release.

P1 Current: Medium priority. Will be fixed prior to next release.
P1 Suggest: Important issue which blocks shipping the next release. Will be fixed prior to next release.

P0 Current: Highest priority. Critical issue. P0 implies highest priority.
P0 Suggest: -

kolea2:
Thanks @crwilcox! Updated based on your suggestions.

packages/label-sync/src/labels.json
```diff
@@ -55,14 +55,19 @@
       "description": "Request for information or clarification. Not an issue.",
       "color": "c5def5"
     },
+    {
+      "name": "priority: p3",
+      "description": "Desirable enhancement or fix. May not be include in next release.",
```
crwilcox:
```suggestion
      "description": "Desirable enhancement or fix. May not be included in next release.",
```

kolea2: oops sorry, missed this. I'll create a typo fix

crwilcox: sgtm. thanks!

PR #422, opened 2020-03-28
chore: get gcf-utils and release-please on latest typescript/probot/gts

bcoe:
I'm a little worried about this one, as [probot is pinned on @octokit/rest at 16.x.x](https://github.com/probot/probot/pull/1152), and effort has been made to get release-please on the latest @octokit/rest@17.x.x.

I don't think that we rely on any of the breaking API changes between the two interfaces, but there's a chance we end up needing to rollback octokit in release-please.

### DIFF
--- a/packages/gcf-utils/package.json
+++ b/packages/gcf-utils/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gcf-utils",
-  "version": "1.6.2",
+  "version": "2.0.0-candidate.0",
   "description": "An extension for running Probot in Google Cloud Functions",
   "scripts": {
     "compile": "tsc -p .",
--- a/packages/gcf-utils/package.json
+++ b/packages/gcf-utils/package.json
@@ -22,8 +22,9 @@
     "@google-cloud/kms": "^1.6.3",
     "@google-cloud/storage": "^4.0.0",
     "@octokit/plugin-enterprise-compatibility": "1.2.2",
+    "express": "^4.17.1",
     "gaxios": "^3.0.0",
-    "probot": "^9.9.4",
+    "probot": "^9.11.2",
     "tmp": "^0.1.0",
     "yargs": "^15.0.0"
   },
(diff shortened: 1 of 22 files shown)


packages/gcf-utils/package.json
```diff
@@ -22,8 +22,9 @@
     "@google-cloud/kms": "^1.6.3",
     "@google-cloud/storage": "^4.0.0",
     "@octokit/plugin-enterprise-compatibility": "1.2.2",
+    "express": "^4.17.1",
```
JustinBeckwith: How did express slip in here?

packages/gcf-utils/package.json
```diff
@@ -42,11 +43,11 @@
     "c8": "^7.1.0",
     "cross-env": "^7.0.0",
     "dotenv": "^8.0.0",
-    "gts": "^1.1.2",
+    "gts": "^2.0.0-alpha.8",
```
JustinBeckwith: Would it be possible to de-tangle the TypeScript/gts upgrades from probot?  If we need to roll back ....

packages/release-please/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "gts/tsconfig-google",
   "compilerOptions": {
+    "lib": ["es2018", "dom"],
```
JustinBeckwith: I don't think we need to do this explicitly 🤷‍♂

packages/gcf-utils/package.json
bcoe: Not a slip, we were relying on `express` as a transitive dependency in `gcf-utils`. The new rules are, in my opinion, rightfully stricter about requiring that we actually define our dependencies.

There's risk involved in relying on a transitive dependency, as you don't have guarantees about what your top level library actually has access to.

packages/gcf-utils/package.json
bcoe: If. we end up needing to rollback, I'll make a hot fix for the couple changes that were necessary in `release-please`. I'm not concerned about the probot upgrade, so much as the octokit update, which is only in one place.

bcoe:
Landing, with the understanding that there's a good chance we rollback part of this PR.

PR #431, opened 2020-04-01
fix: sync repo settings changes

sofisl:


packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -74,27 +75,21 @@ function handler(app: Application) {
...
-    try {
-      await handler.updateRepoTeams(r, context);
-    } catch (err) {
-      console.log(`Error updating repo in org for ${repo}`);
-      console.log(err);
-    }
+    const start = new Date().getTime();
+    // update each settings section
+    await Promise.all([
+      await handler.updateRepoOptions(r, context),
```
bcoe, line 86: when you do `Promise.all` you don't need the await, this would actually still be sequential.

Issue #107, opened 2019-10-04
header-checker-lint: should look for configuration in .github


bcoe:
it looks for configuration in `.bots` currently.

PR #432, opened 2020-04-02
refactor: use the built-in .github config

bcoe:
fixes: https://github.com/googleapis/repo-automation-bots/issues/107


jpoehnelt:
It appears the old version of the bot is still running. What is the release plan for this given it breaks existing users?

chingor13:
@jpoehnelt Sorry, we missed this comment. It's like we broke your config - we assumed we were the only consumers of this bot at the time. Is it still broken for you?

PR #439, opened 2020-04-07
fix(sync-repo-settings): add logic to ignore certain repos, add tests

sofisl:


packages/sync-repo-settings/src/required-checks.json
```diff
@@ -109,10 +109,6 @@
       "googleapis/google-cloud-node"
     ],
     "repoOverrides": [
-      {
-        "repo": "GoogleCloudPlatform/nodejs-docs-samples",
```
bcoe, line 113: why did we delete this entry, wondering if it should be added back?

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -70,10 +70,18 @@ function handler(app: Application) {
...
-    if (!r) {
+    const yoshiRepo = repos.repos.find(x => x.repo === repo);
+    if (!yoshiRepo) {
       return;
     }
+    if (languageConfig[yoshiRepo.language]) {
+      const ignored = languageConfig[yoshiRepo.language].ignoredRepos?.find(
+        x => x === repo
+      );
+      if (ignored) {
```
bcoe, line 81: might be worth adding a log message:

```
ignoring repo "foo".
```

packages/sync-repo-settings/src/required-checks.json
sofisl: Hey! I removed it because the rule is just to add in the branch protection that's already there. So, it would be the same as not updating it, I assume? Since we're not actually ever getting the native branch protection of any of the repos. Since the repo is also in the 'ignored repos' section, we really don't need to do anything to it.

bcoe: oh, as long as it's in ignored repos sounds good to me 👌

PR #452, opened 2020-04-10
chore(Build Cop): refactor tests to reduce duplication

### LINKED ISSUE #448: Build Cop: reduce test duplication
We duplicate a bunch of `nock` code in our Build Cop Bot tests. We should follow the pattern of sync-repo-settings and factor common stuff into functions:

https://github.com/googleapis/repo-automation-bots/blob/bac037022b03a4be280d7a2a8d67d718100884e2/packages/sync-repo-settings/test/test.ts#L130-L138

This is essentially a blocker for additional feature development. The BCB tests are getting unruly.

tbpg:
I also replaced `tbpg` with `GoogleCloudPlatform`. `java-vision` is not in `GoogleCloudPlatform`, but that's OK.

Fixes #448.

### DIFF
--- a/packages/buildcop/test/buildcop.ts
+++ b/packages/buildcop/test/buildcop.ts
@@ -21,34 +21,25 @@ import snapshot from 'snap-shot-it';
 import nock from 'nock';
 import * as fs from 'fs';
 import { expect } from 'chai';
-import Octokit from '@octokit/rest';
 
 nock.disableNetConnect();
 
 const fixturesPath = resolve(__dirname, '../../test/fixtures');
-
-function formatPayload(payload: BuildCopPayload) {
-  if (payload.xunitXML) {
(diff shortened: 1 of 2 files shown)


packages/buildcop/test/buildcop.ts
```diff
@@ -26,11 +26,67 @@ nock.disableNetConnect();
...
+    commit: '123',
+    buildURL: 'http://example.com',
+    xunitXML: Buffer.from(input).toString('base64'),
+  };
+}
+
+// Ignore any warning to make it easier to test. No need to include all of the
+// fields from real issues.
+// tslint:disable-next-line: no-any
+function nockIssues(repo: string, issues: any[] = []) {
```
JustinBeckwith, line 48: If you aren't particular about the type passed in, use `{}` instead of `any`, and you can drop the linter disable.

tbpg: Done. Thanks!

PR #453, opened 2020-04-10
fix: revert try-catch errors

sofisl:


bcoe:
whoops, landed the revert bit instead, we'll figure this out soon in a follow up.

PR #469, opened 2020-04-13
fix: address bugs with tasks logic found after deployment

bcoe:
A few things that needed to be corrected:

1. logic for enqueuing fails unless `content-type` set to `application/json`.
2. status needs to be set in response (or else it's always a `200`).
3. `request.body` was already a string, so we were double serializing.


packages/gcf-utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "gcf-utils",
-  "version": "3.0.1",
+  "version": "3.0.3",
```
JustinBeckwith, line 3: That's interesting.  Are we manually bumping versions in this package?

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -104,17 +104,16 @@ export class GCFBootstrapper {
         request.get('x-github-delivery') ||
         request.get('X-GitHub-Delivery') ||
         '';
-      const delivery =
-        request.get('x-hub-signature') || request.get('X-Hub-Signature') || '';
+      // TODO: add test to validate this signature is used on the initial
```
JustinBeckwith, line 107: Can we up-level this to a GitHub issue so we don't lose it?

packages/gcf-utils/package.json
bcoe: yes, release-please doesn't know how to deal with this package structure. I think the best solution would be to extract `gcf-utils` form the mono-repo.

packages/gcf-utils/src/gcf-utils.ts
bcoe: done: https://github.com/googleapis/repo-automation-bots/issues/470

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -104,17 +104,14 @@ export class GCFBootstrapper {
...
+      // webhook from GitHub:
       const signature =
-        request.get('x-github-delivery') ||
-        request.get('X-GitHub-Delivery') ||
-        '';
+        request.get('x-hub-signature') || request.get('X-Hub-Signature') || '';
```
orthros, line 110: I'm not sure this is correct. Consider these lines
https://github.com/probot/serverless-gcf/blob/master/index.js#L39-L50. I believe the "signature" does need to be "x-github-delivery" and we just named things poorly.

bcoe: works for me, I will address.

bcoe: @orthros okay, reverted the header, @sofisl is testing everything out.

PR #472, opened 2020-04-13
build(WIP)!: update to latest version of gts/typescript

AVaksman:
BREAKING CHANGE: typescript@3.7 introduced some breaking changes


JustinBeckwith:
I think this already happened in #457 😬

Issue #433, opened 2020-04-03
Merge on Green: Bot commented twice after 3 hours


bshaffer:
Merge on Green `automerge` label commented twice with the same comment at the same time:

https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1044#issuecomment-608091697

Also, strangely enough, the comment is different than the one that commented 3 hours later... so maybe there's an old version running somewhere?

sofisl:
I think this should be solved by PR #475. Please let me know if you run into it again!

PR #475, opened 2020-04-13
fix(merge-on-green): bug that comments on PR multiple times

sofisl:
Addresses issue #433 by limiting the interval in which the bot can comment


packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -470,7 +470,7 @@ describe('merge-on-green-', () => {
               repo: 'testRepo',
               number: 1,
               owner: 'testOwner',
-              created: Date.now() - 10800000,
+              created: Date.now() - 10920000,
```
JustinBeckwith, line 473: I kinda have to guess what that number means - would be great to add a comment explaining it's (looks at calculator) a little over 3 hours?

sofisl: Ah, will do! Basically, it limits the time interval in which it comments to 4 minutes, which is the period between runs for cloud scheduler. This *approximates* that it will comment only once, but I'll add some comments!

PR #480, opened 2020-04-14
feat(gcf-utils)!: Use Secret Manager as opposed to KMS + GCS

orthros:
Now uses SecretManager to store Probot configs. This obviates the DIY
solution we had for storing these secrets.


orthros:
Overall, this is functional, but needs to have the `bin` script updated to be simpler (and not use KMS and GCS). Will be looking at that next

packages/gcf-utils/src/bin/genkey-util.ts
```diff
@@ -0,0 +1,63 @@
...
+  webhookSecret: string
+): Promise<Options> {
+  let keyContent = '';
+  try {
+    keyContent = fs.readFileSync(keyfile, 'utf8');
+  } catch (e) {
+    throw Error(`Error reading file: ${keyfile}`);
+  }
+
+  const blob: Options = {
```
bcoe, line 31: I think you could make this a one liner with:

```js
return {
    cert: keyContent,
    id,
    secret: webhookSecret,
}  as Options
```

packages/gcf-utils/test/gcf-utils.ts
```diff
@@ -212,4 +214,98 @@ describe('GCFBootstrapper', () => {
...
   });
+
+  describe('getProbotConfig', () => {
+    let bootstrapper: GCFBootstrapper;
+    let secretClientStub: v1.SecretManagerServiceClient;
+    let secretsStub: sinon.SinonStub;
+    let secretVersionNameStub: sinon.SinonStub;
+
+    beforeEach(() => {
+      secretClientStub = new v1.SecretManagerServiceClient();
```
bcoe, line 225: I just realized that @alexander-fenster's refactor to make gRPC delay initialization made stubbing _way_ easier, this is great.

orthros: Yep! It is so so so nice to use

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -57,36 +61,25 @@ export class GCFBootstrapper {
...
-      ciphertext: contentsBuffer,
+  async getProbotConfig(): Promise<Options> {
+    const name = this.getLatestSecretVersionName();
+    const [version] = await this.secretsClient.accessSecretVersion({
+      name: name,
     });
-
-    const config = JSON.parse(result.plaintext!.toString());
+    // Extract the payload as a string.
+    const payload = version?.payload?.data?.toString() || '';
```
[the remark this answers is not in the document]
orthros: Done; also added some tests for it

packages/gcf-utils/src/bin/genkey-util.ts
orthros: Done

orthros:
Latest commits add a new script to migrate from GCS to SecretManager. This script will not work right now, however as `publish` bot has a secret already named "publish" in our project. 

Once that secret has been migrated, we can run the script and merge this PR.

Issue #383, opened 2020-03-12
Merge on Green: handle cases where the branch protection doesn't quite match the statuses


sofisl:
Example:

https://github.com/googleapis/google-api-php-client/pull/1796

MOG won't merge here because the branch protection is continuous-integration/travis-ci, and the status checks are continuous-integration/travis-ci/pr and continuous-integration/travis-ci/push

bshaffer:
The convention for branch checks is either `[check-name]` or `[check-name]/[subcheck-name]`. Would it be possible to update the bot so that the protected branch checks will match for either of these?

bshaffer:
Let me clarify my above comment. If a branch protection check matches either `[check]` or all `[check]/*`, then Merge on Green can merge.

PR #482, opened 2020-04-15
fix(merge-on-green): bug that won't let subtests pass by adding in regex

sofisl:
Fixes: Issue #383 by adding in regular expression matching. The code will check if a status check that has run matches the beginning of the required status check, with additions to that string allowed. This regular expression matching accounts for subtests, like the ones mentioned in the issue.


packages/merge-on-green/src/merge-logic.ts
```diff
@@ -370,8 +372,9 @@ mergeOnGreen.statusesForRef = async function statusesForRef(
         `Looking for required checks in status checks for ${owner}/${repo}/${pr}.`
       );
       //since find function finds the value of the first element in the array, that will take care of the chronological order of the tests
-      const checkCompleted = checkStatus.find(
-        (element: CheckStatus) => element.context === check
+      const regexCheck = new RegExp(`^${check}`);
+      const checkCompleted = checkStatus.find((element: CheckStatus) =>
+        regexCheck.test(element.context)
```
tbpg, line 377: I've seen this message in the logs:

```
Cannot read property 'context' of undefined
```

Perhaps we need a filter/check on `element`?

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -324,6 +324,52 @@ describe('merge-on-green-', () => {
...
 
+    it.only('rejects status checks that do not match the required check', async () => {
+      const scopes = [
+        getPR(true, 'clean', 'open'),
+        getBranchProtection(["this is what we're looking for"]),
+        getReviewsCompleted([{user: {login: 'octocat'}, state: 'APPROVED'}]),
+        getLatestCommit([{sha: '6dcb09b5b57875f334f61aebed695e2e4193db5e'}]),
+        getMogLabel([{name: 'automerge'}]),
+        getStatusi('6dcb09b5b57875f334f61aebed695e2e4193db5e', [
+          {state: 'success', context: "this is what we're looking fo"},
```
tbpg, line 335:
```suggestion
          {state: 'success', context: "this is what we're looking for"},
```

Or does this intentionally not match? Might want to change to "not what we're looking for" if that's the case.

sofisl: So, this is an interesting test. It intentionally does not match, but I wanted to make the string as close to the test it is checking as possible (minus one character) to check the directionality of the regex (i.e., only if the status check matches the first part of the required status check will it pass, otherwise it will fail). Of course, it's regex, so it could be an entirely different string and it won't pass either. Changing the string to "not what we're looking for" would make more sense, but then in that case it would be like the regular status check test that fails if the status check that runs does not match the required status check. I feel like this is a bit of a ramble, but hopefully it makes sense. With this explanation in mind, what would you recommend? We could take out this test entirely, given that it's pretty much covered by other tests that check to see if the required status checks ran and passed.

packages/merge-on-green/src/merge-logic.ts
sofisl: Hey! So I'm pretty sure that that message is caused by the 'iterateGetStatusi' function, which basically calls up to 10 pages of the status checks, even if they're not there. Currently, it's actually part of the functionality of the bot, but I agree we should have a better way of handling it. I created issue #487 to track it!

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -324,6 +324,52 @@ describe('merge-on-green-', () => {
       scopes.forEach(s => s.done());
     });
 
+    it.only('rejects status checks that do not match the required check', async () => {
```
bcoe, line 327: we should remove this `.only` so that the full test suite runs 👌

packages/merge-on-green/test/merge-on-green.ts
sofisl: I added in a comment explaining the situation in the code, but I'd also love to hear any other suggestions you might have!

PR #489, opened 2020-04-17
feat(infra): add Terraform support

orthros:
Adds a core "main" terraform module with two submodules

1. core

This is the root module which will be `terraform apply`-ed. This
contains some base provider definitions as well as  a declaration of an
AppEngine App (to use Cloud Tasks with), and the GCS Bucket to
upload and store our build artifacts in (which we will use to deploy our
Cloud Functions).

Currently, it is less than elegant to use, as the latest release of Terraform
(v0.12.24) does not support `for`, `for_each`, and `count` keywords on
modules. This means we need to manually add each bot to the list of
resources to be tracked in Terraform. This support is expected to come
in Terraform v0.13.0, but there is no active target date for that
release. There is a comment in main.tf of this module that explains how
we can update it to take advantage of that property.

2. serverless-scheduler-proxy

This module encapsulates everything that is needed to stand up and run
the serverless-scheduler-proxy, including a Service Account which has
run.invoker permissions on it (to be used by the Cloud Scheduler crons).

3. bot

This module encapsulates everything that is needed to run a single bot
(Cloud Function, Cloud Task, Scheduler Event).

This module needs to be expanded to support a Service Account with
Secret Manager access to the secret the Bot uses.


orthros:
TODO: add GitHub Action for Continuous Delivery.

orthros:
TODO:
1. Add Service Account for each bot
1. Merge #480 so Service Accounts can have permissions to access only THEIR secret

JustinBeckwith:
@orthros closing out old PRs - keeping the branch around though if you want to re-kindle this!

Issue #364, opened 2020-03-02
publish pre-release with a tag


alexander-fenster:
I would like to be able to publish a pre-release of the package (e.g. `next` tag) by commenting or adding a label to the release PR.

Suggested workflow (can be discussed):

0. Release PR is created by a bot.
1. I add `release-as:next` label to the release PR and merge it.
2. The bot makes a release but tags it as `next`. Optionally, the version might be set to `x.y.z-next`.
3. The bot creates a release PR that will be used to release the current version as `latest` when merged.

PR #490, opened 2020-04-17
feat(publish): label-based pre-release support

bcoe:
This feature allows you to add a `publish:candidate` label to a release PR, and it will:

1. create a `prerelease` release [like this](https://github.com/googleapis/nodejs-logging/releases/tag/v8.0.0-beta.1).
2. publish the `pre-release` to npm as the `next` tag `next: 8.0.0-beta.1`.
3. add a comment to the release PR, [like so](https://github.com/googleapis/nodejs-logging/pull/745#issuecomment-615454484).

fixes: #364


packages/publish/src/publish.ts
```diff
@@ -105,6 +110,154 @@ function handler(app: Application) {
...
+      } catch (err) {
+        // The most likely error is that the release already existed, and we
+        // had a publication failure; we can skip an error creating the release:
+        app.log.error(err);
+      }
+
+      // Publish the package:
+      const publishConfig = handler.publishConfigFromSecret(secret);
+      const npmRc = handler.generateNpmRc(publishConfig);
+      await handler.publish(npmRc, pkgPath, app, true);
```
alexander-fenster, line 210: It's probably borderline readable now, maybe it's a good time to change it from the linear list of parameters to accepting options?  Hard to guess that `true` means to make prerelease.

packages/publish/src/publish.ts
```diff
@@ -197,9 +355,14 @@ handler.publish = async (npmRc: string, pkgPath: string, app: Application) => {
       app.log.warn(err.message);
     }
     app.log.info(`publishing ${pkgPath}`);
+    // If this is a prerelease, we publish to the "next" dist tag on npm:
+    const publishOpts = ['publish', '--access=public'];
+    if (prerelease) {
+      publishOpts.push('--tag=next');
```
alexander-fenster, line 361: maybe move `next` to a variable? It occurs multiple times in the code.

alexander-fenster: if we decide to expand it later, we might as well want to allow tags like `publish:legacy-10` (then `publish:candidate` would logically become `publish:next`)

packages/publish/src/publish.ts
```diff
@@ -105,6 +110,154 @@ function handler(app: Application) {
...
+          tag_name: `v${candidateVersion.version}`,
+          name: `v${candidateVersion.version} Pre-release`,
+          prerelease: true,
+          target_commitish: 'master',
+          body: `This is a pre-release for testing purposes of #${context.payload.pull_request.number}.\n\nInstall by running \`npm install ${candidateVersion.name}@next\``,
+        });
+      } catch (err) {
+        // The most likely error is that the release already existed, and we
+        // had a publication failure; we can skip an error creating the release:
+        app.log.error(err);
```
JustinBeckwith, line 204: How do these failures get monitored today?  Can we verify it's the error case we think it is, and otherwise re-throw?  Worried about swallowing errors.

packages/publish/src/publish.ts
alexander-fenster: ...actually, extracting tags from the label should not be too hard to do right now?

packages/publish/src/publish.ts
```diff
@@ -105,6 +110,154 @@ function handler(app: Application) {
...
+    `https://registry.npmjs.org/${encodeURIComponent(pkgJson.name)}`
+  )
+    .then(res => {
+      return res.json();
+    })
+    .then(body => {
+      return Object.keys(body.time);
+    });
+  let counter = 0;
+  let candidateVersion = `${futureVersion}-beta.${counter}`;
```
JustinBeckwith, line 249: My brain melted trying to understand what's happening here.  Maybe a comment explaining?

packages/publish/src/publish.ts
```diff
@@ -105,6 +110,154 @@ function handler(app: Application) {
...
+  for (;;) {
+    if (!versions.includes(candidateVersion)) {
+      break;
+    } else {
+      counter++;
+      candidateVersion = `${futureVersion}-beta.${counter}`;
+    }
+  }
+  pkgJson.version = candidateVersion;
+  writeFileSync(`${pkgPath}/package.json`, JSON.stringify(pkgJson), 'utf8');
```
JustinBeckwith, line 259: Why `sync`? Any reason to not use `promises.writeFile`?

packages/publish/src/publish.ts
bcoe: @JustinBeckwith 👍 I'll figure out what error code t is and catch specifically.

packages/publish/src/publish.ts
bcoe: habit, I'll try the promise version.

packages/publish/src/publish.ts
bcoe: talked offline, let's make this follow up work (supporting multiple labels).

PR #494, opened 2020-04-22
fix(Build Cop): keep existing priority and type labels

### LINKED ISSUE #484: Build Cop: don't change priority or type of issues
If an issue already has `priority: pN` or `type: foo` labels, the bot should not add the `priority: p1` or `type: bug` labels. The bot should be helpful, not persnickety.

If there is no `priority` label, the bot should re-add `priority: p1` (for example, https://github.com/googleapis/google-cloud-go/issues/1840). If there is no `type`, it should add `type: bug`.

I don't think this requires an explanatory comment from the bot ("Looks like the `priority` label was removed, re-adding the default `priority: p1`.") But, I could be convinced.

Regardless of type, issues should always have a `priority` label.

cc @JustinBeckwith @meredithslota

tbpg:
Fixes #484.

### DIFF
--- a/packages/buildcop/src/buildcop.ts
+++ b/packages/buildcop/src/buildcop.ts
@@ -200,7 +200,7 @@ buildcop.deduplicateIssues = async (
 
   let modified = false;
 
-  for (const [title, issues] of byTitle) {
+  for (const issues of byTitle.values()) {
     if (issues.length <= 1) {
       continue;
     }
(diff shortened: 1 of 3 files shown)


packages/buildcop/src/buildcop.ts
```diff
@@ -510,7 +510,16 @@ buildcop.markIssueFlaky = async (
   const existingLabels = existingIssue.labels
     ?.map(l => l.name)
     .filter(l => !l.startsWith('buildcop'));
-  const labels = LABELS_FOR_FLAKY_ISSUE.concat(existingLabels);
+  let labels = LABELS_FOR_FLAKY_ISSUE;
+  // If existingLabels contains a priority: or type: label, don't add the
+  // default priority: and type: labels.
+  if (existingLabels?.find(l => l.startsWith('priority:'))) {
+    labels = labels.filter(l => !l.startsWith('priority:'));
```
JustinBeckwith, line 517: I am a little confused by this line.  Don't you want to `concat`?

packages/buildcop/src/buildcop.ts
```diff
@@ -510,7 +510,16 @@ buildcop.markIssueFlaky = async (
   const existingLabels = existingIssue.labels
     ?.map(l => l.name)
     .filter(l => !l.startsWith('buildcop'));
-  const labels = LABELS_FOR_FLAKY_ISSUE.concat(existingLabels);
+  let labels = LABELS_FOR_FLAKY_ISSUE;
```
JustinBeckwith, line 513: Up above is a linter warning.  I wanted to make warnings errors but forgot.  Mind dropping `title` to keep linter errors down?

tbpg: Done.

packages/buildcop/src/buildcop.ts
tbpg: Changed the variable name to hopefully make this more clear. Please take another look.

PR #498, opened 2020-04-24
feat: add a config option in blunderbuss to install by org

sofisl:
This feature allows for users to install blunderbuss to an organization, and specify which repos it wants it to run for (and whether or not there are users assigned to review prs and issues there as well). For existing configurations, the only change that would need to be made would be to add ```repo: ``` to the top of the yml file to accurately represent the new object. However, if a user wants the default configuration (i.e., installing the bot on a single repository), the user does not need to specify the ```repoName```, and the bot will assume it is supposed to run for the repo it is installed on.


kurtisvg:
Could you expand on the motivation for this change for me? I'm not sure how I'm following the intended use case.  

Blunderbuss configs are per repo, and the default behavior if no config is set is to do nothing. If it is enabled org wide nothing would change - it would simply ignore events from repos that don't have a `.bunderbuss.yml` set. 

I'm not sure I understand how specifying multiple repos per config is supposed to work. Would I add a config to Repo A that specifies how Repo B works? That seems like an illogical approach - why not just add a config to Repo B?

sofisl:
@kurtisvg, you are totally right 🤦‍♀️. Going to close this for now. Thanks for the quick review!

PR #506, opened 2020-04-27
feat(gcf-utils)!: Use Secret Manager as opposed to KMS + GCS 

orthros:
Now uses SecretManager to store Probot configs. This obviates the DIY
solution we had for storing these secrets.

This is an incremental step. Bots can gradually opt-in to using the Secret Manager version by not updating to `v4.0.0` of `gcf-utils`. 

This fixes the previous PR (of the same name) by still setting the (now superfluous) environment variables that `v3` of `gcf-utils` uses.

This does, however, migrate `serverless-scheduler-proxy` to using the Secret Manager version, but this should be transparent to bots as the secret key data has been duplicated into Secret Manager.


packages/gcf-utils/src/bin/genkey-util.ts
```diff
@@ -0,0 +1,62 @@
...
+import {v1} from '@google-cloud/secret-manager';
+
+export async function gather(
+  keyfile: string,
+  id: number,
+  webhookSecret: string
+): Promise<Options> {
+  let keyContent = '';
+  try {
+    keyContent = fs.readFileSync(keyfile, 'utf8');
```
JustinBeckwith, line 26: It's preferred to do:

```js
import {promises as fs} from 'fs';
// ....
keyContent = await fs.readFile(keyfile, 'utf8');
```

packages/gcf-utils/src/bin/genkey-util.ts
```diff
@@ -0,0 +1,62 @@
...
+  } catch (e) {
+    throw Error(`Error reading file: ${keyfile}`);
```
JustinBeckwith, line 28: This would just swallow the original exception.  I think you want to preserve the call stack and error message.  Generally in this case, I'd remove the try/catch and let it throw.  If you want to preserve the `Error reading file` bit, you can prepend to the origin `e.message`

packages/gcf-utils/src/bin/genkey-util.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  return {
+    cert: keyContent,
+    id,
+    secret: webhookSecret,
+  } as Options;
```
JustinBeckwith, line 35: I would assume the `as Options` wouldn't be needed here

packages/gcf-utils/src/bin/genkey-util.ts
```diff
@@ -0,0 +1,62 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import fs from 'fs';
+import {Options} from 'probot';
+import {v1} from '@google-cloud/secret-manager';
+
+export async function gather(
```
JustinBeckwith, line 19: Could we get a little comment at the top that tells me what each of these methods are doing?

packages/gcf-utils/test/gcf-utils.ts
```diff
@@ -19,6 +19,8 @@ import {Options} from 'probot';
 import * as express from 'express';
 import sinon from 'sinon';
 import nock from 'nock';
+import {expect} from 'chai';
```
JustinBeckwith, line 22: For the sake of consistency, could we avoid `chai` and use `assert`?

packages/gcf-utils/test/gcf-utils.ts
```diff
@@ -212,4 +214,156 @@ describe('GCFBootstrapper', () => {
...
+    afterEach(() => {
+      secretsStub.reset();
+      secretVersionNameStub.reset();
+    });
+
+    it('gets the config', async () => {
+      secretsStub = sinon
+        .stub(secretClientStub, 'accessSecretVersion')
+        .callsFake(() => {
+          return Promise.resolve([
```
JustinBeckwith, line 244: Instead of using `.callsFake` like this, you can just say `.resolves([{ payload ...`

packages/gcf-utils/test/gcf-utils.ts
```diff
@@ -212,4 +214,156 @@ describe('GCFBootstrapper', () => {
...
+            {
+              payload: {},
+            },
+          ]);
+        });
+
+      const configSpy = sinon.spy(bootstrapper, 'getProbotConfig');
```
JustinBeckwith, line 295: I think this would be simpler if you did something like:
```js
assert.rejects(bootstrapper.getProbotConfig());
```

If I'm understanding the intent here right

packages/gcf-utils/test/gcf-utils.ts
```diff
@@ -212,4 +214,156 @@ describe('GCFBootstrapper', () => {
...
+      } catch (e) {
+        // pass
+      }
+      expect(configSpy.threw());
+    });
+
+    it('throws on empty response', async () => {
+      secretsStub = sinon
+        .stub(secretClientStub, 'accessSecretVersion')
+        .callsFake(() => {
```
JustinBeckwith, line 307: Like above, this can be shortened to `.resolves([{}])`

packages/gcf-utils/src/bin/genkey-util.ts
orthros: Done.

packages/gcf-utils/src/bin/genkey-util.ts
orthros: Done.

packages/gcf-utils/src/bin/genkey-util.ts
orthros: I remember in early iterations of this, not casting the value `as Options` caused a compile error, but apparently that is not the case.

packages/gcf-utils/src/bin/genkey-util.ts
orthros: Done.

packages/gcf-utils/test/gcf-utils.ts
orthros: Done. 

There are a number of bots that need to conform to this as well. I'll make a separate issue to enforce consistency.

* auto-label
* buildcop
* generate-bot
* publish
* snippet-bot

packages/gcf-utils/test/gcf-utils.ts
orthros: Done.

packages/gcf-utils/test/gcf-utils.ts
orthros: It is! 👍

packages/gcf-utils/test/gcf-utils.ts
orthros: Done.

packages/gcf-utils/package.json
```diff
@@ -30,6 +31,7 @@
   },
   "devDependencies": {
     "@types/bunyan": "^1.8.6",
+    "@types/chai": "^4.2.9",
```
JustinBeckwith, line 34: Looks like we can drop this

packages/gcf-utils/package.json
```diff
@@ -40,6 +42,7 @@
     "@types/tmp": "^0.1.0",
     "@types/yargs": "^15.0.0",
     "c8": "^7.1.0",
+    "chai": "^4.2.0",
```
JustinBeckwith, line 45: and this

packages/gcf-utils/package.json
```diff
@@ -49,7 +52,7 @@
     "typescript": "^3.8.3"
   },
   "engines": {
-    "node": ">= 10.13.0"
+    "node": "^10.17.0"
```
JustinBeckwith, line 55: Oh that's interesting!  @bcoe I've never seen proper semver used here before.  Is that vaid?

packages/gcf-utils/package.json
```diff
@@ -62,4 +65,4 @@
   "files": [
     "build/src"
   ]
-}
+}
```
JustinBeckwith, line 68: newline to make the red squiggle go away please :)

packages/gcf-utils/package.json
orthros: Context: It was the only way I could get `gts` to work and not complain about `promise` in the `fs` package not being available/backported.

packages/gcf-utils/package.json
orthros: Is there a formatter that you use that doesn't do this? I've been using VSCode's `Format on save` feature and it always removes the newline at the end of file (which requires a bunch of dancing to disable and re-enable to fix this).

JustinBeckwith: Hmmm, not that I know of for JSON files.  We have it set up properly for prettier integration for ts and js files though.  Maybe I need to see if prettier does JSON....

packages/gcf-utils/package.json
bcoe: I believe it can be any semver:

https://docs.npmjs.com/files/package.json#engines

In their npm example they use `~`.

PR #522, opened 2020-05-04
feat(auto-label): detect labels from issue titles

### LINKED ISSUE #400: Auto Label: label issues based on issue title
Issues in some repos start with the product/package with the issue. For example, Pub/Sub issues start with `pubsub:`.

Build Cop Bot issues start with the package name of the failing test. Many times, that package name includes some form of the product name.

I'd like to automatically apply `api` labels for these issues.

If the issue title contains a fuzzy match of a product name, apply that label.

We could look at the issue title and the issue body. But, it's possible the body would contain something other than what we want (e.g. a Cloud Functions test that uses Pub/Sub is not an `api: pubsub` issue. We may even want to restrict to before the first `:` and ignore the issue if there isn't a `:`.

tbpg:
This only does exact matching; no fuzzy matching (yet).

For example:
* This issue will be labeled `api: spanner`: `spanner/transactions: TestAbort failed`
* However, this issue will not be labeled: `How do I abort a Spanner transaction?`

Fixes #400.


packages/auto-label/src/auto-label.ts
```diff
@@ -148,6 +150,38 @@ handler.checkIfElementIsInArray = function checkIfElementIsInArray(
...
+  github: GitHubAPI,
+  jsonArray: JSONData[],
+  issue: Webhooks.WebhookPayloadIssuesIssue,
+  owner: string,
+  repo: string
+): Promise<string | undefined> => {
+  if (!jsonArray) {
+    return undefined;
+  }
+  let firstPart = issue.title.split(':')[0]; // Before the colon, if there is one.
```
JustinBeckwith, line 168: Not sure we need to be *this* prescriptive about parsing titles.  Would just checking for an includes be terrible?

tbpg: Let's start here. I'd like to take a look at the logs after this and see what title formats don't parse, then we can update this logic.

tbpg:
Thanks!

PR #571, opened 2020-06-03
feat(README): add architecture diagram

orthros:
Adds an architecture diagram to README.md


orthros:
I'm not familiar enough with SVGs to determine why the icons are broken.

PR #574, opened 2020-06-04
feat(auto-label): support more APIs and issue titles

tbpg:
This adds support for some Python issues that have the API as the part before the period (for example, https://github.com/GoogleCloudPlatform/python-docs-samples/issues/3923).

This also adds support for APIs that are prefixed with `cloud` like `cloudiot` and `cloudkms`.

I updated the auto-detect function signature a bit to make it easier to write tests. Let me know if there is a more idiomatic way to write a table-driven style test.


tbpg:
Thanks!

These are for issue titles, which don't seem to follow conventional commits as much. Would probably need regex extraction, so I will leave that for a followup feature.

PR #595, opened 2020-06-16
feat(Build Cop): group tests into a single issue when 10+ fail

### LINKED ISSUE #550: Build Cop: when more than N tests fail in a package, group them into a single issue
If something is wrong with an entire package, all of the tests in that package can fail. This can lead to filing dozens of issues at a time. We should have a stopgap solution to avoid filing dozens of issues at once.

Proposal:
* If more than 10 issues fail in a single build, file an issue with the title, "packagename: many tests failed"
* Include the name of every failed test in the body of the issue.
* The issue is closed when all of the tests pass for that package.
* No new issues should be opened for that package.
* If there are already individual issues for some of the failed tests, we should leave them open and reference them in the overall issue.
  * When that test starts passing, we can close the individual issue.
  * If the test fails after the issue is closed, we should leave it closed and not mark it flaky. We can comment on the issue

tbpg:
If the common setup code for 100 tests fails, in some cases all 100 tests with "fail." Before this change, the bot would file a separate issue for each of those tests, which was very annoying.

This change groups tests into a single issue when 10 or more fail in a single build.

I made the choice to leave existing individual issues open when opening a "grouped" issue. People may be actively discussing a broken test in one of the issues; it would be rude if the bot decided to close the existing issue and end the discussion.

Note: I refactored some of the existing logic into a `findExistingIssue` function. The diff does not make that very clear.

Areas to watch out for in review:
* The choice to leave existing issues open.
* README docs, the grouped issue body, and grouped issue comments. Are they helpful and clear?
* Test coverage -- missing any edge cases that should be tested?
* I'm always open to JS/TS/Node.js style advice.
* And, you know, bugs. :smile: 

Fixes #550

### DIFF
--- a/packages/buildcop/README.md
+++ b/packages/buildcop/README.md
@@ -14,6 +14,12 @@ The Build Cop Bot manages issues for failing tests.
   reopen the issue or open a new one, depending on the issue state/age.
 * If the bot opens duplicate issues (sorry!), it will close the duplicates
   during the next run.
+* If 10 or more tests fail in the same package, they will all be grouped into a
+  single issue. If there are already open issues for tests in the package, they
+  will be left open until the corresponding test passes.
+
+  Note: the "package" is the part of the issue title before the `:`. If the
+  package is wrong, please file an issue.
 
 Issues or feature requests? Please
 [file them on this repo](https://github.com/googleapis/repo-automation-bots/issues/new).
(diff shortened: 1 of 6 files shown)


bcoe:
I'm going to deploy this in 10 minutes or so, once a new version of release-please is out the door (I want to kick off a redeploy).

PR #601, opened 2020-06-17
feat(sync-repo-settings): allow configuring extra team permissions by language

chingor13:
Adds a new teams.json source file to contain the additional teams.

Alternatively, we can add the teams configuration in the `required-checks.json` which we should probably rename to `settings.json` as it's grown beyond just required status checks.


packages/sync-repo-settings/package.json
```diff
@@ -18,7 +18,7 @@
     "probot-app"
   ],
   "scripts": {
-    "compile": "tsc -p . && cp ./src/required-checks.json ./build/src/required-checks.json.json",
+    "compile": "tsc -p . && cp ./src/required-checks.json ./build/src/required-checks.json.json && cp ./src/teams.json ./build/src/teams.json.json",
```
JustinBeckwith, line 21: Why do we need to copy these json files at all into the build?  We could include them in the module via the `files` field in `package.json`, and rewrite the paths so we don't need these gymnastics.

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -35,6 +37,17 @@ interface LanguageConfig {
   };
 }
 
+type Permission = 'pull' | 'push' | 'admin' | 'maintain' | 'triage';
```
JustinBeckwith, line 40: I'm fine having this here, but I also know this is expressed somewhere in OctoKit already :)

packages/sync-repo-settings/src/teams.json
```diff
@@ -0,0 +1,8 @@
+{
+    "java": [
+        {
+            "slug": "java-samples-reviewers",
```
JustinBeckwith, line 4: Per our discussion in the team meeting today, I love the idea of this being generalizable across languages.

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -154,30 +167,43 @@ handler.updateMasterBranchProtection = async function updateMasterBranchProtecti
...
       permission: 'push',
     },
   ];
+};
+
+handler.getRepoTeams = function (repo: Repo): TeamPermission[] {
+  const language = repo.language;
+  let teams = handler.defaultLanguageTeams(language);
+  if (language in languageTeams) {
+    teams = teams.concat(languageTeams[language]);
```
JustinBeckwith, line 191: You could keep `teams` a `const`, and do `teams.push(...languageTeams[language])` as well.  (non-blocking feedback)

packages/sync-repo-settings/src/teams.json
```diff
@@ -0,0 +1,8 @@
...
+            "permission": "push"
+        }
+    ]
+}
```
JustinBeckwith, line 8: Can I get a newline?

packages/sync-repo-settings/src/sync-repo-settings.ts
chingor13: Looks like octokit does not publish a type or enum for these constants

packages/sync-repo-settings/package.json
chingor13: I'm not sure why this is here, but I followed the pattern.

packages/sync-repo-settings/src/sync-repo-settings.ts
chingor13: Done

packages/sync-repo-settings/src/teams.json
chingor13: Done

packages/sync-repo-settings/package.json
chingor13: Filed https://github.com/googleapis/repo-automation-bots/issues/603 to track. I'm not sure how to fix this.

packages/sync-repo-settings/src/teams.json
chingor13: For now, I think only java-samples-reviewers exists and we would like this group on all the java repos. We can refactor this one the internal design is approved.

PR #602, opened 2020-06-17
feat(sync-repo-settings): allow configuring extra team permissions by language

chingor13:
Alternative to #601 

Instead, puts the configuration into required-checks.json which is a misnomer now.


chingor13:
Closing in favor of #601 - required-checks.json is already doing too much

PR #605, opened 2020-06-18
fix: ensure that function does not depend on errors to execute correctly

### LINKED ISSUE #487: Merge-on-green: logs errors when checking pages of statuses that do not exist
MOG will leave the following error message when it checks a certain page of status checks or check runs that do not exist:

```Error in getting statuses: Cannot read property 'context' of undefined```

I believe this functionality is caused by the iterateGetStatusi and iterateGetCheckRuns functions, because they basically call 10 pages of these checks/statuses in a loop. Currently, this actually helps the bot function because it will exit out of the loop once it gets to a page that it cannot read and continue on with the rest of the function. However, should we consider a different implementation that doesn't rely on failure (or, should we just not log the failures?)

@bcoe and @tbpg, would love your thoughts on this!

### PR DESCRIPTION
**Pull request by sofisl**
Fixes #487

### CODE REVIEW

File: packages/merge-on-green/src/merge-logic.ts [hunk 1]
```diff
@@ -211,8 +211,8 @@ mergeOnGreen.getStatusi = async function getStatusi(
       per_page: 100,
       page: num,
     });
-    if(data[0].context === undefined) {
-      console.info(`no further page data`)
+    if (data[0].context === undefined) {
```
tbpg, line 214: Maybe shorten? Same with a few other `=== undefined` and `=== 0` instances below.

```suggestion
    if (!data[0].context) {
```

sofisl: Thank you! I will leave the ```.length === 0``` ones because there are many methods that return an empty array instead of returning falsy.

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -45,6 +45,10 @@ interface CheckRuns {
   check_runs: [{name: string; conclusion: string}];
 }
 
+interface Comments {
+  body: string;
+}
```
tbpg, line 50: Maybe from a merge?

packages/merge-on-green/src/merge-logic.ts
```diff
@@ -240,9 +244,10 @@ mergeOnGreen.iterateGetStatusi = async function iterateGetStatusi(
   let results: CheckStatus[] = [];
   for (let i = 0; i < 10; i++) {
     const temp = await mergeOnGreen.getStatusi(owner, repo, github, headSha, i);
-    if (temp.length !== 0) {
-      results = results.concat(temp);
+    if (temp.length === 0) {
+      return results;
```
tbpg, line 248: Not sure I understand this line -- return early if there is no more `getStatusi`?

sofisl: Yep, exactly. Basically, check to make sure that that page returns 0 results, and then return whichever results you've already accumulated in "results"

PR #607, opened 2020-06-18
fix: remove label from PR if automerge fails

### LINKED ISSUE #578: merge-on-green: Remove label once MoG gives up
Current MoG leaves a comment when the merge has timed out, asking the user to remove the label and reapply. Any reason not to remove the labels as we leave the comment? This way if the user wants to retry, they only have to do one step.

sofisl:
Fixes #578


kurtisvg:
Thanks @sofisl!

lesv:
Yea!

PR #609, opened 2020-06-18
feat: Slo-status-label bot currently with linting feature

### LINKED ISSUE #596: New Bot: SLO Labeler
Need a bot that will automatically label issues that are OOSLO based on an open, defined schema (preferably JSON). 

Requirements:
* Should lint and validate changes to the SLO definitions
* Should allow for SLOs to be defined in-repository or in an Organization (in the `.github`) repository.
* Should periodically (every 5 minutes?) scan and label OOSLO issues (and remove labels on Issues that are in SLO)

hershey127:

Fixes #596 🦕


packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,183 @@
...
+        issue_number,
+        body
+      });
+    } catch (err) {
+        console.log(err);
+        return;
+    }
+  }
+  return;
+}
```
orthros, line 153: For this function, I'd consider either moving the declaration of the `body` variable into the `if(!isValid)` block so as to not "waste" an allocation on the stack.

That said, if we include a second check and comment if it IS valid (i.e. the discussed "Here are the changes to your SLO rules" comment), then leaving that declaration is perfectly fine.

packages/slo-stat-bot/test/slo-stat-label.ts
```diff
@@ -0,0 +1,305 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import {resolve} from 'path';
+import {Probot} from 'probot';
+import nock from 'nock';
+import * as fs from 'fs';
+import {expect} from 'chai';
```
orthros, line 20: In general we don't use `expect` from chai at all and use the plain `assert` package.

packages/slo-stat-bot/test/slo-stat-label.ts
```diff
@@ -0,0 +1,305 @@
...
+      getInstallationAccessToken(): Promise<string> {
+        return Promise.resolve('abc123');
+      },
+    };
+    probot.load(handler);
+  });
+
+  describe('opened or reopened pull request', () => {
+    let payload: Webhooks.WebhookPayloadPullRequest;
+    let stub: any;
```
orthros, line 62: As opposed to providing the `any` type assign it to `sinon.SinonStub`

packages/slo-stat-bot/test/slo-stat-label.ts
```diff
@@ -0,0 +1,305 @@
...
+        fixturesPath,
+        'events',
+        'contains_slo_file'
+      ));
+      const requests = nock('https://api.github.com')
+        .get('/repos/testOwner/testRepo/pulls/6/files?per_page=100')
+        .reply(200,containsSloFile)
+      await probot.receive({name: 'pull_request.opened', payload, id: 'abc123',});
+
+      expect(stub.called).to.equals(true);
```
orthros, line 98: As opposed to using `expect` here, use ` sinon.assert.calledOnce(stub)`

packages/slo-stat-bot/test/slo-stat-label.ts
```diff
@@ -0,0 +1,305 @@
...
+        'events',
+        'not_contains_slo_file'
+      ));
+      const requests = nock('https://api.github.com')
+        .get('/repos/testOwner/testRepo/pulls/6/files?per_page=100')
+        .reply(200,doesNotContainSLOFile)
+
+      await probot.receive({name: 'pull_request.opened', payload, id: 'abc123',});
+
+      expect(stub.called).to.equals(false);
```
orthros, line 114: As opposed to using expect here, use sinon.assert.calledOnce(stub)

packages/slo-stat-bot/test/slo-stat-label.ts
```diff
@@ -0,0 +1,305 @@
...
+          fixturesPath,
+          'events',
+          'issue_slo_rules',
+          'valid_slos',
+          fileName
+        ));
+        let validRes = await handler.lint(schema, slo);
+        let isValid = await validRes.isValid;
+       
+        expect(isValid).to.equals(true);
```
orthros, line 283: Should use `assert.strictEqual(isValid, true)` here

packages/slo-stat-bot/test/slo-stat-label.ts
```diff
@@ -0,0 +1,305 @@
...
+          fixturesPath,
+          'events',
+          'issue_slo_rules',
+          'invalid_slos',
+          fileName
+        ));
+        let validRes = await handler.lint(schema, slo);
+        let isValid = await validRes.isValid;
+       
+        expect(isValid).to.equals(false);
```
orthros, line 301: Should use `assert.strictEqual(isValid, false)

packages/slo-stat-bot/package.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "slo-stat-label",
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
hershey127: Right my bad. Will fix!

packages/slo-stat-bot/test/fixtures/events/contains_slo_file.json
```diff
@@ -0,0 +1,14 @@
+[
+    {
+      "sha": "1d8be9e05d65e03a5e81a1c3c1bf229dce950e25",
+      "filename": "issue_slo_rules.json",
+      "status": "modified",
+      "additions": 5,
+      "deletions": 5,
+      "changes": 10,
+      "blob_url": "https://github.com/hershey127/Testing/blob/f345347ba98f8c24969f1dcf340a9dfb3ecfe36f/issue_slo_rules.json",
+      "raw_url": "https://github.com/hershey127/Testing/raw/f345347ba98f8c24969f1dcf340a9dfb3ecfe36f/issue_slo_rules.json",
```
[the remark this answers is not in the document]
hershey127: That's a good point. Will keep that in mind next time!

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,183 @@
...
+  PullsListFilesResponseItem,
+  Response,
+  ChecksCreateParams
+} from '@octokit/rest';
+import Ajv, { ErrorObject } from 'ajv';
+
+const schema = require("./../utils/schema.json");
+
+type ValidationResults = {
+  isValid: Promise<boolean>,
```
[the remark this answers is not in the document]
hershey127: Thats true. Will fix it!

packages/slo-stat-bot/test/slo-stat-label.ts
hershey127: Oh I didn't realize that because the "auto-label" bot uses expect for the unit tests. But I can change it to assert.

packages/slo-stat-bot/test/slo-stat-label.ts
hershey127: This test is looking to see if "issue_slo_rules.json" is not present, then the handles.slo rules is not called. So I can use sinon.assert.notCalled(stub) instead.

packages/slo-stat-bot/test/slo-stat-label.ts
hershey127: I see why assert is better to use in this scenario than expect because we want the current function to abort fast.

packages/slo-stat-bot/src/slo-stat-label.ts
hershey127: I can see from a reviewers pov how including the specific changes of SLO rules can make it easier for them. I will include this feature in the next iteraton.

packages/slo-stat-bot/test/slo-stat-label.ts
hershey127: I created a sandbox instead so I don't have to keep track of any fakes or stubs in case I make more for future test cases. I was unable to set the return type to be specifically sinon.SinonStub because it was erroring out.

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,183 @@
...
+  errors?: ErrorObject [] | null | undefined
+};
+
+function handler(app: Application) {
+  app.on(['pull_request.opened', 'pull_request.reopened'], async (context: Context) => {
```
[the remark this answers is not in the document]
hershey127: I was thinking the other two PR actions would be: pull_request.synchronized & pull_request.edited

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,219 @@
...
+      if (fileResponse !== undefined) fileList = fileResponse.data;
+
+      for (
+        let i = 0;
+        fileList !== undefined && fileList[i] !== undefined;
+        i++
+      ) {
+        const file = fileList[i];
+
+        if (file.filename === 'issue_slo_rules.json') {
```
orthros, line 66: How does this handle the case where the following files are modified

* ./issue_slo_rules.json
* ./foo/issue_slo_rules.json
* ./.github/issue_slo_rules.json

Looking at it, this would attempt to lint/verify all of them when we would only want to lint `./issue_slo_rules.json` when the repository is named `.github` and we would never want to lint `./foo/issue_slo_rules.json`

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,219 @@
...
+handler.commentPR = async function commentPR(
+  github: GitHubAPI,
+  owner: string,
+  repo: string,
+  issue_number: number,
+  isValid: boolean
+) {
+  if (!isValid) {
+    const body =
+      'ERROR: "issue_slo_rules.json" file is not valid with Json schema';
```
orthros, line 176: Replace Json with JSON

packages/slo-stat-bot/test/slo-stat-label.ts
```diff
@@ -0,0 +1,345 @@
...
+
+// eslint-disable-next-line node/no-extraneous-import
+import Webhooks from '@octokit/webhooks';
+import snapshot from 'snap-shot-it';
+import handler from '../src/slo-stat-label';
+import spies from 'chai-spies';
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const chai = require('chai');
+chai.use(spies);
```
orthros, line 31: These should not be needed anymore, correct?

packages/slo-stat-bot/test/slo-stat-label.ts
```diff
@@ -0,0 +1,345 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const sinon = require('sinon');
```
orthros, line 34: Is there a reason you did this as opposed to `import sinon from 'sinon''`?

hershey127: I am unable to import it because I cannot create a sandbox for easily restoring any stubs or fakes. When using import sinon from 'sinon' it is expecting us to create a SinonSpy<any,any>. I can change my code to define stubs individually w/o using sandbox if needed.

packages/slo-stat-bot/src/slo-stat-label.ts
hershey127: I will add another test to make sure it does not validate wrong file

packages/slo-stat-bot/test/slo-stat-label.ts
orthros: Looking through all the `sandbox` docs, it looks like `sandbox `is probably too heavyweight for what we need to use it for. I'd replace the `sandbox` with a single stub. I'll comment more down below.

packages/slo-stat-bot/test/slo-stat-label.ts
```diff
@@ -0,0 +1,345 @@
...
+      getInstallationAccessToken(): Promise<string> {
+        return Promise.resolve('abc123');
+      },
+    };
+    probot.load(handler);
+  });
+
+  describe('opened or reopened pull request', () => {
+    let payload: Webhooks.WebhookPayloadPullRequest;
+    const sandbox = sinon.createSandbox();
```
orthros, line 64: Replace the sandbox with 

```typescript
let sloStub : sinon.SinonStub;
```

Then in the `beforeEach` replace `sandbox.stub` with

```typescript
sloStub = sinon.stub(handler, 'handle_slos');
```

and remove the `sandbox.restore();` from the `afterEach`.

This _should_ be all you'd need to replace the sandbox

orthros:
@martinvarelaj Summoning to take a look

orthros:
Asking @bcoe for a second set of eyes

packages/slo-stat-bot/utils/schema.json
```diff
@@ -0,0 +1,107 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "SLO rules",
```
bcoe, line 3: I think right now we don't copy over the whole package directory during the cloud build step, we should either update [this script](https://github.com/googleapis/repo-automation-bots/blob/master/scripts/build-functions.sh#L37) so that we copy over everything, including `utils`.

packages/slo-stat-bot/test/fixtures/events/issue_slo_rules/invalid_slos/incorr_time_format.json
```diff
@@ -0,0 +1,13 @@
...
+            "issues": true
+        },
+        "complianceSettings": {
+            "responseTime": 0,
+            "resolutionTime": "1h 40m"
+        }
+    }
+ ]
+ 
+
```
bcoe, line 13: nit: we should make our linter catch multiple training newlines in files like this.

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,221 @@
...
+};
+
+handler.commentPR = async function commentPR(
+  github: GitHubAPI,
+  owner: string,
+  repo: string,
+  issue_number: number,
+  isValid: boolean
+) {
+  if (!isValid) {
```
JustinBeckwith, line 176: This is a place where I'd flip the condition of the `if` to keep the code more flat.  You could say something like:

```js
if (isValid) {
  return;
}
```

Then you could go about the rest of the function without being inside of an `if` clause.

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,221 @@
...
+      'ERROR: "issue_slo_rules.json" file is not valid with JSON schema';
+    try {
+      await github.issues.createComment({
+        owner,
+        repo,
+        issue_number,
+        body,
+      });
+    } catch (err) {
+      console.log(err);
```
JustinBeckwith, line 187: @bcoe @sofisl seeing these make me a littler nervous.  What's our general guidance on exceptions?  Is it bad to "let it throw" in probot apps?

packages/slo-stat-bot/test/fixtures/config/valid-config.yml
```diff
@@ -0,0 +1 @@
+randomBoolean: true
```
JustinBeckwith, line 1: Please add a newline!

packages/slo-stat-bot/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "test": "cross-env LOG_LEVEL=fatal c8 mocha --exit build/test",
+    "test:snap": "SNAPSHOT_UPDATE=1 npm test",
+    "fix": "gts fix",
+    "lint": "gts check"
+  },
+  "dependencies": {
+    "@octokit/rest": "^16.41.2",
+    "@types/chai-spies": "^1.0.1",
+    "assert": "^2.0.0",
+    "chai-spies": "^1.0.0",
```
JustinBeckwith, line 36: I took a closer look, and it doesn't look like you're using `chai` at all!  Maybe remove these dependencies from `package.json`?

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,221 @@
...
+      'pull_request.reopened',
+      'pull_request.edited',
+      'pull_request.synchronize',
+    ],
+    async (context: Context) => {
+      const owner = context.payload.repository.owner.login;
+      const repo = context.payload.repository.name;
+      const pull_number = context.payload.number;
+
+      const fileResponse:
```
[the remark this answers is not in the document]
hershey127: That makes sense. I'll try to follow that style from now on.

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,221 @@
...
+        owner,
+        repo,
+        pull_number,
+        100
+      );
+      let fileList: PullsListFilesResponseItem[] | undefined;
+
+      if (fileResponse !== undefined) fileList = fileResponse.data;
+
+      for (
```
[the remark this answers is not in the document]
hershey127: That is correct! I definitely see how this makes the code much cleaner

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,221 @@
...
+            file.sha
+          );
+          break;
+        }
+      }
+    }
+  );
+}
+
+handler.handle_slos = async function handle_slos(
```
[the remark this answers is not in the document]
hershey127: Sorry about that. I will add method headers to clarify some functions.

packages/slo-stat-bot/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "prepare": "npm run compile",
+    "pretest": "npm run compile",
+    "test": "cross-env LOG_LEVEL=fatal c8 mocha --exit build/test",
+    "test:snap": "SNAPSHOT_UPDATE=1 npm test",
+    "fix": "gts fix",
+    "lint": "gts check"
+  },
+  "dependencies": {
+    "@octokit/rest": "^16.41.2",
+    "@types/chai-spies": "^1.0.1",
```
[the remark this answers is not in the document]
hershey127: My bad. Will change

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,221 @@
...
+  file_sha: string
+) {
+  const sloString = await handler.getFileContents(
+    context.github,
+    owner,
+    repo,
+    file_sha
+  );
+
+  if (sloString !== undefined) {
```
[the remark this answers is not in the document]
hershey127: I was trying to look for if it is not a string. So I'll handle those checks like this from now on!

packages/slo-stat-bot/src/slo-stat-label.ts
sofisl: I catch errors in some of the other bots in the same way. However, I think you want to make sure you're not catching an error that should stop execution, i.e., that you're preserving the functionality of what the bot is supposed to do. So, for example, with getFileContents(), I would make sure you catch the error and make sure you're returning what you want to check for. Later on, you check that function for undefined, so maybe make sure that in the function, if you get an error, you also return ```undefined```. Also, since this bot is supposed to get stats, make sure that you're doing something with that error message, in other words, make sure an error with that function doesn't just get lost in the logs. One idea could be that you return a file content string that has a special value that you can then run stats on to check any errors, i.e., ```return fileContent = 'error in getting contents'```, or for your listFiles response, you can return a ```PullsListFilesResponse``` item that returns a special value you can check for any errors. Or, you can also let the error stop the execution if it's essential to the bot. Just some ideas, I think it ultimately depends on how you want the bot to function!

packages/slo-stat-bot/package.json
```diff
@@ -0,0 +1,65 @@
...
+  },
+  "dependencies": {
+    "ajv": "^6.12.2",
+    "gcf-utils": "1.6.2",
+    "probot": "^9.11.3"
+  },
+  "devDependencies": {
+    "@types/body-parser": "^1.17.0",
+    "@types/bunyan": "^1.8.6",
+    "@types/chai": "^4.1.7",
```
JustinBeckwith, line 41: I think you can remove `@types/chai` and `chai` from here

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,239 @@
...
+
+      const fileList = await handler.listFiles(
+        context.github,
+        owner,
+        repo,
+        pull_number,
+        100
+      );
+
+      if (fileList === null) {
```
JustinBeckwith, line 62: To make sure we capture undefined, I would instead try:

```js
if (!fileList) {
```

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,239 @@
...
+  file_sha: string
+) {
+  const sloString = await handler.getFileContents(
+    context.github,
+    owner,
+    repo,
+    file_sha
+  );
+
+  if (sloString === null) {
```
JustinBeckwith, line 100: Same thing as above, usually better to say:

```js
if (!sloString) {
```

packages/slo-stat-bot/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "prepare": "npm run compile",
+    "pretest": "npm run compile",
+    "test": "cross-env LOG_LEVEL=fatal c8 mocha --exit build/test",
+    "test:snap": "SNAPSHOT_UPDATE=1 npm test",
+    "fix": "gts fix",
+    "lint": "gts check"
+  },
+  "dependencies": {
+    "ajv": "^6.12.2",
+    "gcf-utils": "1.6.2",
```
bcoe, line 35: Could we try using the latest version of `gcf-utils`, it's at 5.x.x.

If the update isn't trivial, please make a tracking issue, and we can upgrade later (the benefit of the latest version, is that it retries on failure).

packages/slo-stat-bot/src/slo-stat-label.ts
```diff
@@ -0,0 +1,239 @@
...
+      );
+
+      if (!fileList) {
+        return;
+      }
+
+      for (const file of fileList) {
+        //Checks to see if file is repo level or org level issue_slo_rules.json
+        if (
+          file.filename === '.github/issue_slo_rules.json' ||
```
bcoe, line 69: @JustinBeckwith @orthros, do we think that this will be where we move some of the sots of configuration that previously lived in sloth?

packages/slo-stat-bot/test/slo-stat-label.ts
```diff
@@ -0,0 +1,338 @@
...
+
+      await probot.receive({
+        name: 'pull_request.opened',
+        payload,
+        id: 'abc123',
+      });
+      requests.done();
+    });
+
+    it('No comment on PR and success check if issue_slo_rules lint is valid', async () => {
```
bcoe, line 260: nit: I usually try to make these read like a sentence:

> it('leaves no comment on the PR, and sets status to success, if issue_slo_rules lint is valid')

bcoe:
@orthros, as long as @sofisl signs off, I would vote that this is ready to merge. I think there's some foot work that will need to be done to create the corresponding app on GitHub?

packages/slo-stat-bot/package.json
hershey127: Yes for sure! I will upgrade the gcf-utils to the most recent version

packages/slo-stat-bot/test/slo-stat-label.ts
hershey127: I will fix the test sentences in the next one! (Which is coming soon)

PR #620, opened 2020-06-23
fix(merge-on-green): condition for error message

sofisl:


packages/merge-on-green/src/merge-logic.ts
```diff
@@ -727,7 +727,7 @@ export async function mergeOnGreen(
       await mergeOnGreen.merge(owner, repo, pr, prInfo, github);
       merged = true;
     } catch (err) {
-      if (err.status === 405) {
+      if (err.message.includes('not authorized to push to this branch')) {
```
tbpg, line 730: Would you mind adding a comment explaining why we can't use `err.status === 405`?

PR #625, opened 2020-06-26
feat(merge-on-green): add a second label system that makes automerging more secure

### LINKED ISSUE #579: merge-on-green: change the default behavior to not merge if new changes are pushed
Currently MoG allows the author of the PR to make additional changes, including merge, rebase, or just add new commits. While this is desired for certain use cases, it's also a potential security flaw. An author without write access can push additional changes to a PR after already receiving a review, which may be non obvious to the reviewer applying the label. 

I suggest that we make the default automerge behavior safe by default and disallow changes, and add a different label (perhaps `merge-on-green: allow changes`) that instructs MoG to allow different changes such as rebase, merge, or new changes. 

It might also be nice to have specific labels that limit what the user can do (like "only-rebase" or "only-merge") to allow for those use cases, but I'm not sure how feasible those are.

sofisl:
Fixes #579 by creating a system which accepts a new label, 'automerge: secure'

@kurtisvg, FYI!


kurtisvg:
Hey - thanks for taking the time to work on this!

One thing I do want to flag - I think it's somewhat important that the "default" behavior should be secure. Users should have to make an intentional decision is it's potentially exposing a vulnerability. Is it possible to make "automerge" more secure and then use a different label for the original behavior?

sofisl:
Hi @kurtisvg, I understand where you're coming from, but I also want to hear from others to see if this makes sense. @JustinBeckwith, @bcoe, @chingor13, what do you think about making the 'secure' functionality the default functionality?

JustinBeckwith:
I am coming at this from an entirely different angle.  *I don't think we should have this feature at all*.  The purpose of MOG is to automate what would otherwise be a human process - clicking the "merge" button when all of the CI has passed.  The moment you approve a PR, you've sent a message to the PR author that they're ok to merge, and anyone with write access can merge that PR.  Merge on green isn't doing anything the human couldn't or wouldn't do.

If I understand what you're saying right - you want to remove approval from a PR if the contents changes.  There's a setting in GitHub for that!
image

I don't think we should hold MOG to a different standard than our normal procedures here.  If the argument is we should check that checkbox for all of our repositories, let's have that chat.  But I don't think we should be pushing it into the bot right now.

kurtisvg:
@JustinBeckwith I think that the problem with depending on that feature is that you are limited to always using it or never using it. (actually, there are other problems with that feature, but I'll get to that in a second).

I'm okay with "unapproved" changes is some scenarios. Similar to internal use, I'm okay with someone with write access addressing nits and merging after getting approval, or making changes for another persons feedback after I've given review.

I'm less okay with someone pushing new changes/commits (who may not have write access, and is PRing from a fork) being able to submit unreviewed changes, as long as the tests pass. Given that the configs for the CI are usually in the repo, a malicious actor could easily open a PR for an innocuous change, get approve + automerge from someone on the ball but in a hurry, and then quickly push a set of commits that could cause tests to (almost) instantly pass and be merged. 
(Additionally, (and maybe this is a separate issue) the "automerge" label could be applied by someone with "triage" permission, which doesn't give write but does allow for the editing of issues).

The reason I'm concerned about it because I don't think the "merge even if this changes" behavior is obvious for most folks - a lot of people are using it without realizing that the author of the PR can still make/push changes and expressed surprise in this behavior. This might be because the old automerge bot used the "secure" behavior, or maybe just a side effect that people don't think about enough.

Back to the "dismiss stale pull request approvals" - having using this feature in the past, here's why I don't find it helpful:
- A change dismisses approvals for everyone - even if I'm only addressing feedback for a single person, I have to get everyone to review again
- A change dismisses all "request changes" - even if I haven't address their feedback, the request is removed and no longer "blocks" submitting
- I can't any "nits" that are small or trivial enough they shouldn't block merging
- I can't opt into the "automerge: allow changes" behavior for the times I do want to allow changes

packages/merge-on-green/src/merge-logic.ts
```diff
@@ -470,6 +477,36 @@ mergeOnGreen.getReviewsCompleted = async function getReviewsCompleted(
...
+/**
+ * Function dismisses a review for the most secure version of this bot
+ * @param owner of pr (from Watch PR)
+ * @param repo of pr (from Watch PR)
+ * @param pull_number pr number
+ * @param review_id review id of the review to dismiss
+ * @param github unique installation id for each function
+ * @returns an array of Review types
+ */
+mergeOnGreen.dismissReview = async function dismissReview(
```
JustinBeckwith, line 489: It kinda looks like this is only called from one location.  Is it worth having it's own function?  Or could you just inline the logic?

kurtisvg:
@sofisl @JustinBeckwith Sorry to be a squeaky wheel here, but it doesn't really feel like this discussion concluded. Was there a decision made somewhere else to keep the default behavior insecure?

PR #626, opened 2020-06-26
build: experiment with optional CI jobs

JustinBeckwith:
I've been thinking a lot about mono-repos lately.  Maybe it's the 1,253 active open PRs we have now.  Maybe it's the start of summer.  Either way, I found myself wondering if it's possible to have my cake and eat it too:
- I want to only run jobs for a given package if it's files changed
- I want to have split CI jobs, so we don't bundle all the checks into a single lump
- I want to have required status checks

It's not elegant right now, but I think I've checked each box:
- The interrogator looks for changed file paths, and matches it to a job that's fed into a matrix
- GitHub actions support conditional execution, but we don't even need it because of the way matrix configs work
- Required status checks are a little weirder.  For that, there's a `finale` job, that scans all jobs for the current workflow, and makes sure the required jobs match up

This may not strictly be necessary for this repository, but in places where we have monorepos and want to stay in them, this _could_ be a useful strategy.  

In case @SurferJeffAtGoogle or @dazuma are interested, plussing them in :)


.github/workflows/finale.js
```diff
@@ -0,0 +1,61 @@
...
+
+const {request} = require('gaxios');
+const args = process.argv.slice(2);
+const requiredJobs = JSON.parse(args[0]);
+const runId = process.env.GITHUB_RUN_ID;
+const repo = process.env.GITHUB_REPOSITORY;
+const githubToken = args[1];
+
+async function main() {
+  for (;;) {
```
chingor13, line 23: Is it possible that a check fails to report? It looks like the finale script will continue execute until the max execution time (if there is one)?

.github/workflows/ci.yaml
```diff
@@ -5,28 +5,27 @@ on:
...
+      goPaths: ${{ steps.interrogate.outputs.goPaths }}
+      bashPaths: ${{ steps.interrogate.outputs.bashPaths }}
+      requiredJobs: ${{ steps.interrogate.outputs.requiredJobs }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - id: interrogate
+        run: node ./.github/workflows/interrogate.js
```
tbpg, line 21: Use the same name as `director`?

.github/workflows/interrogate.js
```diff
@@ -0,0 +1,50 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const {execSync} = require('child_process');
+execSync('git fetch origin master');
+const status = execSync('git diff --name-only origin/master', { encoding: 'utf-8'});
```
bcoe, line 17: could we detect the branch being targeted with the PR, rather than assuming the default branch?

.github/workflows/finale.js
JustinBeckwith: Yeah, it would have been.  I set a (somewhat arbitrary) max execution time of 20 minutes. We can obviously play with it as we go.

.github/workflows/ci.yaml
```diff
@@ -5,28 +5,27 @@ on:
   pull_request:
 name: ci
 jobs:
+  director:
```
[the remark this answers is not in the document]
JustinBeckwith: I went with `changeFinder`

.github/workflows/ci.yaml
```diff
@@ -5,28 +5,27 @@ on:
...
+    runs-on: ubuntu-latest
+    outputs:
+      nodePaths: ${{ steps.interrogate.outputs.nodePaths }}
+      goPaths: ${{ steps.interrogate.outputs.goPaths }}
+      bashPaths: ${{ steps.interrogate.outputs.bashPaths }}
+      requiredJobs: ${{ steps.interrogate.outputs.requiredJobs }}
```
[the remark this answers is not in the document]
JustinBeckwith: Yeah, we can't.  The `changeFinder` is responsible for finding all jobs in a given category that need to be run to be considered tested.

.github/workflows/ci.yaml
```diff
@@ -70,16 +66,26 @@ jobs:
...
-        module: ["scripts"]
+        module: ${{fromJson(needs.director.outputs.bashPaths)}}
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
       - name: bashtest
         run: |
           ./test.sh
         working-directory: ./${{ matrix.module }}
+  finale:
```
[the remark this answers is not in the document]
JustinBeckwith: Heh, it's job is to make sure all of the jobs the director found actually run, and end successfully.  This is the one required status check we'd configure on the repository, and make sure passes every time.

.github/workflows/interrogate.js
JustinBeckwith: Added this in!

stephaniewang526:
@salehsquared FYI

JustinBeckwith:
@bcoe @chingor13 this is updated, and ready for a look again.  I'm not entirely attached to the idea here, but I would like to make a call on if this is the direction we want to go or not :)

bcoe (approved):

Issue #599, opened 2020-06-17
blunderbuss: assigns issue before auto-label runs


codyoss:
I currently work in a mono-repo that makes use of both `blunderbuss` and `auto-label`. I would like it if auto-label ran before bunderbuss as the assignee changes based on the product label.

Example:
- Issue is opened: https://github.com/googleapis/google-cloud-go/issues/2469
- `blunderbuss` assigns the issue to me as I am the fallback: https://github.com/googleapis/google-cloud-go/commit/9c1faca4ed66376b73e1bd7a39076c5838ae7f52#diff-30872c876afbce72ad8d68bc80238c6cR22
- `auto-label` adds the `api: storage` label, as it should have

If `blunderbuss` ran after `auto-label` it would have correctly assigned someone else: https://github.com/googleapis/google-cloud-go/commit/9c1faca4ed66376b73e1bd7a39076c5838ae7f52#diff-30872c876afbce72ad8d68bc80238c6cR8

tbpg:
Two ideas:
* If there is an `assign_issues_by` config, wait N seconds before fetching assignees and checking there are still none.
* Import the auto-label logic and don't assign the issue if it's about to be labeled. The label action will retrigger Blunderbuss.

PR #630, opened 2020-06-29
feat(gcf-utils): add ability to delay processing time of tasks

codyoss:
Updates: #599


codyoss:
The delay could happen in blunderbuss and it would have the same effect. Although, if it is implemented at that level it will have to just stall the runner where as this would keep the execution time down and build the delay into when the task is scheduled to run. 

>  It only needs to happen for certain events, not labeled, so this seems like overkill.

That is a good point, I will open up a new PR.

PR #631, opened 2020-06-29
chore(deps): upgrade to nock 13

JustinBeckwith:


packages/merge-on-green/src/merge-logic.ts
```diff
@@ -640,7 +640,7 @@ mergeOnGreen.removeLabel = async function removeLabel(
   github: GitHubAPI
 ) {
   try {
-    github.issues.removeLabel({
+    await github.issues.removeLabel({
```
JustinBeckwith, line 643: @sofisl look what I found!  A bug was hiding here!

sofisl: Beautiful :) Thank you!!

Issue #611, opened 2020-06-19
auto-label: add nightly cron for multiple purposes


sofisl:
Now that we have changed the infrastructure for our bots, let's add a nightly cron for label sync that:

1. Labels unlabeled issues for a repo (backfills) (P2)
2. Relabels an issue if there is a new relationship of repo to api (i.e., either the repo changes to a more specific api or the repo gets a new API) (P2)
3. Ensure that the appropriate label exists for a given repo for future (P1)
4. Checks that labels are only specific to the appropriate repo/API relationship if a 1:1 relationship exists (only for monorepos). Can perform batch request to see if there's any other API label instead of performing request on many repos. (P3)

@bcoe and @martinvarelaj - feel free to add anything else as needed.

Issue #617, opened 2020-06-22
Auto-add `sample` label for repositories with `-samples` in the name


JustinBeckwith:
We'd like to start specifically issues related to samples across the platform.  To make that easier, I'd like to ask for two things:
- I would like to add a `sample` label to every repositories labels list
- For repositories with `-samples` in the name, I would like to automatically add a `sample` label to every issue

PR #633, opened 2020-06-30
feat(auto-label): add nightly cron job

sofisl:
Fixes all the issues in #611 by adding a nightly cron job, cleans up other apis


packages/auto-label/src/auto-label.ts
```diff
@@ -51,6 +56,27 @@ handler.addLabels = async function addLabels(
   }
 };
 
+handler.removeLabels = async function removeLabels(
```
tbpg, line 59:
```suggestion
handler.removeLabel = async function removeLabel(
```

packages/auto-label/src/auto-label.ts
```diff
@@ -29,8 +29,13 @@ interface JSONData {
   repo: string;
 }
 
+interface Labels {
```
tbpg, line 32:
```suggestion
interface Label {
```

packages/auto-label/src/auto-label.ts
```diff
@@ -211,22 +236,128 @@ handler.autoDetectLabel = (
   )?.github_label;
 };
 
-const BACKFILL_LABEL = 'auto-label:backfill';
+handler.mainLogic = async function mainLogic(
```
tbpg, line 239: Is there a more descriptive name for this function? `handler` also has a decent amount of logic, so it's not clear from the names.

packages/auto-label/test/auto-label.ts
```diff
@@ -275,20 +275,164 @@ describe('auto-label', () => {
...
+            id: 1811802233,
+            node_id: 'MDU6TGFiZWwxODExODAyMjMz',
+            url:
+              'https://api.github.com/repos/sofisl/mergeOnGreenTest/labels/anotherLabel',
+            name: 'myGitHubLabel',
+            color: 'FEFEFA',
+            default: false,
+            description: null,
+          },
+        ])
```
tbpg, line 305: Do we need full replies? Not sure what the best practice is.

packages/auto-label/test/auto-label.ts
```diff
@@ -275,20 +275,164 @@ describe('auto-label', () => {
...
+              'https://api.github.com/repos/sofisl/mergeOnGreenTest/labels/anotherLabel',
+            name: 'myGitHubLabel',
+            color: 'FEFEFA',
+            default: false,
+            description: null,
+          },
+        ]);
+      handler.callStorage = async () => downloadedFile;
+      await probot.receive({
+        name: 'schedule.repository',
```
tbpg, line 321: What triggers these events? Also, does this mean every old issue will be automatically labeled? Seems like it would only need to be done once per repository, not nightly?

packages/auto-label/src/auto-label.ts
sofisl: Good point! Essentially, the handler function deals with the different logic that can trigger the main function, which is eventually adding a label to the repo and/or issue. I added a different title, hopefully it's a little better :)

packages/auto-label/test/auto-label.ts
sofisl: Good call! Shortened them.

packages/auto-label/test/auto-label.ts
sofisl: Ah, good catch. So, in theory, this function was supposed to run nightly for every repo, for two reasons:

1. We wanted to institute a solution that eliminated the need for a 'backfill' label, and 
2. We wanted to make sure repos had the appropriate labels if their relationship to a product changed (i.e., if they covered a more specific product, we wanted the label to be updated and the old label to be removed).

But, in thinking through this a little more, I realized this would be wasteful and time consuming - both of these cases are fairly infrequent, to your point. So, I implemented a solution that writes the repository names to a file when it runs, and the function checks if the repo name already exists on that file. If it does, then the nightly function won't run for that repo. This way, we're only catching new repos or newly-named repos (i.e., in the case that we further specify a repo).

I'm not sure if this solution would work; take a look and let me know what you think!

sofisl:
@JustinBeckwith, once I land this, I'll start on #617!

packages/auto-label/src/auto-label.ts
```diff
@@ -211,66 +240,175 @@ handler.autoDetectLabel = (
...
-      app.log.info(
-        `ignoring non-backfill label action (${context.payload.issue.labels.join(
-          ', '
-        )})`
-      );
       return;
     }
 
+    console.log(path.join(__dirname, './repos.txt'));
+    fs.writeFileSync(
```
tbpg, line 375: This seems like it writes to the Cloud Function's local file system? The file system is not shared between Cloud Functions instances, so I don't think this will work.

It would be OK with me to run over every issue every time if we know that it's correct.

sofisl: I was worried about that... thanks for confirming!

Well, to your point, it's every issue for every one of our repos.

Is it worthwhile/are there any disadvantages to writing these repos to Datastore? My only concerns here are that (i) I feel like this solution would just replicate repos.json in sloth (which is something we're trying to eliminate), and (ii), a repo may change a product overview without actually changing its title, in which case this cron job wouldn't run for it.

packages/auto-label/repos.txt
```diff
@@ -0,0 +1 @@
+testOwner/testRepo
```
JustinBeckwith, line 1: Kind of a weird thing to have hanging out in the root of the project :) I'm curious what this will be used for!

packages/auto-label/src/auto-label.ts
```diff
@@ -18,19 +18,28 @@ import {Storage} from '@google-cloud/storage';
 import {Application} from 'probot';
 // eslint-disable-next-line node/no-extraneous-import
 import {GitHubAPI} from 'probot/lib/github';
-// eslint-disable-next-line node/no-extraneous-import
-import Webhooks from '@octokit/webhooks';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const colorsData = require('./colors.json');
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
JustinBeckwith, line 24: For both `fs` and `path`, you should use `import * as fs from 'fs'` style syntax.

packages/auto-label/src/auto-label.ts
```diff
@@ -211,66 +240,175 @@ handler.autoDetectLabel = (
   )?.github_label;
 };
 
-const BACKFILL_LABEL = 'auto-label:backfill';
+handler.addLabeltoRepoAndIssue = async function addLabeltoRepoAndIssue(
+  owner: string,
+  repo: string,
+  issueNumber: number,
+  context: any,
```
JustinBeckwith, line 247: Is this truly an `any`?  Anything more specific we can do here?  As an aside - let's please treat linter warnings as errors.  This is something I intend to fix in the next release of `gts`.

packages/auto-label/src/auto-label.ts
```diff
@@ -211,66 +240,175 @@ handler.autoDetectLabel = (
...
+      colorsData[colorNumber].color
+    );
+  } else {
+    console.log(
+      'This label already exists on the repository, will check if it also exists on the issue'
+    );
+  }
+
+  if (labelsOnIssue) {
+    const found = labelsOnIssue.find(
```
JustinBeckwith, line 320: This is the kind of thing where I'd expect the type of `labelsOnIssue` to have nice static typing, so you don't have to inline the `{name: string}` type you have beow.

packages/auto-label/src/auto-label.ts
```diff
@@ -211,66 +240,175 @@ handler.autoDetectLabel = (
...
+
+  return;
+};
+
+//main function, responds to label being added
 function handler(app: Application) {
-  app.on(['issues.labeled'], async context => {
-    const owner = context.payload.repository.owner.login;
+  //nightly cron that backfills and corrects api labels
+  app.on(['schedule.repository'], async context => {
```
JustinBeckwith, line 358: Do we also have an event for when the app is installed?  Thinking that right after install is the right time to do the first creation of labels, and then the cron to keep them up to date as we go feels right.

packages/auto-label/src/auto-label.ts
```diff
@@ -211,66 +240,175 @@ handler.autoDetectLabel = (
...
     ) {
-      app.log.info(
-        `ignoring non-backfill label action (${context.payload.issue.labels.join(
-          ', '
-        )})`
-      );
       return;
     }
 
+    console.log(path.join(__dirname, './repos.txt'));
```
JustinBeckwith, line 374: Calling @bcoe :) Don't you have some code lying about that fetches the list of repositories on which the app is installed?  That's what we really want here.

sofisl: Added in installation event. Thanks!

packages/auto-label/src/auto-label.ts
sofisl: Added an installation event!

PR #635, opened 2020-06-30
feat(release-please): run nightly around 1 am on a cron

chingor13:
This will fix 2 minor issues:

* Java releases make a snapshot release after a real release - currently there is a race condition for tagging the release/checking if we should run release-please. This will ensure the follow-up PR is created no matter what.
* The release notes and changelog include the release date. If we wait a few days to merge a release PR, it will show the wrong release date.

Note: that this will close/reopen every open release PR nightly...


chingor13:
After 5.5.0 is released, I'll redo this PR and change the published job trigger to cron and force snapshot only

PR #640, opened 2020-07-01
feat(commit-lint): use title if automerge label

bcoe:
Since automerge uses the title (not the commit text) we can lint the title rather than the commit when:

1. there's only one commit, but.
2. there's an automerge label.


packages/conventional-commit-lint/test/conventional-commit-lint.ts
```diff
@@ -98,7 +100,9 @@ describe('ConventionalCommitLint', () => {
         './pull_request_synchronize'
       ));
       // create a history that has one valid commit, and one invalid commit:
-      const invalidCommits = require(resolve(fixturesPath, './invalid_commit'));
+      const invalidCommits = [
+        ...require(resolve(fixturesPath, './invalid_commit')),
```
bcoe, line 104: because Node.js caches `require` I was actually modifying the initial require in each test.

PR #645, opened 2020-07-02
feat(gcf-utils,label-sync)!: use /installation/repositories to get repo list

### LINKED ISSUE #623: Drop dependency on `repos.json`
We are trying to [delete repos.json](https://github.com/googleapis/sloth/issues/730) from sloth.  This repository takes a dependency on the file, even though it _probably_ shouldn't be using it.  Adding this to track the work.  

## Label Sync bot
The label sync bot uses `repos.json` to find all repositories it should update when the main list of available labels changes.  It wants to go to every repository, and trigger an update of its labels.  Using `repos.json` for this is not good, because it will miss repositories that have the bot installed, but aren't in `repos.json`.  The offending code is here:
https://github.com/googleapis/repo-automation-bots/blob/2e999b00724d1c1b9ad7a6d2922ad32f09c140cb/packages/label-sync/src/label-sync.ts#L94-L109

I am pretty sure we should instead interrogate GitHub to find which repositories have the `label-sync` application installed instead.  I think you use this API to do that:
https://developer.github.com/v3/apps/installations/#list-repositories-accessible-to-the-app-installation

The challenge is that this API requires an [installation access token](https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps/#authenticating-as-an-installation), and I have little idea how Probot manages getting it.  Some poking about landed me on this issue:
https://github.com/probot/probot/issues/1003

It looks like we would want to do something like:

```js
// Find every repository that has this GitHub Application installed.
// To access the apps/installation list endpoint, you need to have a
// JWT obtained via createInstallationToken.
const github = await app.auth();
const token = await github.apps.createInstallationToken({
  installation_id: context.payload.installation.id
});
const ok = new Octokit({
  auth: token,
});
const res = await ok.apps.listRepos();
```

I ran into issues with `context.payload` not having an `installation` property, so someone will need to dig further into this.  I also have no earthly idea how I would have tested my code. 

## GCF Utils
The same story plays out in GCF Utils, specifically while figuring out which repositories to run cron jobs against.  *We shouldn't use repos.json for this, rather we should use the list of repositories on which the app is installed*.  The offending code is here:
https://github.com/googleapis/repo-automation-bots/blob/2e999b00724d1c1b9ad7a6d2922ad32f09c140cb/packages/gcf-utils/src/gcf-utils.ts#L198-L205

Independent of my desire to drop `repos.json`, I think we should clean these up because it creates unintended consequences for consumers of the app.  It's entirely unclear that I need to be registered in sloth in order to use one of these bots.

bcoe:
Rather than using a central `repos.json` pulled from sloth, this will now use the GitHub API to pull al list of repos.

**Danger, Will Robinson 🤖 we need to make sure all the bots are moved to v3 of gcf-utils, and that it's tested, before we delete repos.json from sloth.**

This drops the `push` hook from label-sync, assuming that we will instead have a cron that triggers the sync (_@JustinBeckwith at this point crons are like magic, and behave exactly like a normal repo hook, the fan out of the cron is handled in gcf-utils_).

Fixes #623


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -15,8 +15,8 @@
 import {createProbot, Probot, ApplicationFunction, Options} from 'probot';
 import {CloudTasksClient} from '@google-cloud/tasks';
 import {v1} from '@google-cloud/secret-manager';
-import {request} from 'gaxios';
 import * as express from 'express';
+import {Octokit} from '@octokit/rest';
```
alexander-fenster, line 19: missing dependency?

bcoe: this is a peer dependency in `probot`, so I think we're guaranteed to have it.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -195,17 +198,26 @@ export class GCFBootstrapper {
...
-      const url =
-        'https://raw.githubusercontent.com/googleapis/sloth/master/repos.json';
-      const res = await request<Repos>({url});
-      const {repos} = res.data;
-      for (const repo of repos) {
-        await this.scheduledToTask(repo.repo, id, body, eventName, signature);
+      const octokit = new Octokit({
+        auth: await this.getInstallationToken(body.installation.id)
+      });
+      const {data} = await octokit.request('/installation/repositories', {
```
chingor13, line 204: nit: can you add a link to https://developer.github.com/v3/apps/installations/

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -31,6 +31,9 @@ interface Repos {
 
 interface Scheduled {
   repo?: string;
+  installation: {
+    id: number;
+  }
```
chingor13, line 36: What's populating this field?

chingor13:
Also, does this mean we may run on unexpected repositories within the organization that were not explicitly in the sloth list? This could be intrusive for apps installed at the org level that were only taking action on repos configured in sloth.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -195,17 +199,33 @@ export class GCFBootstrapper {
...
       await this.scheduledToTask(body.repo, id, body, eventName, signature);
     } else {
-      // Job should be run on all managed repositories:
-      const url =
-        'https://raw.githubusercontent.com/googleapis/sloth/master/repos.json';
-      const res = await request<Repos>({url});
-      const {repos} = res.data;
-      for (const repo of repos) {
-        await this.scheduledToTask(repo.repo, id, body, eventName, signature);
+      const octokit = new Octokit({
```
JustinBeckwith, line 202: There are other bots that want to ask for all repositories where the bot is installed.  Would it make sense to lift this up to a utility function?

JustinBeckwith:
Another note: I don't believe this truly fixes #623.  We need to release gcf-utils, update everyone, and then update the code in autolabel and a few others that also query repos.json.

PR #646, opened 2020-07-02
refactor(failurechecker): clean up dependencies and exports

JustinBeckwith:
This is me poking at a few things which may not need to be poked.  For some reason I'm bothered by the way we need to pin all of the exportable methods to a single function.  It makes the code harder to read in other bots.  I kind of wanted to test this out here, and see if I could make things work without doing it.  Someone please check me to make sure I'm not breaking something :)

Second - cleaning up some dependencies.

### DIFF
--- a/packages/failurechecker/package.json
+++ b/packages/failurechecker/package.json
@@ -28,12 +28,10 @@
     "lint": "gts check"
   },
   "dependencies": {
-    "gcf-utils": "1.6.1",
-    "moment": "2.27.0"
+    "gcf-utils": "1.6.1"
   },
   "devDependencies": {
     "@types/bunyan": "^1.8.6",
-    "@types/dotenv": "^6.1.1",
     "@types/express": "^4.17.0",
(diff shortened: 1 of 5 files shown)


packages/failurechecker/src/app.ts
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import {GCFBootstrapper} from 'gcf-utils';
-import appFn from './failurechecker';
+import {failureChecker} from './failurechecker';
```
bcoe, line 16: seems reasonable to me.

packages/failurechecker/package.json
```diff
@@ -28,28 +28,23 @@
     "lint": "gts check"
   },
   "dependencies": {
-    "gcf-utils": "1.6.1",
-    "moment": "2.27.0"
+    "gcf-utils": "1.6.1"
```
bcoe, line 31: let's use the latest version of `gcf-utils`.

Issue #653, opened 2020-07-06
Provide a standardized logger for automation bots


azizsonawalla:
Provide a standardized logger for repo-automation-bots to use for basic logging as well as logging metrics related data.

**Problem**: automation bots don't have a standardized way to log information and use `console.log()` with non-standard strings instead

**Proposed Solution**: provide a logger from within gcf-utils that bots can import and use

PR #654, opened 2020-07-06
feat(gcf-utils): provide a standardized logger for repo automation bots

azizsonawalla:
Provides a new `GCFLogger` from within gcf-utils that can be imported by automation bots for logging purposes. 

Fixes https://github.com/googleapis/repo-automation-bots/issues/653


packages/gcf-utils/test/gcf-utils.ts
```diff
@@ -33,311 +36,374 @@ function nockRepoList() {
...
-      req.headers['x-cloudtasks-taskname'] = 'my-task';
-
-      await handler(req, response);
-
-      sinon.assert.calledOnce(configStub);
-      sinon.assert.notCalled(sendStatusStub);
-      sinon.assert.calledOnce(sendStub);
-      sinon.assert.calledOnce(spy);
-    });
+  describe('GCFBootstrapper', () => {
```
azizsonawalla, line 101: Existing tests were moved into the nested describe block `GCFBootstrapper`, hence the high number of changed lines

chingor13: Do these tests need to live in the same test file? Maybe it's worthwhile to rename the `gcf-utils.ts` to `gcf-bootstrapper.ts` and have a separate `gcf-logger.ts` test.

It would also make the diff easier to read.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -41,6 +48,44 @@ interface EnqueueTaskParams {
...
+    dest?: pino.DestinationStream
+  ): pino.Logger {
+    const defaultOptions: pino.LoggerOptions = {
+      customLevels: {
+        metric: 30,
+      },
+      level: 'trace',
+    };
+    if (options) {
+      Object.assign(defaultOptions, options);
```
chingor13, line 80: I think there's a typescript equivalent for this:
```
options = {...defaultOptions, ...options};
```
or something like that.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -41,6 +48,44 @@ interface EnqueueTaskParams {
...
+  public static get(): pino.Logger {
+    if (!this.logger) {
+      this.logger = this.initLogger();
+    }
+    return this.logger;
+  }
+
+  private static initLogger(
+    options?: pino.LoggerOptions,
+    dest?: pino.DestinationStream
```
chingor13, line 71: Are these options used anywhere?

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: cool, that looks more elegant. I'll take a look into that

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: these params are used by the tests to inject a mock/custom destination. This could also be made public in the future if we wanted to support bots adding multiple destinations for their logs

azizsonawalla: oh I see what you mean with the options param - right, I was using them in the tests before but not anymore. I can take that out

packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
+// Copyright 2020 Google LLC
```
chingor13, line 1: Since this file is a git mv in history, you should keep the 2019

packages/gcf-utils/test/gcf-logger.ts
```diff
@@ -0,0 +1,81 @@
...
+import {ObjectWritableMock} from 'stream-mock';
+import pino from 'pino';
+import {validateLogs, LogLine} from './test-helpers';
+
+describe('GCFLogger', () => {
+  describe('get()', () => {
+    it('returns a new pino-based logger instance on first call', () => {
+      const logger = GCFLogger.get();
+      assert.equal(logger.constructor.name, 'Pino');
+    });
```
chingor13, line 27: Personally, I think we should test behavior of the logger and not the class of the implementation.

packages/gcf-utils/test/gcf-logger.ts
```diff
@@ -0,0 +1,81 @@
...
+      const logger2 = GCFLogger.get();
+      assert.deepEqual(logger1, logger2);
+    });
+  });
+
+  describe('logger instance', () => {
+    let writeStream: ObjectWritableMock;
+    let logger: pino.Logger;
+
+    function readLogsAsObjects(writeStream: ObjectWritableMock): LogLine[] {
```
chingor13, line 39: `writeStream` looks re-used or overloaded.

packages/gcf-utils/test/integration/gcf-logger-integration.ts
```diff
@@ -0,0 +1,82 @@
...
+import {validateLogs, LogLine} from '../test-helpers';
+import SonicBoom from 'sonic-boom';
+import fs from 'fs';
+
+describe('GCFLogger Integration', () => {
+  let logger: pino.Logger;
+  const testStreamPath = './test-stream.txt';
+  let writeStream: SonicBoom;
+
+  function readLogsAsObjects(writeStream: SonicBoom): LogLine[] {
```
chingor13, line 27: Same comment here on re-using the variable name.

packages/gcf-utils/test/integration/gcf-utils-integration.ts
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
+// Copyright 2020 Google LLC
```
chingor13, line 1: This should be 2019

packages/gcf-utils/test/test-helpers.ts
```diff
@@ -0,0 +1,74 @@
...
+): void {
+  if (expectedLineCount) {
+    assert.equal(
+      logs.length,
+      expectedLineCount,
+      `expected exactly ${expectedLineCount} line(s) to be logged`
+    );
+  }
+  if (expectedMessages) {
+    for (let i = 0; i < expectedMessages.length; i++) {
```
chingor13, line 46: Is is possible that we log in separate async calls so that the order is non-deterministic? Do we want to assert on the exact order?

packages/gcf-utils/test/gcf-bootstrapper.ts
azizsonawalla: Got it - I wasn't sure if a move counted as a new file for the license checker bot or not

chingor13: If the bot flags it, then the bot is wrong and we can fix

packages/gcf-utils/test/gcf-logger.ts
azizsonawalla: Well `GCFLogger.get()` has a return type of `pino.Logger` and so the caller expects everything that comes with a pino logger. So if we were to swap this out in the future for say Bunyan, and there were some Pino features missing in Bunyan, this test would tell us it's a breaking change. 

I could on the other hand test all the Pino features, but that would just be duplicating Pino's own tests and we'd have to add more every time they add more features.

Lastly, I did try to check `logger instanceof pino.Logger` but it wouldn't let me do that because `pino.Logger` is a `type` rather than a `class` or `interface`

packages/gcf-utils/test/test-helpers.ts
azizsonawalla: In this case we do actually want to assert on the order because we're testing a synchronous logger (we pass `{sync:true}` to pino) so it should be flushing after each log statement.

packages/gcf-utils/test/integration/gcf-utils-integration.ts
azizsonawalla: Will do

packages/gcf-utils/test/gcf-logger.ts
azizsonawalla: If you feel like we don't need to check the type for the logger, then I can remove this test. In fact, in that case, I might change it so that GCFLogger extends pino.Logger and then we completely abstract away pino from the bots

chingor13: With interfaces, the goal is that it shouldn't matter what the underlying implementation class is called. In some rare cases, you may want to know what the implementation is, but in this case, it really seems like we only care about the behavior.

If you do change the underlying implementation, tests usually shouldn't need to be rewritten to match the interface (sometimes mocks need to change though). Refactoring to extend pino.Logger, assuming it provides the same interface, shouldn't require touching the tests.

packages/gcf-utils/package.json
```diff
@@ -28,9 +28,12 @@
     "@types/ioredis": "^4.14.8",
     "@types/lru-cache": "^5.1.0",
     "@octokit/plugin-enterprise-compatibility": "1.2.5",
+    "@types/pino": "^6.3.0",
     "express": "^4.17.1",
     "gaxios": "^3.0.0",
+    "pino": "^6.3.2",
     "probot": "^9.11.2",
+    "sonic-boom": "^1.0.1",
```
bcoe, line 36: It looks like this could be a development dependency, as it's only used for types in tests (_pino will pull in this dep for you_).

It might be worth simply assuming you will have `sonic-boom` available, through `pino`, and simply adding an ignore rule for linting.

packages/gcf-utils/package.json
```diff
@@ -28,9 +28,12 @@
     "@types/ioredis": "^4.14.8",
     "@types/lru-cache": "^5.1.0",
     "@octokit/plugin-enterprise-compatibility": "1.2.5",
+    "@types/pino": "^6.3.0",
```
bcoe, line 31: I'm surprised so many of our types are `dependencies`, vs., `devDependencies`. @orthros remind me if there was a reason for this?

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -45,6 +52,38 @@ interface EnqueueTaskParams {
...
+    const defaultOptions: pino.LoggerOptions = {
+      customLevels: {
+        metric: 30,
+      },
+      level: 'trace',
+    };
+    if (!dest) {
+      dest = pino.destination({sync: true});
+    }
+    return pino(defaultOptions, dest);
```
bcoe, line 83: I agree with @chingor13 on this topic, my advise would be to encapsulate `pino`, and expose an interface like:

```
logger.info()
logger.warn()
logger.error()
```

vs., returning an instance of pino.

packages/gcf-utils/package.json
chingor13: I think @JustinBeckwith just did that because users of gcf-utils generally needed those types (#647)

packages/gcf-utils/README.md
```diff
@@ -4,6 +4,22 @@
...
 ## Usage
 
+### GCFLogger
+
+`GCFLogger` is a [Pino](https://github.com/pinojs/pino)-based logger with added configurations for Google Cloud Functions. Note that this logger is a singleton instance shared across all modules.
+
+```
+import {GCFLogger} from 'gcf-utils';
+
+let logger = GCFLogger.get();
```
JustinBeckwith, line 14: I'm surprised you're allowed to even have a method named `get` in JavaScript!  From a usability perspective, I'd rather have:

```ts
import {getLogger} from 'gcf-utils';
const logger = getLogger();
```

packages/gcf-utils/package.json
JustinBeckwith: Yup.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -45,6 +52,38 @@ interface EnqueueTaskParams {
   name: string;
 }
 
+/**
+ * A pino-based singleton logger for use within Google Cloud Functions.
+ * Provides a metric() method to log metrics-related data
+ */
+export class GCFLogger {
```
JustinBeckwith, line 59: Is there a reason for this to exist as a class?  I'd rather have a `getLogger()` method which is directly exported, unless there is a reason for encapsulation, or private data in the class instance.

packages/gcf-utils/test/gcf-logger.ts
azizsonawalla: Got it! My initial goal was to abstract away Pino completely but it would mean adding wrappers to all of Pino's methods. But I see that @bcoe has suggested exactly that below.

I'll incorporate both your comments here and push another commit to fix this. Thanks both!

packages/gcf-utils/package.json
azizsonawalla: Yeah the linting gave me an `extraneous dependency` error for this so I had to list it out. But since we're abstracting away Pino now, I don't think I'll need this here anyway

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: Yep, sounds good. I planned to do this initially but thought it might be too much wrapping, but since you and @chingor13 both seem to like this approach, I'll go ahead and do that. Makes for more robust code I think.

packages/gcf-utils/README.md
azizsonawalla: Oops! Didn't think of that. I'm not super familiar with JavaScript conventions. In Java world this would be `getInstance()` so I think `getLogger()` makes sense. I'll fix it - thanks for the catch!

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: As @bcoe and @chingor13 mentioned above, we want to wrap Pino and abstract it away, hence the GCFLogger class. The purpose of this class is not apparent right now since I haven't actually hidden Pino completely here, but I'll be doing that on the next commit

packages/gcf-utils/package.json
bcoe: sounds good.

packages/gcf-utils/README.md
```diff
@@ -4,6 +4,22 @@
 
 ## Usage
 
+### GCFLogger
+
+`GCFLogger` is a standardized logger for Google Cloud Functions. Note: this logger is a singleton shared across the
```
chingor13, line 9: nit: hanging sentence

packages/gcf-utils/README.md
```diff
@@ -4,6 +4,22 @@
...
 ## Usage
 
+### GCFLogger
+
+`GCFLogger` is a standardized logger for Google Cloud Functions. Note: this logger is a singleton shared across the 
+
+```
+import {GCFLogger, getLogger} from 'gcf-utils';
+
+let logger: GCFLogger = getLogger();
```
chingor13, line 14:
```suggestion
const logger = getLogger();
```

You shouldn't have to declare the type as the compiler can figure it out from the assignment.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -45,6 +47,59 @@ interface EnqueueTaskParams {
...
+  dest?: NodeJS.WritableStream | SonicBoom
+): GCFLogger {
+  const defaultOptions: pino.LoggerOptions = {
+    customLevels: {
+      metric: 30,
+    },
+    level: 'trace',
+  };
+  const defaultDestination = pino.destination({sync: true});
+  const logger = pino(defaultOptions, dest || defaultDestination);
```
chingor13, line 92: Should we only instantiate the defaultDestination if none is provided? If one is provided, we're immediately throwing away the default one we just instantiated.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -45,6 +47,59 @@ interface EnqueueTaskParams {
   name: string;
 }
 
+interface LogFn {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
azizsonawalla, line 51: I've disabled linting for the `any[]` in these lines because this is the signature for the underlying Pino logger and this seemed like the most elegant way to wrap it. If we are more specific about types then we'll have to add a bunch of code to transform the inputs before passing to Pino

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -45,6 +47,59 @@ interface EnqueueTaskParams {
...
+): GCFLogger {
+  const defaultOptions: pino.LoggerOptions = {
+    customLevels: {
+      metric: 30,
+    },
+    level: 'trace',
+  };
+  const defaultDestination = pino.destination({sync: true});
+  const logger = pino(defaultOptions, dest || defaultDestination);
+  return {
```
azizsonawalla, line 93: Pino contains a lot more properties than GCFLogger so I'm only returning the properties here relevant to GCFLogger

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: good point. I did this from a readability perspective but didn't think of the unnecessary instantiation

packages/gcf-utils/README.md
azizsonawalla: sure, I can fix that. On second thought, since we're now not in a class anymore, any point having a `getLogger()` method or can I simply just export a logger const?

JustinBeckwith: I was having the exact same thought

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -45,6 +47,48 @@ interface EnqueueTaskParams {
...
+  (msg: string, ...args: any[]): void;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (obj: object, msg?: string, ...args: any[]): void;
+}
+
+/**
+ * A logger standardized logger for Google Cloud Functions
+ */
+export interface GCFLogger {
+  [key: string]: Function;
```
chingor13, line 61: Do we need this? It looks like it's used for tests to access the log functions by grabbing the function via js object properties.

Is there a better way to access that in TS? @bcoe @JustinBeckwith

packages/gcf-utils/src/gcf-utils.ts
chingor13: If we can remove the generic string->Function type, we should be able to directly return the Pino logger (as it will match the GCFLogger interface) so we can avoid returning a re-bound object.

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: This is just for testing, specifically here to do:

```typescript

export const logLevels: {[index: string]: number} = {
  trace: 10,
  debug: 20,
  info: 30,
  metric: 30,
  warn: 40,
  error: 50,
};

for (let level of Object.keys(logLevels)) {
  logger[level](`log a statement of severity ${level$}`);
}
```

This makes testing a lot more concise, but if there's a workaround for this I'm happy to remove the mapping

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: So actually there's 2 reasons to not remove ``` [key: string]: Function ``` other than it helping with testing...

1. We won't be able to directly return `logger` if we remove the above mapping. This is because `GCFLogger` has function `metric()` which Pino only has as a custom level. We get this error when we remove the mapping and directly return logger: 
```
Property 'metric' is missing in type 'BaseLogger & { [key: string]: LogFn; }' but required in type 'GCFLogger'
```
(`BaseLogger & { [key: string]: LogFn; }` is the interface that Pino implements)

2. In the next step of my implementation I need to add a `flush` function to GCFLogger which will not be a direct binding like the other logging functions are. For `flush` I'll be passing a custom function that checks the destination type and flushes accordingly. So in this case I can't pass the logger directly anyway

packages/gcf-utils/src/gcf-utils.ts
JustinBeckwith: I would really avoid doing that here.  Adding `[key: string]: Function;` will really water down the effectiveness of the type safety provided, and make it possible to do all kinds of stuff that would fail at runtime, but the compiler would happily allow.  

If the only reason you want this is for testing, I would do a cast in the test itself to something looser, while not loosening the types for library consumers.

azizsonawalla: Makes sense - I'll try to see if I can make it work without the `[key: string]: Function` since it's being used for casting Pino as well other than just a testing help. I think I have some workarounds that might do.

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: Ok I've removed `[key:string]: Function` from the `GCFLogger` interface but that doesn't let us return the `logger` as is for the issues mentioned above. 

I've also tried doing:
```typescript
return {...logger, metric: logger.metric.bind(logger)}
```
but that doesn't work either because (due to the way Pino is setup) we need to bind the other functions as well to the logger object. Doing the above satisfies the compiler but results in runtime issues within Pino because the other logging functions like info, trace, etc. don't have the logger object bound to it

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -45,6 +47,45 @@ interface EnqueueTaskParams {
...
+}
+
+export const logger: GCFLogger = initLogger();
+
+export function initLogger(
+  dest?: NodeJS.WritableStream | SonicBoom
+): GCFLogger {
+  const defaultOptions: pino.LoggerOptions = {
+    customLevels: {
+      metric: 30,
```
bcoe, line 76: perhaps a comment on this line, what does the 30 represent?

packages/gcf-utils/test/gcf-logger.ts
```diff
@@ -0,0 +1,64 @@
...
+        writeStream.end();
+        const lines: string[] = writeStream.data;
+        return lines.map(line => JSON.parse(line));
+      } catch (error) {
+        throw new Error(`Failed to read stream: ${error}`);
+      }
+    }
+
+    function testAllLevels() {
+      for (const level of Object.keys(logLevels)) {
```
bcoe, line 36: I tend to not like code to generate test cases ... but this is fairly elegant, and I like it 😆

packages/gcf-utils/test/integration/gcf-logger-integration.ts
```diff
@@ -0,0 +1,80 @@
...
+            logLevels[level]
+          );
+          done();
+        });
+      });
+    }
+  }
+
+  beforeEach(() => {
+    destination = pino.destination(testStreamPath);
```
bcoe, line 71: Does this imply that `pino` is a singleton, and this will change the destination globally?

packages/gcf-utils/test/gcf-logger.ts
```diff
@@ -0,0 +1,64 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {GCFLogger, initLogger} from '../src/gcf-utils';
+import {describe, beforeEach, it} from 'mocha';
+import {ObjectWritableMock} from 'stream-mock';
```
bcoe, line 17: This seems like really good approach for testing. My one concern is if there is a possibility that there is some async work in the event loop, when our GCF function finishes, when we use the `sonic-boom` transport.

Is there any way we could get an `await`able Promise from `sonic-boom`, that resolves when logs have actually written to their intended destination? I think we should create a tracking issue, and test the heck out of the behavior of logs that occur immediately before a response from a cloud function.

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: 30 is the severity level assigned to `info` by GCP and we want `metric` to basically be an alias for `info` at this point. Here's the complete list https://github.com/googleapis/repo-automation-bots/blob/d9f4be19dd844b9cdc69895c4fd583c942e6f6a6/packages/gcf-utils/test/test-helpers.ts#L17-L24

packages/gcf-utils/test/gcf-logger.ts
azizsonawalla: haha it saves a lot of duplicated code, and considering we might add more custom levels like `metric` in the future, it'll be much easier to add test cases for those in this format (just have to add it to `logsLevel`)

packages/gcf-utils/test/integration/gcf-logger-integration.ts
azizsonawalla: no the `destination()` method is just a constructor to create a `SonicBoom` wrapped file stream. We then pass this destination to the `initLogger()` function that builds the logger using this destination.

packages/gcf-utils/test/gcf-logger.ts
azizsonawalla: So there's 2 points here:

1. The way we're creating a logger right now is by passing the `{ sync: true }` option to `SonicBoom` which means that the logs will be flushed each time a log statement is written. So there's no async work happening in the background.

2. For logging within the `gcf()` function (coming in next PR) where we log the trigger info and github actions, we call `flushSync()` on SonicBoom at the end so that it synchronously flushes everything - so again, no async work.

This of course does come with some performance hit but in our case I think having all the logs is more important than performance (considering most bots are well below the GCF timeout). 

Either way, let's definitely test the heck out of this

PR #656, opened 2020-07-06
feat(blunderbuss): add delay and label fetch for new issues

codyoss:
Fixes: #599


packages/blunderbuss/src/blunderbuss.ts
```diff
@@ -156,7 +160,21 @@ export = (app: Application) => {
         return;
       }
 
-      const labels = issue.labels?.map(l => l.name);
+      let labels: string[] = [];
+      if (byConfig !== undefined) {
```
tbpg, line 164: Should we also check `context.payload.action == 'issues.opened'`?

codyoss: Good call, updated.

PR #658, opened 2020-07-07
fix: paginate result set when listing installations

bcoe:
we were not paginating over all repos, so we would have only had a partial result set.

### DIFF
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -203,19 +203,22 @@ export class GCFBootstrapper {
         auth: await this.getInstallationToken(body.installation.id),
       });
       // Installations API documented here: https://developer.github.com/v3/apps/installations/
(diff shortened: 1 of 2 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -203,19 +203,22 @@ export class GCFBootstrapper {
         auth: await this.getInstallationToken(body.installation.id),
       });
       // Installations API documented here: https://developer.github.com/v3/apps/installations/
-      const {data} = await octokit.request('/installation/repositories', {
+      for await (const response of octokit.paginate.iterator({
```
JustinBeckwith, line 206: This is kinda hard to read.  Maybe perform the call to `octokit.paginate.iterator`, and then `for await` over that result instead of inlining it?

sofisl:
Glad you caught this!

PR #661, opened 2020-07-07
feat: Determines if issue applies to and is compliant with slo rules and handles OOSLO labeling

hershey127:


packages/slo-stat-bot/test/fixtures/config/valid-config.yml
```diff
@@ -0,0 +1 @@
+randomBoolean: true
```
sofisl, line 1: You can delete this file, it's just part of the generation of a probot in case you want to use a config file.

packages/slo-stat-bot/src/slo-logic.ts
```diff
@@ -0,0 +1,504 @@
...
+  title: string
+) {
+  if (!rule) {
+    return true;
+  }
+
+  rule = rule.toLowerCase();
+  const includes =
+    issueLabels.includes(rule) || issueLabels.includes(title + rule);
+  if (!includes) {
```
sofisl, line 182: I think you could just return includes. Same for other similar functions in this file.

packages/slo-stat-bot/src/slo-logic.ts
```diff
@@ -0,0 +1,504 @@
...
+  login: string;
+  permissions: {
+    pull: boolean;
+    push: boolean;
+    admin: boolean;
+  };
+}
+
+//Checking if slo applies to a given issue
+getSLOStatus.appliesTo = async function appliesTo(
```
sofisl, line 77: Super nitpicky, can we have a slightly more descriptive title for this function?

packages/slo-stat-bot/src/slo-logic.ts
```diff
@@ -0,0 +1,504 @@
...
+  slo: SLORules,
+  issueLabels: string[] | null
+): Promise<boolean> {
+  if (issueLabels === null || issueLabels.length === 0) {
+    return false;
```
sofisl, line 82: I think it would be good to make explicit various return paths, just to make sure you know what you're getting if you ever have to debug. You could add an else statement, or else you could make a placeholder variable whose default value is true. (same goes for some of the other similar functions below)

packages/slo-stat-bot/src/slo-logic.ts
```diff
@@ -0,0 +1,504 @@
...
+  }
+
+  //Checking if all the githublabels are subset of issue labels
+  const githubLabels = slo.appliesTo.gitHubLabels;
+  const isValidGithubLabels = await getSLOStatus.isValidGithubLabels(
+    issueLabels,
+    githubLabels
+  );
+  if (!isValidGithubLabels) {
```
sofisl, line 91: same, I think here you can just return isValidGithubLabels.

packages/slo-stat-bot/src/slo-logic.ts
```diff
@@ -0,0 +1,504 @@
...
+interface ReposListCollaboratorsItem {
+  login: string;
+  permissions: {
+    pull: boolean;
+    push: boolean;
+    admin: boolean;
+  };
+}
+
+//Checking if slo applies to a given issue
```
sofisl, line 76: These comments are awesome! I think it might be great to add comments like this to all functions to help guide the reader on what the functions are doing. It would also be helpful to have JSDOC-style annotation as well!

packages/slo-stat-bot/src/slo-logic.ts
```diff
@@ -0,0 +1,504 @@
...
+    owners = await getSLOStatus.convertToArray(owners);
+    for (const ownerPath of owners) {
+      const content = await getSLOStatus.getFilePathContent(
+        github,
+        owner,
+        repo,
+        ownerPath
+      );
+      const users = content.match(/@([^\s]+)/g);
+      if (users) {
```
sofisl, line 361: neat trick I learned for Node: you can just do users?.forEach to test whether it exists!

packages/slo-stat-bot/test/fixtures/events/valid_blob.json
```diff
@@ -0,0 +1,9 @@
+{
```
sofisl, line 1: This is more of an overall comment. You have a lot of these event files, and I'm not sure I see all of them in your test logic. I'd encourage you to delete some if you don't end up using them. Another point is that you don't actually have to return the 'exact' response for webhook events with github. For example, say you're requesting all the issues in a repository. But, you only ever use the number of the issue, not anything else. You can just specify an array of numbers if you don't ever do anything else with any other part of the response. LMK if you have any questions on this, I know this comment is pretty general.

packages/slo-stat-bot/src/slo-logic.ts
hershey127: I'm not sure if I can return the boolean value there. Because I'm checking all of the properties of the slo. SLO properties for checking if issue applies to contains: GitHub labels it must include, labels it must exclude, the priority, and the type of issue. If any of those properties is not satisfied, then the issue does not apply to the given slo.

packages/slo-stat-bot/src/slo-logic.ts
hershey127: That's true. I'll fix it!

packages/slo-stat-bot/test/fixtures/events/valid_blob.json
hershey127: That makes sense. It will make the code a lot more cleaner.

packages/slo-stat-bot/src/slo-label.ts
```diff
@@ -0,0 +1,131 @@
...
+interface SLOStatus {
+  appliesTo: boolean;
+  isCompliant: boolean | null;
+}
+
+/**
+ * Function gets ooslo label name in repo from the config file
+ * @returns the name of ooslo label
+ */
+handle_labeling.getLabelName = async function (): Promise<string> {
```
JustinBeckwith, line 29: Few things on this function:
- Use camelCase for the name
- This function doesn't need to be `async`!  Using `async` comes at the cost of instantiating a promise/closure, so it's not free.  Generally, you only want to use `async` for your function if you need to `await` something inside of it.

packages/slo-stat-bot/src/slo-label.ts
```diff
@@ -0,0 +1,131 @@
...
+      owner,
+      repo,
+      issue_number: issueNumber,
+      name,
+    });
+  } catch (err) {
+    console.error(
+      `Error removing OOSLO label in repo ${repo} for issue number ${issueNumber}\n ${err.request}`
+    );
+    return;
```
JustinBeckwith, line 95: The return here isn't really needed.

packages/slo-stat-bot/src/slo-label.ts
```diff
@@ -0,0 +1,131 @@
...
+    const labels = [name];
+    await github.issues.addLabels({
+      owner,
+      repo,
+      issue_number: issueNumber,
+      labels,
+    });
+  } catch (err) {
+    //Error if ooslo label does not exist in repo
+    throw `Error in adding ooslo label for org ${owner} in repo ${repo} since it does not exist \n ${err}`;
```
JustinBeckwith, line 64: In cases where you catch an exception, and rethrow, it's important to preserve the context of the original exception.  The original exception will have a full call stack, original error, and all the details that explain what *really* happened here.  I suggest a pattern like this:

```js
} catch (err) {
  err.message = `Error in adding ooslo label for org ${owner} in repo ${repo} since it does not exist \n ${err.message} \n ${err}`;
  throw err;
}
```

packages/slo-stat-bot/src/slo-lint.ts
```diff
@@ -0,0 +1,277 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {GitHubAPI} from 'probot/lib/github';
+import Ajv, {ErrorObject} from 'ajv';
+import {Context} from 'probot';
```
JustinBeckwith, line 17: You can delete line 15 and export `GitHubAPI` directly from Probot :)

packages/slo-stat-bot/src/slo-lint.ts
```diff
@@ -0,0 +1,277 @@
...
+  try {
+    const listOfFiles = await github.pulls.listFiles({
+      owner,
+      repo,
+      pull_number,
+      per_page,
+    });
+    return listOfFiles.data;
+  } catch (err) {
+    console.warn(
```
JustinBeckwith, line 66: When do we use `console.error` vs `console.warn` vs `console.log`?  Maybe a better question for @bcoe :)

packages/slo-stat-bot/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "test:snap": "SNAPSHOT_UPDATE=1 npm test",
+    "fix": "gts fix",
+    "lint": "gts check"
+  },
+  "dependencies": {
+    "ajv": "^6.12.2",
+    "gcf-utils": "5.0.0",
+    "moment": "^2.27.0",
+    "probot": "^9.11.3",
+    "ts-auto-mock": "^2.3.2"
```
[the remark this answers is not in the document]
hershey127: Sorry I was gonna use it to mock the github api but decided against that. Will remove!

packages/slo-stat-bot/src/slo-bot.ts
```diff
@@ -0,0 +1,212 @@
...
+      'pull_request.edited',
+      'pull_request.synchronize',
+    ],
+    async (context: Context) => {
+      const owner = context.payload.repository.owner.login;
+      const repo = context.payload.repository.name;
+      const pullNumber = context.payload.number;
+      const labelsResponse = context.payload.pull_request.labels;
+
+      await handle_lint(context, owner, repo, pullNumber);
```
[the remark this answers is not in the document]
hershey127: My bad will fix!

packages/slo-stat-bot/src/slo-label.ts
hershey127: Thanks for the tip. I will keep that in mind when creating function headers.

packages/slo-stat-bot/src/slo-bot.ts
```diff
@@ -0,0 +1,212 @@
...
+    );
+  }
+  if (sloRules === 'not found') {
+    //Error if org level does not exist
+    throw `Error in finding org level config file in ${owner}`;
+  }
+  return sloRules;
+};
+
+export = handler;
```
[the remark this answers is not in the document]
hershey127: That makes sense. I'll go ahead and change the way I export in all of the file!

packages/slo-stat-bot/README.md
```diff
@@ -1,11 +1,19 @@
...
+# SLO Bot
 
 Lints for slo rules (issue_slo_rules.json) file and creates a check on the PR. Comments on the PR if it is invalid.
 
 (https://github.com/googleapis/repo-automation-bots/blob/master/README.md) for deploying and testing your bots.
 
 This bot uses nock for mocking requests to GitHub, and snap-shot-it for capturing responses; This allows updates to the API surface to be treated as a visual diff, rather than tediously asserting against each field.
 
+## Usage:
+Create Out of SLO label in the repo. Update `label_name.yml` with the name of OOSLO label.
```
orthros, line 10: nit: we usually have the yaml file name be the name of the bot + .yaml

packages/slo-stat-bot/src/slo-label.ts
```diff
@@ -0,0 +1,133 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// eslint-disable-next-line node/no-extraneous-import
+import {GitHubAPI, Context} from 'probot';
+
+const CONFIGURATION_FILE_PATH = 'label_name.yml';
```
orthros, line 18: Same as above =>  slo-stat-bot

packages/slo-stat-bot/src/slo-label.ts
```diff
@@ -0,0 +1,133 @@
...
+
+interface SLOStatus {
+  appliesTo: boolean;
+  isCompliant: boolean | null;
+}
+
+interface IssueLabelResponseItem {
```
orthros, line 25: nit: rename to Config

packages/slo-stat-bot/src/slo-label.ts
```diff
@@ -0,0 +1,133 @@
...
+  try {
+    const labelName = (await context.config(
+      CONFIGURATION_FILE_PATH
+    )) as IssueLabelResponseItem;
+    return labelName.name;
+  } catch (err) {
+    err.message = `Unable to get ooslo name from config-label file \n ${err.message}`;
+    throw err;
+  }
+};
```
orthros, line 43: Should have a default config for when a the configuration file does not exist

packages/slo-stat-bot/src/slo-logic.ts
hershey127: I can add that in getSloStatus if the issue does not apply to slo

packages/slo-stat-bot/src/slo-appliesTo.ts
```diff
@@ -0,0 +1,210 @@
...
+  return false;
+};
+
+/**
+ * Function checks if all the githublabels are subset of issue labels
+ * @param issueLabels of the issue
+ * @param githubLabels is slo rule for github labels that must exist in issue
+ * @returns true if githubLabels applies to issues else false
+ */
+export const isValidGithubLabels = async function isValidGithubLabels(
```
sofisl, line 48: these seem like very small functions. Is it possible to make them inline in your code? Might make it easier to read! It sounds like you could just do something like 
```if(!githubLabels) {
//code I want to execute
}
```
instead of returning true to then cause something else to happen. Just a thought!

sofisl: I think the same goes for ```isValidExcludedLabels```, ```isElementExist```, etc.

packages/slo-stat-bot/src/slo-appliesTo.ts
```diff
@@ -0,0 +1,210 @@
...
+  issueLabels: string[] | null,
+  number: number
+): Promise<boolean> {
+  const sloString = JSON.stringify(slo, null, 4);
+
+  if (Object.keys(slo.appliesTo).length === 0) {
+    return true;
+  }
+
+  if (issueLabels === null) {
```
sofisl, line 147: Might be worth checking for falsy vs. null, I'm never quite sure what you get with these methods.

packages/slo-stat-bot/src/slo-appliesTo.ts
```diff
@@ -0,0 +1,210 @@
...
+    return false;
+  }
+
+  const excludedGitHubLabels = slo.appliesTo.excludedGitHubLabels;
+  const hasNoExLabels = await isValidExcludedLabels(
+    issueLabels,
+    excludedGitHubLabels
+  );
+  if (!hasNoExLabels) {
+    console.info(
```
sofisl, line 185: meta note: I think I saw @JustinBeckwith recommend we set all our console.logs and console.infos to use the logger from gcf-utils in the GH Automation Chat today. Not sure necessarily where that applies, but just FYI.

packages/slo-stat-bot/src/slo-compliant.ts
```diff
@@ -0,0 +1,377 @@
...
+      repo,
+      path,
+    });
+    const data = fileResponse.data as {content?: string};
+    const content = Buffer.from(data.content as string, 'base64').toString(
+      'utf8'
+    );
+    return content;
+  } catch (err) {
+    return 'not found';
```
sofisl, line 60: I think we want to start throwing errors if we don't recognize them... see #722. See if you can catch the normal errors you'd get, and maybe let's use the body of the error message to surface a usable error. For other errors, we can just let it throw if that goes along with the execution of the bot (i.e., if you ultimately want the bot to fail there anyways).

packages/slo-stat-bot/src/slo-appliesTo.ts
hershey127: I could just do check for `(!issueLabels)` which will consider both the case if it is null or falsy.

packages/slo-stat-bot/src/slo-appliesTo.ts
hershey127: I was gonna try switching the logging and see if the app is still functional

packages/slo-stat-bot/src/slo-appliesTo.ts
hershey127: That is a fair point. Initially I wanted to make every function isolated so it would be easier to understand the decision tree and the reasoning for a specific outcome. But it does result in extra code and space. So I can try condensing it down.

hershey127: I also wanted to make determining if it is `isValidGithubLabels`, `isValidExcludedLabels`, `isElementExist` more faster by making those functions asynchronous. If the slo does not apply to any one of the given functions it would immediately return false instead of waiting for the other lines to execute. But considering these are much simpler functions I don't think that should be a problem since it should compute fast.

PR #667, opened 2020-07-08
feat(buildcop): use tested config type

tbpg:
I'm not 100% sure of this design. So, I'm open to other ideas. Perhaps we should put the long error message in a custom error type?

I thought about having `func newConfig`, but that had the same issue of having too many arguments and being confusing.

We could also use something like Viper, but it doesn't seem necessary.

### DIFF
--- a/packages/buildcop/buildcop.go
+++ b/packages/buildcop/buildcop.go
@@ -97,7 +97,6 @@ type config struct {
 	commit         string
 	logsDir        string
 	serviceAccount string
-	buildURL       string
 }
 
 func (cfg *config) setDefaults() (ok bool) {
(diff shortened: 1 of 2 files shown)


packages/buildcop/buildcop.go
```diff
@@ -191,9 +228,9 @@ func processLog(ctx context.Context, repo, installationID, commit string, topic
 			Name:         "buildcop",
 			Type:         "function",
 			Location:     "us-central1",
-			Installation: githubInstallation{ID: installationID},
-			Repo:         repo,
-			Commit:       commit,
+			Installation: githubInstallation{ID: cfg.installationID},
+			Repo:         cfg.repo,
+			Commit:       cfg.commit,
 			BuildURL:     buildURL,
```
codyoss, line 234: cfg.buildURL?

tbpg: It's not plumbed through the `config` type yet. I removed it from the struct.

PR #670, opened 2020-07-09
feat(gcf-utils): Log bot execution trigger information from gcf-utils

azizsonawalla:
Log incoming trigger information for bots

Fixes https://github.com/googleapis/repo-automation-bots/issues/675


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -146,35 +167,88 @@ export class GCFBootstrapper {
     return config as Options;
   }
 
+  parseRequestHeaders(
```
azizsonawalla, line 170: This logic already existed before - I just pulled it out into a separate function

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -146,35 +167,88 @@ export class GCFBootstrapper {
...
+      request.get('X-GitHub-Delivery') ||
+      '';
+    const taskId =
+      request.get('X-CloudTasks-TaskName') ||
+      request.get('x-cloudtasks-taskname') ||
+      '';
+    return {name, id, signature, taskId};
+  }
+
+  parseTriggerType(name: string, taskId: string): TriggerType {
```
azizsonawalla, line 192: Again, existing logic that I just pulled into a separate function and added `TriggerType`s

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -101,6 +115,13 @@ export interface CronPayload {
   cron_org: string;
 }
 
+export enum TriggerType {
```
azizsonawalla, line 118: I'm only exporting this for testing purposes. I could not export it and then use either `rewire` or just strings to replicate these values but that seems more fragile. Is this acceptable the way it is? Any other ideas?

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -146,35 +167,88 @@ export class GCFBootstrapper {
...
+    } else if (name) {
+      return TriggerType.TASK;
+    }
+    return TriggerType.UNKNOWN;
+  }
+
+  buildTriggerInfo(
+    triggerType: TriggerType,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    requestBody: {[key: string]: any}
```
azizsonawalla, line 209: Need to have `any` here because the content of the payload depends on GitHub so can't really type-check it

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -146,35 +167,88 @@ export class GCFBootstrapper {
...
     appFn: ApplicationFunction
   ): (request: express.Request, response: express.Response) => Promise<void> {
     return async (request: express.Request, response: express.Response) => {
       // Otherwise let's listen handle the payload
       this.probot = this.probot || (await this.loadProbot(appFn));
+      const {name, id, signature, taskId} = this.parseRequestHeaders(request);
+      const triggerType: TriggerType = this.parseTriggerType(name, taskId);
+
+      if (triggerType !== TriggerType.TASK) {
+        // we don't want to log TASK triggers since we enqueue the task ourselves
```
chingor13, line 247: I *think* tasks can be requeued on failure, so we might want to see trigger info on the task types?

Without logging the task executions, it still won't be easy to easily find the real logs we are looking for.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -146,35 +167,88 @@ export class GCFBootstrapper {
...
+      return TriggerType.TASK;
+    }
+    return TriggerType.UNKNOWN;
+  }
+
+  buildTriggerInfo(
+    triggerType: TriggerType,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    requestBody: {[key: string]: any}
+  ) {
```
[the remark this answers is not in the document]
azizsonawalla: Yep, I'll add that

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: Ideally when we enqueue a task after getting the webhook from Github, we would also pass on the execution id of the function that received the webhook so we could have a mapping between them. But yeah, your point makes sense. For now I'll just log all triggers as-is.

azizsonawalla: Looks like it's not possible to access the `execution_id` from the function itself: https://groups.google.com/g/firebase-talk/c/5x4MtKxSdNQ

chingor13: Can we associate the events via the [GitHub delivery GUID](https://developer.github.com/webhooks/event-payloads/#delivery-headers)?

azizsonawalla: Yeah I was going to say we can create our own GUID but this works too

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -64,6 +64,7 @@ export interface GCFLogger {
   warn: LogFn;
   error: LogFn;
   metric: LogFn;
+  flushSync: Function;
```
chingor13, line 67: Can we explicitly declare the param signature here?

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -146,35 +180,97 @@ export class GCFBootstrapper {
...
+    const id =
+      request.get('x-github-delivery') ||
+      request.get('X-GitHub-Delivery') ||
+      '';
+    // TODO: add test to validate this signature is used on the initial
+    // webhook from GitHub:
+    const signature =
+      request.get('x-github-delivery') ||
+      request.get('X-GitHub-Delivery') ||
+      '';
```
chingor13, line 197: Strange that these are both reading the same value, although it was in the previous version too.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -146,35 +180,97 @@ export class GCFBootstrapper {
...
+      return TriggerType.SCHEDULER;
+    } else if (!taskId && name) {
+      return TriggerType.GITHUB;
+    } else if (name) {
+      return TriggerType.TASK;
+    }
+    return TriggerType.UNKNOWN;
+  }
+
+  buildTriggerInfo(
```
chingor13, line 219: This looks like new public interface - can we add jsdocs?

Also, this can probably be static, private, or live outside the class completely.

packages/gcf-utils/src/gcf-utils.ts
chingor13: This can probably be static, private, or live outside the class completely.

packages/gcf-utils/src/gcf-utils.ts
chingor13: This can probably be static, private, or live outside the class completely.

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: sure thing, I'll add that

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: Yeah I'm not sure if it was intentional or not, but I left it as-is

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: sure, I'll move it

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: I can move it to be outside the class/static. Also, this is public just for testing purposes. I can make it private and then access it in tests using JavaScript props. Either way, happy to add JSDocs

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: I've made them static + private but kept them in the class since they're only used within it

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -77,13 +78,26 @@ export function initLogger(
...
+  const logger = pino(defaultOptions, dest);
   Object.keys(logger).map(prop => {
     if (logger[prop] instanceof Function) {
       logger[prop] = logger[prop].bind(logger);
     }
   });
-  return {...logger, metric: logger.metric.bind(logger)};
+
+  const flushSync = () => {
+    if (dest instanceof SonicBoom) {
```
chingor13, line 91: Let's add a comment about why we only need to flush SonicBoom destinations.

azizsonawalla:
> This looks really solid, I like that you actually refactored and cleaned up some of the messy code in `gcf-utils` 👍

always leave it better than you found it :)

PR #671, opened 2020-07-09
fix(buildcop): add extra newline to prevent title formatting

tbpg:
See https://github.com/googleapis/google-cloud-go/issues/2571#issuecomment-656331671 for an example of the accidental title formatting.


tbpg:
I added the `cla: yes` label because the CLA bot is currently running behind and I am the only author (and I work at Google :smiley:).

PR #674, opened 2020-07-10
fix: page iterator in auto-label needs to iterate over issues also

sofisl:
Auto-label is erroring out because we need to iterate over issues in each page. Also, changed the logic so that auto-label checks if labels were _not_ added to each issue when it runs on a scheduler job, and will exit the loop if 5 issues in each repository did not have labels added to them, so that auto-label doesn't  waste checking all the labels for that repo.


packages/auto-label/src/auto-label.ts
```diff
@@ -125,30 +125,6 @@ handler.checkExistingIssueLabels = async function checkExistingIssueLabels(
...
-      console.log(`${owner}/${repo}#${issueNumber} already has an api: label`);
-      return undefined;
-    }
-    await handler.addLabels(github, owner, repo, issueNumber, [label]);
-    return label;
-  }
-  await handler.addLabels(github, owner, repo, issueNumber, [label]);
-  return label;
-};
-
```
sofisl, line 151: removed this logic because it essentially replicates what the main logic does: checks to see if the label already exists, and then either does or doesn't add it. I changed the logic below to incorporate the possibility of an auto-detected label. I also added a clean-up function, which will remove any tags that are not either the auto-detected one or assigned by the file.

packages/auto-label/src/auto-label.ts
```diff
@@ -236,9 +212,7 @@ handler.addLabeltoRepoAndIssue = async function addLabeltoRepoAndIssue(
     repo
   );
 
-  let autoDetectedLabelAdded: string | undefined;
-
-  let wasAdded = false;
+  let wasNotAdded = true;
```
JustinBeckwith, line 215: I generally avoid negatives in variable names, especially with booleans.  It's possible this is fine, just saying it makes things harder to read.

packages/auto-label/src/auto-label.ts
```diff
@@ -247,94 +221,78 @@ handler.addLabeltoRepoAndIssue = async function addLabeltoRepoAndIssue(
...
+  }
+  const colorNumber =
+    jsonArray?.findIndex((object: JSONData) => objectInJsonArray === object) %
+      colorsData.length || 0;
+
+  const githubLabel = objectInJsonArray?.github_label || autoDetectedLabel;
+
+  let alreadyExists;
+  if (githubLabel) {
+    alreadyExists = await handler.checkExistingLabels(
```
JustinBeckwith, line 240: If I'm reading this right, it looks like you make an API call to check if a given issue has a label, and if it doesn't, you add it.  Instead - why not just always try to add the label?  As long as it has the same text/title, the GitHub API (I think) should figure it out, and it would reduce the complexity and number of API calls.

packages/auto-label/src/auto-label.ts
```diff
@@ -247,94 +221,78 @@ handler.addLabeltoRepoAndIssue = async function addLabeltoRepoAndIssue(
...
+        console.log('This label already exists on this issue');
+        return;
+      } else {
+        await handler.addLabels(github, owner, repo, issueNumber, [
+          githubLabel,
+        ]);
+        wasNotAdded = false;
+      }
+      for (const dirtyLabel of cleanUpOtherLabels) {
+        await github.issues
```
JustinBeckwith, line 280: I really like this style of calling the API where you don't rely on wrapper functions.

packages/auto-label/src/auto-label.ts
sofisl: Yeah, I agree. However, it's ultimately what we want to count for, since we want to exit out of checking that individual repo if the label was not added x amount of times.

packages/auto-label/src/auto-label.ts
sofisl: good call!

PR #686, opened 2020-07-14
feat(gcf-util): add a plugin for Octokit to automatically log bot actions

azizsonawalla:
A plugin to hook into request cycle for outgoing GitHub API calls in order to log metrics

Currently only wraps Octokit endpoints that:
1. Are being used by our bots currently
2. AND endpoints that make a change on GitHub (so endpoints that just get information are ignored)

Fixes https://github.com/googleapis/repo-automation-bots/issues/687


packages/gcf-utils/package.json
```diff
@@ -42,6 +42,7 @@
     "@types/tmp": "^0.2.0",
     "@types/yargs": "^15.0.0",
     "@types/pino": "^6.3.0",
+    "@octokit/rest": "^16.43.1",
```
azizsonawalla, line 45: This is not the latest version but it's the version that Probot uses

azizsonawalla: I could also completely remove this dependency from here since Probot pulls it in anyway (and that way we always have the same version as Probot) but the lint gives an 'extraneous dependency' error

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -17,9 +17,10 @@ import {CloudTasksClient} from '@google-cloud/tasks';
 import {v1} from '@google-cloud/secret-manager';
 import * as express from 'express';
 import pino from 'pino';
-// eslint-disable-next-line node/no-extraneous-import
 import {Octokit} from '@octokit/rest';
 import SonicBoom from 'sonic-boom';
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const LoggingOctokitPlugin = require('../src/logging-octokit-plugin.js');
```
azizsonawalla, line 23: Had to disable the `no-var-requires` here because when I changed the `const LoggingOctokitPlugin` to `import LoggingOctokitPlugin` Octokit wouldn't accept it (gave a type mismatch).

TomKristie: Have you tried `import * as LoggingOctokitPlugin`

packages/gcf-utils/src/logging-octokit-plugin.ts
```diff
@@ -0,0 +1,245 @@
...
+ * @param details action details
+ */
+function logGithubAction(logger: MetricLogger, details: GitHubActionDetails) {
+  if (!details.type || details.type === GitHubActionType.UNKNOWN) {
+    // don't log unknown action types
+    return;
+  }
+  const general = {
+    type: details.type,
+    value: details.value || 'UNKNOWN',
```
TomKristie, line 204: Minor: it might be easier using a `const` for these in case you have typos in the future

packages/gcf-utils/src/logging-octokit-plugin.ts
```diff
@@ -0,0 +1,245 @@
...
+) => {
+  const octoLogger = pluginOptions.customLogger || logger;
+
+  octokit.hook.before('request', async options => {
+    const actionDetails: GitHubActionDetails = parseActionDetails(options);
+
+    logGithubAction(octoLogger, actionDetails);
+  });
+
+  (octokit as Octokit & {
```
TomKristie, line 242: Are you changing the type and properties of an octokit instance? Will this have a side-effect on users that assume their octokit's properties aren't modified?

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: Yeah I already tried that and the message I get is 
```
Argument of type 'typeof import("/home/asonawalla/github-repos/repo-automation-bots/packages/gcf-utils/src/logging-octokit-plugin")' is not assignable to parameter of type 'Plugin | Plugin[]'.
```
I think the issue is that the plugin is a `.js` file (as required by Octokit) and so types are messy/non-existent

packages/gcf-utils/src/logging-octokit-plugin.ts
azizsonawalla: The cast is only local to that statement so I'm not changing the type returned to the bots. While this does add new properties to the octokit instance (as does the rest of the plugin) it maintains backward compatibility

packages/gcf-utils/src/logging-octokit-plugin.ts
```diff
@@ -0,0 +1,247 @@
...
+ * Hooks into outgoing requests from Octokit to log metrics
+ */
+module.exports = (
+  octokit: Octokit,
+  pluginOptions: {customLogger?: MetricLogger}
+) => {
+  const octoLogger = pluginOptions.customLogger || logger;
+
+  octokit.hook.before('request', async options => {
+    const actionDetails: GitHubActionDetails = parseActionDetails(options);
```
bcoe, line 239: I like this approach, we ultimately end up logging at a slightly lower level than you were intending if I follow? and intercept all requests.

packages/gcf-utils/test/logging-octokit-plugin.ts
```diff
@@ -0,0 +1,373 @@
...
+  it('logs information for issues.createLabel', () => {
+    loggingOctokit.issues
+      .createLabel({
+        owner: 'fooOwner',
+        repo: 'barRepo',
+        name: 'labelName',
+        color: 'blue',
+      })
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      .catch(e => {
```
bcoe, line 115: feel free to omit the catch if we're not using it (_I'm guessing this was used during debugging?_)

packages/gcf-utils/test/logging-octokit-plugin.ts
```diff
@@ -0,0 +1,373 @@
...
+              object_id: 3,
+            },
+          },
+        };
+        const actual = logger.lastLogData;
+        assert.deepEqual(actual, expected);
+      });
+  });
+
+  it('does not log information for unknown actions', () => {
```
bcoe, line 360: I appreciate how thoroughly this is tested 👍

packages/gcf-utils/src/logging-octokit-plugin.ts
azizsonawalla: Yep - there was an oversight in my original design: the Octokit plugin infrastructure is built on JavaScript and I didn't realize how much of an issue types will become. It got to the point where I decided that manipulating any types here will just lead to brittle code, so I changed the approach

packages/gcf-utils/test/logging-octokit-plugin.ts
azizsonawalla: The catch here is for catching and ignoring the failed call to Octokit (it fails because the Octokit is not authenticated). If I omit it then it throws errors in the test (uncaught rejected promise) and it will fail the test.

packages/gcf-utils/test/integration/gcf-bootstrapper-integration.ts
```diff
@@ -55,5 +56,23 @@ describe('GCFBootstrapper Integration', () => {
...
+    it('provides github with logging plugin', async () => {
+      const pb = await bootstrapper.loadProbot((app: Application) => {
+        app.on('foo', async context => {
+          assert(
+            (context.github as GitHubAPI & {
+              loggingOctokitPluginVersion: string;
+            }).loggingOctokitPluginVersion === '1.0.0'
+          );
+        });
+      });
```
chingor13, line 69: We probably want to assert that this block is actually hit.

packages/gcf-utils/src/logging-octokit-plugin.ts
```diff
@@ -0,0 +1,245 @@
...
+    case GitHubActionType.ISSUE_UPDATE:
+      return 'updated: ' + hasIssueUpdateProps.join(',');
+    case GitHubActionType.ISSUE_CREATE:
+      return String(options.title);
+    case GitHubActionType.PR_DISMISS_REVIIEW:
+      return `dismiss ${options.review_id}: ${options.message}`;
+    case GitHubActionType.PR_MERGE:
+    case GitHubActionType.PR_UPDATE_BRANCH:
+    default:
+      return 'NONE';
```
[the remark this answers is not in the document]
azizsonawalla: You mean the `NONE` value? The switch here is returning the `value` property based on `GitHubActionType` so `NONE` here refers to the `value` rather than the type. I already have an `UNKNOWN` item in `GitHubActionType. Does that answer your question or am I misunderstanding?

packages/gcf-utils/test/integration/gcf-bootstrapper-integration.ts
azizsonawalla: Yes I'm actually still working on getting the integration tests to run -- running into some more access issues. Will add the assert once I can get them running properly.

azizsonawalla:
I'm having some access issues running the integration test so best to hold off on merging until I can run them to make sure that the 1 integration test passes. Unit tests are all fine.

packages/gcf-utils/test/integration/gcf-bootstrapper-integration.ts
```diff
@@ -14,9 +14,32 @@
...
 import {GCFBootstrapper} from '../../src/gcf-utils';
 import {describe, beforeEach, afterEach, it} from 'mocha';
-import {Application} from 'probot';
+import {Application, GitHubAPI} from 'probot';
 import {resolve} from 'path';
 import {config} from 'dotenv';
+import assert from 'assert';
+
+/**
+ * How to run these tests:
```
azizsonawalla, line 23: @bcoe @chingor13 @orthros I was able to get these tests to finally run following these instructions - I added them in here in case someone else runs into the same issue, but I don't know if this is the 'official'/'correct' way of doing this. Let me know if this sounds okay!

packages/gcf-utils/src/logging-octokit-plugin.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  octokit.hook.before('request', async options => {
+    const actionDetails: GitHubActionDetails = parseActionDetails(options);
+
+    logGithubAction(octoLogger, actionDetails);
+  });
+
+  (octokit as Octokit & {
+    loggingOctokitPluginVersion: string;
+  }).loggingOctokitPluginVersion = '1.0.0';
```
chingor13, line 246: nit: make this a constant and put at the top of the file?

packages/gcf-utils/src/logging-octokit-plugin.ts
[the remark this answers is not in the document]
chingor13: It just seems a bit odd to have every other option return a `GitHubActionType` enum value (which is a string) yet `NONE` is just a string. It's possible to make this method return `GitHubActionType` if we want to (and if it makes sense).

packages/gcf-utils/test/integration/gcf-bootstrapper-integration.ts
chingor13: This seems reasonable to me. Many integration tests require setup in a test project (since most won't have direct access to the production project resources).

packages/gcf-utils/test/logging-octokit-plugin.ts
chingor13: Should we rethrow if it's not a rejected promise? If something else changes that throws and exception, this test may not catch it.

Issue #699, opened 2020-07-16
flaky-tests: write test for skipping background tasks


bcoe:
you can now configure gcf-utils such that it skips background tasks. We should add a test for this.

PR #700, opened 2020-07-16
feat(gcf-utils): some bots have large payloads not appropriate for cloud task payload 

bcoe:
We should add a test for this, but I thought it would be good to get a hot fix in for build cop quickly.

This also unblocks @azizsonawalla's work  as well, which requires that bots have `@types/sonic-boom` available.

Refs: https://github.com/googleapis/repo-automation-bots/issues/699 (tracking issue for test).

### DIFF
--- a/packages/gcf-utils/package.json
+++ b/packages/gcf-utils/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gcf-utils",
-  "version": "5.1.0",
+  "version": "5.2.2",
   "description": "An extension for running Probot in Google Cloud Functions",
   "scripts": {
     "compile": "tsc -p .",
@@ -27,6 +27,7 @@
     "@types/express": "^4.17.0",
     "@types/ioredis": "^4.14.8",
     "@types/lru-cache": "^5.1.0",
+    "@types/sonic-boom": "^0.7.0",
     "@octokit/plugin-enterprise-compatibility": "1.2.5",
     "express": "^4.17.1",
     "gaxios": "^3.0.0",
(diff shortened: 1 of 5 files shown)


packages/gcf-utils/package.json
```diff
@@ -27,6 +27,7 @@
     "@types/express": "^4.17.0",
     "@types/ioredis": "^4.14.8",
     "@types/lru-cache": "^5.1.0",
+    "@types/sonic-boom": "^0.7.0",
```
bcoe, line 30: gcf-utils had problems on the main branch do to exporting `@types/sonic-boom`, but only having it as dev dep.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -283,8 +287,13 @@ export class GCFBootstrapper {
   }
 
   gcf(
-    appFn: ApplicationFunction
+    appFn: ApplicationFunction,
+    wrapOptions?: WrapOptions
   ): (request: express.Request, response: express.Response) => Promise<void> {
+    let background = true;
+    if (wrapOptions) {
```
JustinBeckwith, line 294: 🐛 - If I pass WrapOptions here as `{}`, or maybe with a second property down the road, it will set the default of wrapOptions to false, while it looks like you want the default to be true.

Instead I'd suggest:
- changing line 291 to be `wrapOptions: WrapOptions = {}` 
- use extend to do:

```
const defaults = { wrapOptions: true };
const options = extend(true, defaults, wrapOptions, {});
```

bcoe: The interface requires that `WrapOptions`, if provided, has a boolean `background` property,  I don't understand how you're passing `{}`, and if you did the side effect would just be that `background` is falsy.

bcoe: so, the `if (wrapOptions) {` is just checking whether or not `wrapOptions` is `undefined`.

bcoe: updated the code to be less confusing.

packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -59,7 +64,7 @@ describe('GCFBootstrapper', () => {
 
     let enqueueTask: sinon.SinonStub;
 
-    beforeEach(async () => {
+    async function mockBootstrapper(wrapOpts?: WrapOptions) {
```
tbpg, line 67: Why the change from `beforeEach`?

PR #702, opened 2020-07-17
feat: auto label adds sample tag, and fixes broken tests

### LINKED ISSUE #701: Auto-label tests are broken
I was running the tests for `auto-label` today, and noticed some sketchy output:

```
deletes extraneous labels
[@octokit/rest] `const Octokit = require("@octokit/rest")` is deprecated. Use `const { Octokit } = require("@octokit/rest")` instead
running for org testOwner
RequestError [HttpError]: request to https://api.github.com/repos/testOwner/testRepo/labels failed, reason: Nock: No match for request {
  "method": "POST",
  "url": "https://api.github.com/repos/testOwner/testRepo/labels",
  "headers": {
    "accept": [
      "application/vnd.github.v3+json"
    ],
    "user-agent": [
      "octokit.js/16.43.2 Node.js/12.18.0 (macOS Catalina; x64)"
    ],
    "authorization": [
      "token abc123"
    ],
    "content-type": [
      "application/json; charset=utf-8"
    ],
    "content-length": [
      "41"
    ],
    "accept-encoding": [
      "gzip,deflate"
    ],
    "connection": [
      "close"
    ]
  },
  "body": "{\"name\":\"myGitHubLabel\",\"color\":\"FEFEFA\"}"
}
    at /Users/beckwith/Code/repo-automation-bots/packages/auto-label/node_modules/@octokit/request/dist-node/index.js:107:11
        -> /Users/beckwith/Code/repo-automation-bots/packages/auto-label/node_modules/@octokit/request/dist-src/fetch-wrapper.js:88:15
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Function.createLabel (/Users/beckwith/Code/repo-automation-bots/packages/auto-label/build/src/auto-label.js:54:22)
        -> /Users/beckwith/Code/repo-automation-bots/packages/auto-label/src/auto-label.ts:85:18 {
  status: 500,
  headers: {},
  request: {
    method: 'POST',
    url: 'https://api.github.com/repos/testOwner/testRepo/labels',
    headers: {
      accept: 'application/vnd.github.v3+json',
      'user-agent': 'octokit.js/16.43.2 Node.js/12.18.0 (macOS Catalina; x64)',
      authorization: 'token [REDACTED]',
      'content-type': 'application/json; charset=utf-8'
    },
    body: '{"name":"myGitHubLabel","color":"FEFEFA"}',
    request: { hook: [Function: bound bound register], validate: [Object] }
  }
}
```

Despite that console output, the tests pass!?  Turns out it's because of the way we do `try` wrapping around many of the functions, preventing them from throwing.  As a result, `nock` can't interject to fail a given set of tests.   This effectively means our tests may not be working.

sofisl:
Fixes #701 and #617


packages/auto-label/src/auto-label.ts
```diff
@@ -200,21 +200,25 @@ handler.addLabeltoRepoAndIssue = async function addLabeltoRepoAndIssue(
   jsonArray: JSONData[],
   github: GitHubAPI
 ) {
+
```
JustinBeckwith, line 203: Looks like we need a `npm run fix` :)

packages/auto-label/src/auto-label.ts
```diff
@@ -225,36 +229,42 @@ handler.addLabeltoRepoAndIssue = async function addLabeltoRepoAndIssue(
   }
   const colorNumber =
     jsonArray?.findIndex((object: JSONData) => objectInJsonArray === object) %
```
JustinBeckwith, line 231: This statement is very hard to read.  Multiple split into a few lines so it's easier to follow?

packages/auto-label/src/auto-label.ts
```diff
@@ -269,10 +279,32 @@ handler.addLabeltoRepoAndIssue = async function addLabeltoRepoAndIssue(
       }
     } else {
       await handler.addLabels(github, owner, repo, issueNumber, [githubLabel]);
+      console.log(`Label added to ${owner}/${repo} for issue ${issueNumber} is ${githubLabel}`);
       wasNotAdded = false;
     }
   }
 
+  let foundSamplesTag = undefined;
```
JustinBeckwith, line 287: Generally you don't need to assign `undefined`, cause that's what you get by default! What's the type of `foundSamplesTag` supposed to be here?

packages/auto-label/src/auto-label.ts
```diff
@@ -269,10 +279,32 @@ handler.addLabeltoRepoAndIssue = async function addLabeltoRepoAndIssue(
...
+  if (labelsOnIssue) {
+    foundSamplesTag = labelsOnIssue.find(
+      (element: {name: string}) => element.name === 'sample'
```
JustinBeckwith, line 290: The fact that you need to add the type of `element` makes me think `labelsOnIssue` is an `any` or `any[]`.  Can we put a type on `labelsOnIssue` so you don't need to do the type down here?

packages/auto-label/src/auto-label.ts
```diff
@@ -269,10 +279,32 @@ handler.addLabeltoRepoAndIssue = async function addLabeltoRepoAndIssue(
...
+    );
+  }
+  if (!foundSamplesTag && repo.includes('sample')) {
+    await handler.createLabel(
```
JustinBeckwith, line 294: Thank you :)

packages/auto-label/test/auto-label.ts
```diff
@@ -339,6 +419,56 @@ describe('auto-label', () => {
       });
       ghRequests.done();
     });
+
+    it('will add a samples tag for a samples repo', async () => {
+      const ghRequests = nock('https://api.github.com')
```
JustinBeckwith, line 424: Gosh these nock chains are painful. I don't have a better answer, but we should put our heads together with @bcoe to maybe consider a different testing strategy (after this PR lands of course)

PR #703, opened 2020-07-17
feat(monitoring-system): create a REST service for the Data Processor

azizsonawalla:
A REST service for Data Processor to be deployed to Cloud Run.

Note: This application is not complete - this is just a first pull request with some initial code (didn't want to create a single pull request with the entire implementation)


packages/monitoring-system/data-processor/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "bugs": "https://github.com/googleapis/repo-automation-bots/issues",
+  "main": "./build/src/task-service.js",
+  "author": "asonawalla@google.com",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "express": "^4.17.1",
+    "firebase-admin": "^9.0.0"
+  },
+  "devDependencies": {
+    "@types/chai": "^4.2.11",
```
JustinBeckwith, line 27: Let's use the built in `assert` module please :)

JustinBeckwith: For the HTTP stuff, supertest is what we generally use here

packages/monitoring-system/data-processor/src/data-processor-factory.ts
```diff
@@ -0,0 +1,46 @@
...
+      case Task.ProcessLogs:
+        return new CloudLogsProcessor();
+      case Task.ProcessGCF:
+        return new GCFProcessor();
+      case Task.ProcessTaskQueue:
+        return new CloudTasksProcessor();
+      case Task.ProcessGitHub:
+        return new GitHubProcessor();
+      default:
+        throw new Error(`Couldn't identify a data processor for task: ${task}`);
```
TomKristie, line 43: If I am following correctly this error message is never logged. Is there benefit in not logging an error?

packages/monitoring-system/data-processor/package.json
azizsonawalla: I believe the chai agent extends supertest but I can definitely swap it out!

packages/monitoring-system/data-processor/src/data-processor-factory.ts
```diff
@@ -0,0 +1,46 @@
...
+  getDataProcessor(task: Task): DataProcessor;
+}
+
+export class DataProcessorFactory implements Factory {
+  /**
+   * Return a relevant data processor for the given task
+   * @param task a processing task
+   * @throws if no appropriate data processor is found for task
+   */
+  public getDataProcessor(task: Task): DataProcessor {
```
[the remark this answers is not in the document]
azizsonawalla: Haha yep, my thought exactly, until I realized that you can't declare static functions in a TypeScript interface. So if I want to have the `Factory` interface (which I do general abstraction and mocking in tests) I need to keep this non-static

packages/monitoring-system/data-processor/src/data-processor-factory.ts
azizsonawalla: This is basically future-proofing the code. If someone adds another value to the `Task` enum in the future and forgets to update the switch here, this error will be helpful in debugging.

TomKristie: My question is more if the error logged/output anywhere. If we don't log it/output it, can we observe the error thrown? For instance the caller in `TaskService` catches the error but does nothing with the error. I

packages/monitoring-system/data-processor/src/data-processor-factory.ts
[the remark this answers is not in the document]
TomKristie: Looks like there is some workaround using decorators, not sure if this is of interest: https://stackoverflow.com/a/43674389.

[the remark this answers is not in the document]
azizsonawalla: Yeah I saw that as well, but I figure that this class will only be called once per function execution and so there's not really much of a performance advantage to making it static. The workaround will add more code and make it more complex as opposed to just having to add the `new` keyword in front of the call.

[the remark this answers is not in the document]
azizsonawalla: Thanks for looking into it though!

packages/monitoring-system/data-processor/src/data-processor-factory.ts
azizsonawalla: ah ok, I see what you mean. So there was a reason I don't log the error in `TaskService` or even return the info in the rejected response - basically it's to avoid credentials leaks. I expected that the most probable errors in this application will be auth errors and so I didn't want to send the auth error back in a response or put it into the logs. 

But in this case, I can log it from within DataProcessorFactory

azizsonawalla: The ideal way to do it would be to run the error through DLP and then log it, but baby steps

PR #706, opened 2020-07-19
fix(sync-repo-settings): run on all installed repositories

JustinBeckwith:


sofisl:
Is this PR in response to #623?

JustinBeckwith:
Yes!

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -83,15 +62,39 @@ function handler(app: Application) {
     const name = context.payload.repository.name;
     const repo = `${owner}/${name}`;
 
-    // find the repo record in repos.json
-    const repos = await handler.getRepos();
-    const yoshiRepo = repos.repos.find(x => x.repo === repo);
-    console.log(yoshiRepo?.repo);
-    if (!yoshiRepo) {
+    // Fetch the list of languages used in this repository
+    const langRes = await context.github.repos.listLanguages({
```
sofisl, line 66: This is cool! I didn't know you could do this.

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -83,15 +62,39 @@ function handler(app: Application) {
...
+
+    // Use the language with the highest line count
+    let language = languages[0][0].toLowerCase();
+
+    // Add a little hackery for node
+    if (language === 'javascript' || language === 'typescript') {
+      language = 'nodejs';
+    }
+    console.log(`Determined ${repo} is ${language}`);
+
```
sofisl, line 94: should we add some use cases *just in case* Github returns a language we either don't support/don't recognize for some reason? or, at least logging out a message saying which language we returned for the repo?

sofisl: On second thought, I guess it should just error out in case the language is not supported. We'll at least know what language it is based on the logging!

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -83,15 +62,39 @@ function handler(app: Application) {
...
+    });
+
+    // Given an object like this:
+    // {
+    //   python: 24
+    //   javascript: 22
+    // }
+    // Sort the keys based on instance count.
+    const languages = Object.entries(langRes.data).sort((a, b) => {
+      return a[1] > b[1] ? -1 : a[1] < b[1] ? 1 : 0;
```
tbpg, line 78: Consider:

```suggestion
      return a[1] - b[1];
```

JustinBeckwith: This is a kinda weird API where `1` means `a` is greater than `b`, `-1` means `b` is greater than `a`, and `0` means `a` equals `b`.  We are doing it backwards so the highest number comes first.  I don't think a subtraction would work.

tbpg: Ah, I missed it was swapped. `b[1] - a[1]` should work. It can be any value greater or less than 0 -- doesn't need to be 1 or -1 exactly.

| Relation | Result |
| --- | --- |
| a > b | b - a < 0 |
| a = b | b - a = 0 |
| a < b | b - a > 0 |

JustinBeckwith: Ah, that makes a lot more sense!  Nice find.  Will include this in a follow PR I'm building now.

PR #709, opened 2020-07-20
chore: add repo override for proto-plus-python

busunkim96:
@software-dov Could you check the list of required checks? They were accidentally erased (and replaced with "kokoro") when I added proto-plus to Sloth's list of repos.

### DIFF
--- a/packages/sync-repo-settings/src/required-checks.json
+++ b/packages/sync-repo-settings/src/required-checks.json
@@ -198,7 +198,7 @@
       "cla/google"
     ],
     "repoOverrides": [
-    {
+      {
         "repo": "googleapis/proto-plus-python",
         "requiredStatusChecks": [
           "ci/circleci: docs",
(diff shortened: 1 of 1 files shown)


packages/sync-repo-settings/src/required-checks.json
```diff
@@ -197,6 +197,22 @@
...
+          "ci/circleci: style-check",
+          "ci/circleci: unit-3.5",
+          "ci/circleci: unit-3.6",
+          "ci/circleci: unit-3.7",
+          "ci/circleci: unit-cpp-3.5",
+          "ci/circleci: unit-cpp-3.6",
+          "ci/circleci: unit-cpp-3.7",
+          "cla/google"
+        ]
+      },
```
JustinBeckwith, line 214: No trailing commas in JSON :(

busunkim96: Removed

PR #710, opened 2020-07-20
fix: exclude monitoring-system from builds

sofisl:

### DIFF
--- a/scripts/build-functions.sh
+++ b/scripts/build-functions.sh
@@ -27,7 +27,7 @@ npm install -g typescript
 cd packages || exit 1
 for f in *; do
     # Skip symlinks and our gcf-utils/generate-bot directory as it is not a function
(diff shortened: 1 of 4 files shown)


scripts/build-functions.sh
```diff
@@ -27,7 +27,7 @@ npm install -g typescript
 cd packages || exit 1
 for f in *; do
     # Skip symlinks and our gcf-utils/generate-bot directory as it is not a function
-    if [[ -d "$f" && ! -L "$f" && "$f" != "gcf-utils" && "$f" != "generate-bot" ]]; then
+    if [[ -d "$f" && ! -L "$f" && "$f" != "gcf-utils" && "$f" != "generate-bot" && "$f" != "monitoring-system" ]]; then
```
JustinBeckwith, line 30: This is scaling poorly, and I think we need to separate the `packages` folder into `bots` and `tools`.  Maybe file a ticket to track?

sofisl: good call!

PR #719, opened 2020-07-21
fix(mog): do not double schedule work

bcoe:
Merge on green already has retry logic, I believe using background tasks for it was leading to a thundering herd of retries.

### DIFF
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -315,12 +315,12 @@ export class GCFBootstrapper {
         // installation ID:
         try {
           if (wrapOptions?.background) {
-            console.info('scheduling cloud task');
+            console.info(`${id}: scheduling cloud task`);
             await this.handleScheduled(id, request, name, signature);
           } else {
             // a bot can opt out of running through tasks, some bots do this
             // due to large payload sizes:
-            console.info('skipping cloud tasks');
+            console.info(`${id}: skipping cloud tasks`);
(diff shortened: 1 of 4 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -314,7 +314,19 @@ export class GCFBootstrapper {
         // clever and instead pull up a list of repos we're installed on by
         // installation ID:
         try {
-          await this.handleScheduled(id, request, name, signature);
+          if (wrapOptions?.background) {
+            console.info('scheduling cloud task');
```
JustinBeckwith, line 318: I suspect we'll see a lot of those messages ...  any identifier we can attach to make this more useful when reading the logs?

packages/merge-on-green/src/app.ts
```diff
@@ -16,4 +16,6 @@ import {GCFBootstrapper} from 'gcf-utils';
 import appFn from './merge-on-green';
 
 const bootstrap = new GCFBootstrapper();
-module.exports['merge_on_green'] = bootstrap.gcf(appFn);
+module.exports['merge_on_green'] = bootstrap.gcf(appFn, {
+  background: false,
```
JustinBeckwith, line 20: I don't have a great understanding of what `background: false` really does to be honest.  Could I trouble you for a somewhat wordy explanation here, and the other places we use it?

bcoe: added comments 👍

PR #720, opened 2020-07-21
Revert "fix(sync-repo-settings): always add team memberships"

sofisl:
Reverts googleapis/repo-automation-bots#713

Screen Shot 2020-07-21 at 12 39 22 PM

Not catching the error is causing there to be a backup in the Cloud Tasks. I think what's happening is that these tasks stay in the queue because they're unsuccessful, and bottlenecking future tasks.

Maybe there's a better solution than reverting this PR?

### DIFF
--- a/packages/sync-repo-settings/src/app.ts
+++ b/packages/sync-repo-settings/src/app.ts
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import {GCFBootstrapper} from 'gcf-utils';
-import {handler} from './sync-repo-settings';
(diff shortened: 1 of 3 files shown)


bcoe:
@sofisl rather than reverting, I wonder if we could catch this specific type of error? I'd love us, as a goal for architecture, to mostly let failures happen and retry, and to identify errors that aren't of this category.

bcoe:
closing in favor of https://github.com/googleapis/repo-automation-bots/pull/722.

PR #722, opened 2020-07-22
fix(sync-repo-settings): catch specific categories of errors

bcoe:
_@sofisl @JustinBeckwith this is what I had in mind, and is closer to how I've approached handling errors in production Node.js applications._

1. you should catch the error close to the source.
2. rather than suppressing all errors, you should make an error to suppress the specific errors you understand.

In this PR I worked through all the jobs that were failing and retrying, due to some state of a repository on googleapis, and documented the potential failures.

This does not revert back to catching all errors, so should continue to appropriately bubble issues with nock.

### DIFF
--- a/packages/sync-repo-settings/src/sync-repo-settings.ts
+++ b/packages/sync-repo-settings/src/sync-repo-settings.ts
@@ -146,21 +146,34 @@ async function updateMasterBranchProtection(repo: Repo, context: Context) {
       checks = customConfig.requiredStatusChecks;
     }
   }
-  await context.github.repos.updateBranchProtection({
-    branch: 'master',
-    owner,
-    repo: name,
-    required_pull_request_reviews: {
-      dismiss_stale_reviews: false,
-      require_code_owner_reviews: false,
-    },
-    required_status_checks: {
-      contexts: checks,
-      strict: config.requireUpToDateBranch,
-    },
-    enforce_admins: true,
-    restrictions: null!,
-  });
+  try {
+    await context.github.repos.updateBranchProtection({
+      branch: 'master',
+      owner,
+      repo: name,
+      required_pull_request_reviews: {
+        dismiss_stale_reviews: false,
+        require_code_owner_reviews: false,
+      },
+      required_status_checks: {
+        contexts: checks,
+        strict: config.requireUpToDateBranch,
+      },
+      enforce_admins: true,
+      restrictions: null!,
+    });
+  } catch (err) {
+    if (err.status === 401) {
+      console.warn(
+        `updateMasterBranchProtection: received 401 error updating ${owner}/${name}`
+      );
+    } else {
+      console.error(
+        `updateMasterBranchProtection: received ${err.status} error updating ${owner}/${name}`
+      );
+      throw err;
+    }
+  }
   console.log(`Success updating master branch protection for ${repo.repo}`);
 }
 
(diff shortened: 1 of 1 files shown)


packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -146,21 +146,34 @@ async function updateMasterBranchProtection(repo: Repo, context: Context) {
...
+      required_status_checks: {
+        contexts: checks,
+        strict: config.requireUpToDateBranch,
+      },
+      enforce_admins: true,
+      restrictions: null!,
+    });
+  } catch (err) {
+    if (err.status === 401) {
+      console.warn(
```
JustinBeckwith, line 167: Why is a 401 ok in this case?  Any time we do this, adding a meaty comment explaining the situation would be lovely.

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -193,17 +206,34 @@ async function updateRepoTeams(repo: Repo, context: Context) {
...
+          owner,
+          org: owner,
+          permission: membership.permission as 'push',
+          repo: name,
+        });
+      })
+    );
+  } catch (err) {
+    const knownErrors = [
+      401, // bot does not have permission to access this repository.
```
JustinBeckwith, line 223: This is interesting!  Do you know what situation could lead to that?  Is it when someone installs the app, then we changed the required permissions after the fact?

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -193,17 +206,34 @@ async function updateRepoTeams(repo: Repo, context: Context) {
...
+      404, // team being added does not exist on repo.
```
JustinBeckwith, line 224: Again, interesting!  I'd love to know which teams we're missing that's creating the problem and go address it.  This brings up two issues:
- Did you notice in the logs which ones came up?  
- On an ongoing basis, what would be the right way to look for this situation and react to it?  Or by swallowing the error, are we saying "we are ok never addressing this"?

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -223,13 +253,30 @@ async function updateRepoOptions(repo: Repo, context: Context) {
...
+      repo: name,
+      owner,
+      allow_merge_commit: false,
+      allow_rebase_merge: config.enableRebaseMerge,
+      allow_squash_merge: config.enableSquashMerge,
+    });
+  } catch (err) {
+    const knownErrors = [
+      401, // bot does not have permission to access this repository.
+      403, // thrown if repo is archived.
```
JustinBeckwith, line 268: This is a fun one.  In the general sense - should bots ever run against archived repositories?  I'm wondering if we should do this check up front and then quietly exit if it is.

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -146,21 +146,34 @@ async function updateMasterBranchProtection(repo: Repo, context: Context) {
...
+      },
+      required_status_checks: {
+        contexts: checks,
+        strict: config.requireUpToDateBranch,
+      },
+      enforce_admins: true,
+      restrictions: null!,
+    });
+  } catch (err) {
+    if (err.status === 401) {
```
sofisl, line 166: Just as a heads up, I tried this with merge-on-green as well (catching a specific error). Tyler and I were working on surfacing a specific error message when the bot didn't have access to push to a repository. It seemed like github at times was reusing error codes? In any case, it made me lose confidence that all error codes related to all errors in a 1:1 relationship.

packages/sync-repo-settings/src/sync-repo-settings.ts
sofisl: This method itself is kind of finicky... these are all the requirements:

- The authenticated user must have admin access to the repository, and must be able to see the team
- The repository must be owned by the organization, or a direct fork of a repository owned by the organization. 

I suspect it's caused by (1) failing for new repositories, as the bot may not have admin access to that repo.

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -193,17 +206,34 @@ async function updateRepoTeams(repo: Repo, context: Context) {
...
-        org: owner,
-        permission: membership.permission as 'push',
-        repo: name,
-      });
-    })
-  );
+  try {
+    await Promise.all(
+      teamsToAdd.map(membership => {
+        return context.github.teams.addOrUpdateRepoInOrg({
```
sofisl, line 212: I think this might actually be ```addOrUpdateRepoPermissionsInOrg``` (https://octokit.github.io/rest.js/v18#teams-add-or-update-repo-permissions-in-org)

packages/sync-repo-settings/src/sync-repo-settings.ts
bcoe: @sofisl I think you found that the bot needed to explicitly be added as an admin to some repos or something? is this an extra permission setting that's configurable?

PR #730, opened 2020-07-22
feat(slo-stat-bot): Added scheduler for cron events

hershey127:


packages/slo-stat-bot/src/slo-bot.ts
```diff
@@ -40,17 +63,16 @@ async function handleIssues(
   owner: string,
   repo: string,
   type: string,
+  number: number,
```
JustinBeckwith, line 66: This is probably too many parameters for a function to be easily readable.  It feels like many of these fields would be available on an `Issue` interface.  I'm also guessing in the `Octokit` namespace there would be a `Issue` type we could readily use here.

packages/slo-stat-bot/src/slo-bot.ts
```diff
@@ -99,6 +121,34 @@ async function getSloFile(
...
+  try {
+    const state = 'open';
+    const issueList = await github.issues.listForRepo({
+      owner,
+      repo,
+      state,
+    });
+    return issueList.data;
+  } catch (err) {
+    console.error(
```
JustinBeckwith, line 145: A few things here!
- For all logging, we're now using `logger.error`, where `logger` comes from `import {logger} from 'gcf-utils';`
- When dealing with errors, it's import you both log the error, *and* it's call stack. The pattern we're starting to widely use here is:

```js
err.message = `Error in getting list of issues from repo ${repo}: ${err.message}`;
logger.error(err);
```

Logging the entire error object ensures all fields will be included and we use the right logging level in Cloud Logging.  Please do a double check and make sure this is happening broadly!

packages/slo-stat-bot/src/slo-bot.ts
orthros: I agree; I suggested she do a second PR to do that fix in order to keep this PR relatively small and stable. Let us know if you want to add this to this PR though

packages/slo-stat-bot/src/slo-bot.ts
hershey127: Sorry that's my bad. Didn't catch that inconsistency but will fix it!

packages/slo-stat-bot/src/slo-bot.ts
hershey127: As Colin mentioned I just created the new PR which includes the issue types: https://github.com/googleapis/repo-automation-bots/pull/795

PR #731, opened 2020-07-22
feat(monitoring-system): add a data processor for Cloud Tasks data

azizsonawalla:
Add a data processor for Cloud Tasks data to monitoring-system/data-processor.

Also adds a `MockFirestore` client and `MockCloudTasksClient`, where the former can be reused for testing other data processors too.

Note: I'll be adding integration tests (that will use [Firestore Emulator](https://github.com/firebase/quickstart-testing)) in another PR -- thought it would be too big for this PR. I'm not using it for the current unit tests because AFAIK:
(1) it can only test security rules and access -- so can't purposefully make it throw things like connection errors, etc
(2) it needs to be spun-up as a separate application before running tests and that seems odd to do for unit tests


packages/monitoring-system/data-processor/package.json
```diff
@@ -4,10 +4,11 @@
   "description": "Data ingestion and transformation system for GitHub Repo Automation Bots",
   "scripts": {
     "compile": "tsc -p .",
+    "clean-compile": "rm -rf ./build; rm -rf ./node_modules; npm install && npm run compile",
     "pretest": "npm run compile",
     "fix": "gts fix",
     "lint": "gts check",
-    "test": "cross-env LOG_LEVEL=fatal c8 mocha --exit ./build/test",
+    "test": "cross-env LOG_LEVEL=fatal c8 mocha --exit --recursive ./build/test/unit",
```
azizsonawalla, line 11: Moved tests to `test/unit` because I'll be adding integration tests under `test/integration`

packages/monitoring-system/data-processor/package.json
```diff
@@ -20,15 +21,17 @@
   "author": "asonawalla@google.com",
   "license": "Apache-2.0",
   "dependencies": {
-    "express": "^4.17.1",
-    "firebase-admin": "^9.0.0"
+    "@google-cloud/firestore": "^4.1.1",
```
azizsonawalla, line 24: Using the `@google-cloud` version of the Firestore client for now as @bcoe recommended, but @chingor13 suggested that we might want to use the `firebase-admin` version to be consistent with the client we'll use for the web app/browser. 

Either way, it should be easy to swap one for the other at a later stage

packages/monitoring-system/data-processor/src/data-processors/cloud-tasks-data-processor.ts
```diff
@@ -13,9 +13,127 @@
...
+  [queueName: string]: number;
+}
+
+/**
+ * Collects the current status of Cloud Task queues associated
+ * with repo automation bots and inserts it into Firestore
+ */
 export class CloudTasksProcessor extends DataProcessor {
+  private PROJECT_ID = 'repo-automation-bots';
+  private LOCATION = 'us-central1';
```
azizsonawalla, line 35: Not sure if `PROJECT_ID` and `LOCATION` should be hardcoded (either in the class like here, or in a separate config file) or if we should be pulling this information from GCP somehow. There's no guarantee though that this data processor will always be in the same project/location as the bots so we can't just use the 'current' project/location.

PR #733, opened 2020-07-23
feat: logger must be turned on with config option

bcoe:
The release of `gcf-utils` with a custom logger seems to be causing issues with `release-please`.

I think this is probably a snowflake, because release-please bot passes the GitHub instance around, and uses many more methods than most of our bots. However, I still think we should try rolling out the logger gradually given this hiccup.

A bot can simply opt in via:

```js
gcf(handler, {
  background: true,
  logging: true
})
```

_Note: I've already tested this code with release-please._

**I've turned logging on for build cop to start, since it already had configuration, let's QA this tomorrow @azizsonawalla.**

### DIFF
--- a/packages/gcf-utils/package.json
+++ b/packages/gcf-utils/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gcf-utils",
-  "version": "5.2.5",
+  "version": "5.2.6",
   "description": "An extension for running Probot in Google Cloud Functions",
   "scripts": {
     "compile": "tsc -p .",
@@ -43,7 +43,6 @@
     "@types/sinon": "^9.0.0",
     "@types/tmp": "^0.2.0",
     "@types/yargs": "^15.0.0",
-    "@octokit/rest": "^16.43.1",
     "c8": "^7.1.0",
     "cross-env": "^7.0.0",
     "dotenv": "^8.0.0",
--- a/packages/gcf-utils/package.json
+++ b/packages/gcf-utils/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gcf-utils",
-  "version": "5.2.3",
+  "version": "5.2.5",
   "description": "An extension for running Probot in Google Cloud Functions",
   "scripts": {
     "compile": "tsc -p .",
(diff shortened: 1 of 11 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -190,8 +194,14 @@ export class GCFBootstrapper {
...
     const config = JSON.parse(payload);
-    const LoggingOctokit = Octokit.plugin(LoggingOctokitPlugin);
-    return {...config, Octokit: LoggingOctokit} as Options;
+    if (logging) {
+      console.info('custom logging instance enabled');
+      const LoggingOctokit = Octokit.plugin(LoggingOctokitPlugin);
+      return {...config, Octokit: LoggingOctokit} as Options;
+    } else {
+      console.info('custom logging instance not enabled');
+      return config;
```
azizsonawalla, line 203: I think this will need to be `return config as Options;` -- that is what it was before I added the logging part.

packages/release-please/package.json
```diff
@@ -30,7 +30,7 @@
   "dependencies": {
     "@octokit/rest": "^18.0.0",
     "@octokit/webhooks": "^7.1.1",
-    "gcf-utils": "^5.0.0",
+    "gcf-utils": "^5.2.4",
```
azizsonawalla, line 33: you mean `^5.2.5` ?

packages/gcf-utils/src/gcf-utils.ts
azizsonawalla: Just for type safety purposes -- don't think it will make a difference at runtime

packages/gcf-utils/package.json
```diff
@@ -43,7 +43,6 @@
     "@types/sinon": "^9.0.0",
     "@types/tmp": "^0.2.0",
     "@types/yargs": "^15.0.0",
-    "@octokit/rest": "^16.43.1",
```
bcoe, line 46: because of the backflips being done by `release-please` to use octokit@18, even though probot is pinning us to 16, adding this dep causes issues with tests.

Issue #744, opened 2020-07-23
Minor logging issue with Merge On Green


azizsonawalla:
When MOG ignores labels, it tries to log them but the logs show up like:

```
textPayload: "21:01:38.815Z  INFO probot: ignoring non-force label action ([object Object], [object Object], [object Object])" 
```

**Solution**: stringify the objects before logging

PR #745, opened 2020-07-23
fix(merge-on-green): stringify object before logging

azizsonawalla:
Fix minor logging issue with MOG

Fixes https://github.com/googleapis/repo-automation-bots/issues/744


azizsonawalla:
PS: do not merge this -- I'm also testing my Octokit logging plugin with MOG

sofisl:
Thanks for fixing this!

PR #754, opened 2020-07-26
feat(sync-repo-settings): support repo level config

JustinBeckwith:
Add support for storing a `.github/sync-repo-settings.yaml` in the installed github repository, instead of relying on the centralized config.  The change is happening in stages.  This PR:
- Updates the format of the language used in `required-checks.json` to more closely match the new schema we want to use in individual repositories
- Does a check for a `.github/sync-repo-settings.yaml` file in the target repository, and prefers this configuration over top of the centrally provided config in `required-checks.json`
- Switches to a system where repo permissions (previously `teams.json`) is now part of the per-repository configuration, and not a global.  

This change will need to happen in stages.  Followup PRs will include:
- Creating PRs in synthtool to add templates in Python, Node.js, and Java for individual repo configs
- Working the total number of required config in `required-checks.json` down to 0, and deleting it
- Triggering the workflow on pushes to this specific config file, so users don't have to wait until night time
- Status checks for validating the config on pull requests


JustinBeckwith:
Good feedback :) I tried to expand heavily on the description of what's happening.

JustinBeckwith:
Good call out - added!

Issue #727, opened 2020-07-22
Buildcop: Missing reports


tmatsuo:
This morning we had some failures, but we didn't get reports from build cop bot.
This time, the logs should be small, so I think it's another failure, so I'm filing a new bug.

[Python 3.6 build](https://source.cloud.google.com/results/invocations/0bdc07d0-caac-4b1a-861e-108b8b364441/targets)

Excerpts from the log:
```

=================================== FAILURES ===================================
_________________________ test_model_list_get_evaluate _________________________
Traceback (most recent call last):
  File "/workspace/translate/automl/model_test.py", line 80, in test_model_list_get_evaluate
    assert "evaluation_metric" in out
AssertionError: assert 'evaluation_metric' in 'name: "projects/1012616486416/locations/us-central1/models/VAR6465236123462402048/modelEvaluations/452876416420741254"\ncreate_time {\n  seconds: 1595365133\n  nanos: 496593000\n}\nevaluated_example_count: 30\n\n'
=============================== warnings summary ===============================
dataset_test.py:28
  /workspace/translate/automl/dataset_test.py:28: PytestUnknownMarkWarning: Unknown pytest.mark.slow - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
    @pytest.mark.slow

-- Docs: https://docs.pytest.org/en/latest/warnings.html
-------- generated xml file: /workspace/translate/automl/sponge_log.xml --------
=========================== short test summary info ============================
FAILED model_test.py::test_model_list_get_evaluate - assert 'evaluation_metri...
======== 1 failed, 4 passed, 1 skipped, 1 warning in 199.61s (0:03:19) =========
Command pytest --junitxml=sponge_log.xml failed with exit code 1
Session py-3.6 failed.
[Buildcop] Sending logs to Build Cop Bot...
[Buildcop] See https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop.
[Buildcop] Published sponge_log.xml (1365873899161698)!
[Buildcop] Done!
```

You can download the test.xml file from [the artifact page](https://source.cloud.google.com/results/invocations/0bdc07d0-caac-4b1a-861e-108b8b364441/artifacts/github%2Fpython-docs-samples%2Ftranslate%2Fautoml)

These are also builds with the same failure:
[...]

PR #758, opened 2020-07-27
fix(gcf-utils): add repository details to payload for pub/sub triggered executions

azizsonawalla:
Executions triggered by pub/sub need to have a `repository` object added to the payload.

Fixes https://github.com/googleapis/repo-automation-bots/issues/727


packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -374,6 +374,24 @@ describe('GCFBootstrapper', () => {
   });
 
   describe('buildTriggerInfo', () => {
+    it('returns correct pub/sub trigger info', () => {
```
bcoe, line 377: thank you for taking the time to add a test.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -157,10 +157,11 @@ export interface CronPayload {
  * Type of function execution trigger
  */
 export enum TriggerType {
-  GITHUB = 'GITHUB_WEBHOOK',
-  SCHEDULER = 'SCHEDULER',
-  TASK = 'TASK',
-  UNKNOWN = 'UNKNOWN',
+  GITHUB = 'GitHub Webhook',
```
bcoe, line 160: nit: but  curious why we also changed the value of this enum in this PR; not a blocker just a question.

azizsonawalla: just for better logs readability

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -497,8 +459,45 @@ export class GCFBootstrapper {
...
+  }
+
+  private parseRequestBody(req: express.Request): Scheduled {
+    let body = (Buffer.isBuffer(req.body)
+      ? JSON.parse(req.body.toString('utf8'))
+      : req.body) as Scheduled;
+    // PubSub messages have their payload encoded in body.message.data
+    // as a base64 blob.
+    if (body.message && body.message.data) {
+      body = JSON.parse(Buffer.from(body.message.data, 'base64').toString());
```
bcoe, line 493: Perhaps add a TODO message to add a test for this.

azizsonawalla: Yeah, our tests should have caught my oversight. I'll add that in.

PR #770, opened 2020-07-29
test(data-processor): implement an octokit middleware to help github related testing

azizsonawalla:
Octokit Middleware intercepts outgoing GitHub requests and returns mock responses instead


packages/monitoring-system/data-processor/test/unit/data-processors/mocks/octokit-middleware.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  private static instance: OctokitMiddleware;
+  private mockResponses: MockResponses = {};
+
+  /**
+   * Returns a mock Octokit instance with OctokitMiddleware
+   * injected into it.
+   */
+  public static getMockOctokit(): Octokit {
+    const OctokitWithMiddleware = Octokit.plugin(
```
chingor13, line 73: Would you ever want to chain other plugins on as well? e.g. if you wanted the logger plugin added

packages/monitoring-system/data-processor/test/unit/data-processors/mocks/octokit-middleware.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  /**
+   * Returns the singleton instance of OctokitMiddleware
+   */
+  public static getMiddleware(): OctokitMiddleware {
+    if (!this.instance) {
+      this.instance = new OctokitMiddleware();
+    }
+    return this.instance;
+  }
```
chingor13, line 88: Will a static instance cause weird test behavior if tests are running asyncronously?

packages/monitoring-system/data-processor/test/unit/data-processors/mocks/octokit-middleware.ts
```diff
@@ -0,0 +1,173 @@
...
+        };
+      };
+    };
+  };
+}
+
+/**
+ * A middleware to intercept outgoing Octokit requests and
+ * return predefined responses instead.
+ */
```
chingor13, line 60: Can we add documentation/simple usage example?

packages/monitoring-system/data-processor/test/unit/data-processors/mocks/octokit-middleware.ts
azizsonawalla: You might want to but it would be hard to support because plugins usually do one of four things:

1. Do something before a request
2. Do something after a request
3. Wrap a request
4. Add new methods

Since this plugin is wrapping the request, adding more hooks before/after might cause undefined behaviour. Also, if a plugin adds new methods, those might not be intercepted by the middleware.

packages/monitoring-system/data-processor/test/unit/data-processors/mocks/octokit-middleware.ts
azizsonawalla: Hmm, good point. Let me see if I can find a way around it. 

So far in my testing though I haven't had any issues

packages/monitoring-system/data-processor/test/unit/data-processors/mocks/octokit-middleware.ts
azizsonawalla: Sure! For now I'll add it within the file itself since we're not publishing/exporting it. When we do, I'll add usage guides in the README

azizsonawalla:
> Haven't given this a deep review yet, but this seems like a clever idea for mocking ... could potentially cleanup some of the deep nock chains we have in release-please.

Exactly my thought!

PR #772, opened 2020-07-30
fix(sync-repo-settings): add a little more logging for overrides

JustinBeckwith:


packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -113,10 +113,11 @@ export function handler(app: Application) {
       }
 
       if (languageConfig[language]?.repoOverrides) {
-        const customConfig = languageConfig[language].repoOverrides!.find(
+        const customConfig = languageConfig[language].repoOverrides?.find(
           x => x.repo === repo
         );
         if (customConfig) {
+          console.log(`Discovered override config for ${repo}`):
```
chingor13, line 120:
```suggestion
          console.log(`Discovered override config for ${repo}`);
```

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -113,10 +113,11 @@ export function handler(app: Application) {
       }
 
       if (languageConfig[language]?.repoOverrides) {
-        const customConfig = languageConfig[language].repoOverrides!.find(
+        const customConfig = languageConfig[language].repoOverrides?.find(
           x => x.repo === repo
         );
         if (customConfig) {
+          console.log(`Discovered override config for ${repo}`);
```
bcoe, line 120: could this use @azizsonawalla's new logger that's being injected?

JustinBeckwith: oh wow good catch. This won't be the last time I make that mistake.

PR #785, opened 2020-07-31
fix(merge-on-green): payload switches between Buffer and object

bcoe:
In debugging the issues with merge on green on GoogleCloudPlatform,
we ultimately were able to isolate the problem to the fact that
Cloud Scheduler tasks for googlapis were parsed as objects upon
arrival to our Cloud Function, whereas payloads for GoogleCloudPlatform
were still a Buffer.

We already have a helper that conditionally decodes the payload
appropriately let's use this.

Note: we should still try to root cause this, if only because it's
bizarre.

### DIFF
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -368,7 +368,14 @@ export class GCFBootstrapper {
             logger.info(`${id}: skipping Cloud Tasks`);
           }
           let payload = request.body;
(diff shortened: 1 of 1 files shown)


bcoe:
> I wonder out loud if there's a way to tell if we're being handed an encoded payload vs plain text, so we can key off of that instead of the trigger type?

I think we'd probably be safe to always decode the message, I thought it might be good to make the minimal change thought, and I'd like to come back with a more thorough set of unit tests for all these edge-cases.

PR #787, opened 2020-07-31
test(data-processor): add tests for cloud logs data processor

azizsonawalla:
Adds tests for the Cloud Logs Data Processor. No implementation yet for that processor, but the test cases were long enough that I think they should be their own PR.


packages/monitoring-system/data-processor/src/data-processor-factory.ts
```diff
@@ -46,19 +50,26 @@ export class DataProcessorFactory implements Factory {
...
       default:
-        console.error(`Couldn't identify a data processor for task: ${task}`)
+        console.error(`Couldn't identify a data processor for task: ${task}`);
         throw new Error(`Couldn't identify a data processor for task: ${task}`);
     }
   }
 
+  private getLogsProcessorOptions(): CloudLogsProcessorOptions {
+    return {
+      subscription: new PubSub().subscription('TODO'),
```
azizsonawalla, line 68: will add actual subscription name via config later

packages/monitoring-system/data-processor/src/data-processors/github-data-processor.ts
```diff
@@ -104,33 +104,37 @@ export class GitHubProcessor extends DataProcessor {
...
-        for (const event of eventsPayload.data) {
-          gitHubEvents.push(
-            this.githubEventResponseToEvent(
-              (event as unknown) as GitHubEventResponse
-            )
-          );
-        }
-        return gitHubEvents;
-      });
+    return (
```
azizsonawalla, line 107: I think the linter changed this file to fix indentation -- no code changes here

packages/monitoring-system/data-processor/test/unit/data-processors/cloud-logs-data-processor.spec.ts
```diff
@@ -0,0 +1,601 @@
...
+          document: {
+            ...executionRecordStart.document,
+            ...executionRecordEnd.document,
+          },
+          collectionName: 'Bot_Execution',
+        };
+
+        describe('when no execution record exists', () => {
+          it('creates a new execution record and stores execution start logs', () => {
+            return testValidMessage(VALID_MESSAGES.execution_start, [
```
chingor13, line 218: This method returns a promise - do we need to await all of these to ensure the assertion runs/completes?

packages/monitoring-system/data-processor/test/unit/data-processors/cloud-logs-data-processor.spec.ts
```diff
@@ -0,0 +1,601 @@
...
+ * @param preExistingRecords records to add to firestore before test
+ */
+async function testValidMessage(
+  message: {},
+  expectedRecords: MockRecord[],
+  preExistingRecords?: MockRecord[]
+): Promise<void> {
+  if (preExistingRecords) {
+    preExistingRecords.forEach(record => mockFirestore.addRecord(record));
+  }
```
chingor13, line 101: I would probably do this in the preamble of the test - setting up the conditions of the test rather than putting it in this test method.

packages/monitoring-system/data-processor/test/unit/data-processors/cloud-logs-data-processor.spec.ts
```diff
@@ -0,0 +1,601 @@
...
+import {CloudLogsProcessor} from '../../../src/data-processors/cloud-logs-data-processor';
+import {loadFixture} from './util/test-util';
+import {ObjectWritableMock} from 'stream-mock';
+import pino from 'pino';
+
+let mockSubscription: MockSubscription;
+let mockFirestore: MockFirestore;
+let processor: CloudLogsProcessor;
+
+const VALID_MESSAGES: {[name: string]: {}} = loadFixture(
```
chingor13, line 29: I might split each of these into their own file & constant - the aggregate of these don't seem to mean anything. Also, it might make sense to load it closer to the test it's used in to make the individual tests easier to read.

azizsonawalla: Sure, I can do that

packages/monitoring-system/data-processor/test/unit/data-processors/cloud-logs-data-processor.spec.ts
```diff
@@ -0,0 +1,601 @@
...
+  describe('collectAndProcess()', () => {
+    beforeEach(() => {
+      mockFirestore = new MockFirestore({
+        Bot_Execution: {},
+        Error: {},
+        Trigger: {},
+        Action: {},
+        Action_Type: {},
+        GitHub_Repository: {},
+        GitHub_Object: {},
```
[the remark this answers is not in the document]
azizsonawalla: Are you asking if the default should be empty collections or if this should be the default data in `MockFirestore` class?

packages/monitoring-system/data-processor/test/unit/data-processors/cloud-logs-data-processor.spec.ts
```diff
@@ -0,0 +1,601 @@
...
+      describe('execution start and execution end logs', () => {
+        // what the execution record would look like after receiving just the start log
+        const executionRecordStart = {
+          document: {
+            '4ww4alqs7ikq': {
+              execution_id: '4ww4alqs7ikq',
+              bot_id: 'merge_on_green',
+              start_time: 1595536893000,
+              logs_url:
+                'https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22' +
```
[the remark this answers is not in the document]
azizsonawalla: Sure! I didn't realize it automatically redirected!

packages/monitoring-system/data-processor/test/unit/data-processors/cloud-logs-data-processor.spec.ts
```diff
@@ -0,0 +1,601 @@
...
+          },
+          collectionName: 'Bot_Execution',
+        };
+
+        // what the execution record would look like after receiving just the end log
+        const executionRecordEnd = {
+          document: {
+            '4ww4alqs7ikq': {
+              execution_id: '4ww4q2vqvkl1',
+              bot_id: 'auto_label',
```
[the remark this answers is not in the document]
azizsonawalla: good catch!

packages/monitoring-system/data-processor/test/unit/data-processors/cloud-logs-data-processor.spec.ts
azizsonawalla: If you return a promise mocha automatically waits for it's completion. It's equivalent to doing `await` and then `done()`

packages/monitoring-system/data-processor/test/unit/data-processors/cloud-logs-data-processor.spec.ts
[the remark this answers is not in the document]
chingor13: The latter, should MockFirestore handle this for you?

packages/monitoring-system/data-processor/test/unit/data-processors/cloud-logs-data-processor.spec.ts
chingor13: Gotcha, TIL.

packages/monitoring-system/data-processor/test/unit/data-processors/cloud-logs-data-processor.spec.ts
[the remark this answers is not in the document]
azizsonawalla: `MockFirestore` is designed as a general purpose mock instead of a mock for the Data Processor's Firestore. So I didn't want to add Data Processor specific things into `MockFirestore` itself because I wanted to keep it as reusable as possible.

That being said, I could also create a `MockDataProcessorFirestore` that `extends MockFirestore` to add this default data, but it might just add more layers and complexity.

packages/monitoring-system/data-processor/test/unit/data-processors/github-data-processor.spec.ts
```diff
@@ -38,12 +38,17 @@ let mockFirestoreData1: FirestoreData;
 
 function resetMockData() {
```
azizsonawalla, line 39: I changed `loadFixture`'s `copy` param to be `false` by default instead of true so had to adjust parameters here.

packages/monitoring-system/data-processor/test/unit/data-processors/cloud-tasks-data-processor.spec.ts
```diff
@@ -40,10 +40,22 @@ describe('Cloud Tasks Data Processor', () => {
   let MockFirestoreData3: FirestoreData;
 
   function resetMockData() {
```
azizsonawalla, line 42: I changed `loadFixture`'s `copy` param to be `false` by default instead of `true` so had to adjust parameters here.

azizsonawalla:
Changes from feedback:
* Fixture files are broken-down more and organized into separate folders
* Fixtures are loaded closer to the test itself
* Fixed some issues with expected test results
* Added some missing return statements

azizsonawalla:
Unfortunately there's not really any way around that while also keeping these tests 'black-box'. Even using something like `sinon` to stub firestore would require the same data merging to be implemented within the stub and that would in fact be messier code.

The alternative would be to use spies to ensure that `firestore.update()` was called instead of `firestore.set()` but that would make these tests whitebox and more fragile/prone to breaking if the implementation is changed

Issue #777, opened 2020-07-30
feat(gcf-utils): logger should keep scope for all log statements


chingor13:
In gcf-utils, we can capture the repository/other trigger information and add that data to all structured logs. This will make logs much more discoverable in the logging console.

azizsonawalla:
Adding an example here of what we're looking for:

```ts
(in gcf-utils.ts)

const triggerInfo  = {                                     // this is what we parse from incoming payload
    trigger: {
        trigger_type: 'GITHUB_WEBHOOK',
        trigger_sender: 'some sender',
        payload_hash: '123456',
        trigger_source_repo: {
            owner: 'foo owner',
            owner_type: 'Org',
            repo_name: 'bar name',
            url: 'some url',
        },
    },
};

...

// bind the triggerInfo above to the exported logger
```

```ts
(in some-bot.ts)

logger.info({ some-property: 'some-value' });
```
```
(Cloud Logs Viewer)

{
    insertId: "xyz",
    jsonPayload: {
        some-property: "some-value",                // what the bot logged
        trigger: {                                   // what was automatically attached to logs
            trigger_type: 'GITHUB_WEBHOOK',
            trigger_sender: 'some sender',
            payload_hash: '123456',
            trigger_source_repo: {
                owner: 'foo owner',
                owner_type: 'Org',
                repo_name: 'bar name',
                url: 'some url',
            },
        },
    },
    resource: {
        type: "cloud-function",
        labels: {
            function_name: "some-bot",
            project_id: "repo-automation-bots",
            region: "us-central1"
        },
    },
    timestamp: "12345",
    severity: "INFO",
    labels: {
        execution_id: "abcd"
    }
}
```

tmatsuo:
@chingor13 Why did we re-open this issue?

chingor13:
I think we reverted the fix because of scope pollution between requests (not reset). The trigger info is not being logged on all the log messages during the request

chingor13:
Fixed in #3516 and #3582

PR #796, opened 2020-08-03
feat(gcf-utils): add trigger info as bindings to all log statements

azizsonawalla:
* Adds support for initializing GCFLogger with bindings
* Adds support for child loggers (with bindings) to GCFLogger
* Uses above mentioned child logger functionality to bind trigger information to all log statements for an execution

Resulting log statements will look something like this:

```ts
(in gcf-utils.ts)

const triggerInfo  = {
    trigger: {
        trigger_type: 'GITHUB_WEBHOOK',
        trigger_sender: 'some sender',
        payload_hash: '123456',
        trigger_source_repo: {
            owner: 'foo owner',
            owner_type: 'Org',
            repo_name: 'bar name',
            url: 'some url',
        },
    },
};

...

bindTriggerInfoToLogger(triggerInfo);
```

```ts
(in some-bot.ts)

// after gcf-utils has done all the trigger parsing and logged the trigger information the first time

logger.info({ some-property: 'some-value' });
```
```ts
(Cloud Logs Viewer)

{
    insertId: "xyz",
    jsonPayload: {
        some-property: "some-value",                // what the bot logged
        trigger: {                                   // what was automatically attached to logs
            trigger_type: 'GITHUB_WEBHOOK',
            trigger_sender: 'some sender',
            payload_hash: '123456',
            trigger_source_repo: {
                owner: 'foo owner',
                owner_type: 'Org',
                repo_name: 'bar name',
                url: 'some url',
            },
        },
    },
    resource: {
        type: "cloud-function",
        labels: {
            function_name: "some-bot",
            project_id: "repo-automation-bots",
            region: "us-central1"
        },
    },
    timestamp: "12345",
    severity: "INFO",
    labels: {
        execution_id: "abcd"
    }
}
```
Fixes https://github.com/googleapis/repo-automation-bots/issues/777

### DIFF
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -18,7 +18,7 @@ import {v1} from '@google-cloud/secret-manager';
 import * as express from 'express';
 // eslint-disable-next-line node/no-extraneous-import
 import {Octokit} from '@octokit/rest';
-import {buildTriggerInfo, TriggerInfo} from './logging/trigger-info-builder';
+import {buildTriggerInfo} from './logging/trigger-info-builder';
 import {GCFLogger, initLogger} from './logging/gcf-logger';
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const LoggingOctokitPlugin = require('../src/logging/logging-octokit-plugin.js');
@@ -182,18 +182,16 @@ export class GCFBootstrapper {
   }
 
   /**
-   * Binds the trigger info to the logger so that every logged statement
-   * includes the trigger info. Does not bind the message property.
+   * Binds the given properties to the logger so that every logged statement
+   * includes these properties.
    *
    * NOTE: statements logged before this method is called will not include
-   * the trigger info
-   * NOTE: this method modifies the triggerInfo parameter given
+   * this information
    *
-   * @param triggerInfo trigger information for current execution
+   * @param bindings object containing properties to bind to the logger
    */
-  private static bindTriggerInfoToLogger(triggerInfo: TriggerInfo) {
-    delete triggerInfo.message;
-    logger = logger.child(triggerInfo);
+  private static bindPropertiesToLogger(bindings: {}) {
+    logger = logger.child(bindings);
   }
 
   gcf(
@@ -217,7 +215,9 @@ export class GCFBootstrapper {
 
       const triggerInfo = buildTriggerInfo(triggerType, id, request.body);
       logger.metric(triggerInfo);
-      GCFBootstrapper.bindTriggerInfoToLogger(triggerInfo);
+
+      delete triggerInfo.message; // we don't want to bind the message to every log entry
+      GCFBootstrapper.bindPropertiesToLogger(triggerInfo);
 
       try {
         if (triggerType === TriggerType.UNKNOWN) {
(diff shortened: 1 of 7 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -181,6 +181,21 @@ export class GCFBootstrapper {
...
+   * includes the trigger info. Does not bind the message property.
+   *
+   * NOTE: statements logged before this method is called will not include
+   * the trigger info
+   * NOTE: this method modifies the triggerInfo parameter given
+   *
+   * @param triggerInfo trigger information for current execution
+   */
+  private static bindTriggerInfoToLogger(triggerInfo: TriggerInfo) {
+    delete triggerInfo.message;
```
bcoe, line 195: because `triggerInfo` passes by reference, this  will actually  remove the `message` parameter from the originating object,  this might not be the intended  behavior.

You could:

```js
const meta = {...triggerInfo}
logger = logger.child(meta);
```
```

Comment (azizsonawalla) [reply packages/gcf-utils/src/gcf-utils.ts:195]: The removal of the property directly from the `triggerInfo` was intentional (I made a note of this in the method docs) because I thought cloning the object would be an expensive operation and right now we aren't using the `triggerInfo` object anywhere else after this method. 

However, if you think it might lead to some 'gotcha' situations in the future, I can use your suggestion above to clone it (or sometimes I even do `const meta = JSON.parse(JSON.stringify(triggerInfo))` for quick cloning)

Comment (azizsonawalla) [reply packages/gcf-utils/src/gcf-utils.ts:195]: Also added an example to the PR description above -- will also add it to the issue

Comment (chingor13) [reply packages/gcf-utils/src/gcf-utils.ts:195]: I agree with @bcoe here on this one. Modifying the input object can lead to error prone code in the future. Also, the description of the function and name don't describe the modification of the input.

For performance optimizations like this, we should also consider the amount of memory saved in practice - the trigger info is really small (< 1KB) so this gain is probably negligible.

Comment (chingor13) [reply packages/gcf-utils/src/gcf-utils.ts:195]: If you do want to still delete the field in this manner, you can do it outside of this function in the caller.

File: packages/gcf-utils/src/gcf-utils.ts [hunk 2]
```diff
@@ -200,7 +213,11 @@ export class GCFBootstrapper {
         taskId
       );
 
-      logger.metric(buildTriggerInfo(triggerType, id, request.body));
+      const triggerInfo = buildTriggerInfo(triggerType, id, request.body);
+      logger.metric(triggerInfo);
+
+      delete triggerInfo.message; // we don't want to bind the message to every log entry
+      GCFBootstrapper.bindPropertiesToLogger(triggerInfo);
```
chingor13, line 220: Could we make `buildTriggerInfo` not include `message`, set the logger to the child logger, then do `logger.metric("message originally in buildTriggerInfo")`?

azizsonawalla: You mean like this?

```ts
      const triggerInfo = buildTriggerInfo(triggerType, id, request.body);
      GCFBootstrapper.bindPropertiesToLogger(triggerInfo); // no message in triggerInfo

      logger.metric({ 
        message: `Execution started by ${triggerType}`, 
        ...triggerInfo
      });
```
This works too!

chingor13: Since the trigger info should already be added to all the log entries, you should only have to call 
```typescript
logger.metric(`Execution started by ${triggerType}`);
```

azizsonawalla: ahh ok ok - makes sense

azizsonawalla: done!

PR #800, opened 2020-08-04
fix(sync-repo-settings): attempt freezes to avoid mutating global state

JustinBeckwith:


chingor13:
Looks like we found the bug

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -107,7 +122,7 @@ export function handler(app: Application) {
       }
       logger.info(`Determined ${repo} is ${language}`);
 
-      config = languageConfig[language];
+      config = extend(true, {}, languageConfig)[language];
```
chingor13, line 125: nit, you could probably move the indexed value into extend so we don't have to copy extra entries.

JustinBeckwith: I'll suffer the inefficiency

PR #806, opened 2020-08-05
feat(data-processor): implement Cloud Logs data processor, add more type checking, move constants to config

azizsonawalla:
* Implement Cloud Logs Data Processor
* Add type checking helpers for Cloud Logs
* Reorganize type checking helpers into separate directory
* Move Cloud Logs subscription name and timeout to config file


packages/monitoring-system/data-processor/src/data-processors/cloud-logs-data-processor.ts
```diff
@@ -48,7 +88,341 @@ export class CloudLogsProcessor extends DataProcessor {
...
+            this.logger.debug('All messages processed');
+            resolve();
+          })
+          .catch(error => {
+            this.logger.error('Some messages could not be processed');
+            reject(error);
+          });
+      };
+
+      setTimeout(stopListening, this.listenLimit * 1000);
```
chingor13, line 120: nit: can you add comments about why this method is using a setTimeout

packages/monitoring-system/data-processor/src/data-processors/cloud-logs-data-processor.ts
```diff
@@ -48,7 +88,341 @@ export class CloudLogsProcessor extends DataProcessor {
...
-    throw new Error('Method not implemented.');
+    this.subscription.on('message', this.processMessage.bind(this));
+    this.logger.debug('Now listening for PubSub messages');
+
+    return new Promise((resolve, reject) => {
+      const stopListening = () => {
+        this.subscription.removeAllListeners();
+        this.logger.debug(`Stopped listening after ${this.listenLimit}s`);
+
+        Promise.all(this.tasksInProgress)
```
chingor13, line 109: Does this check every message in the list or does it short-circuit?

azizsonawalla: It short-circuits (i.e. `rejects` with the first rejected promise in the array) but recall that we only `reject` promises for runtime errors (eg. Pub/Sub connectivity issues or Firestore issues) and not for issues related to individual messages (eg. a malformed log entry).

So it makes sense to short-circuit on the first rejection because for runtime errors it's highly probable that the other promises down the line will face the same issue and just throw errors.

packages/monitoring-system/data-processor/src/data-processors/cloud-logs-data-processor.ts
azizsonawalla: will do!

PR #813, opened 2020-08-06
fix(slo-stat-bot): Does not throw error when org level config is missing but logs an error

hershey127:


packages/slo-stat-bot/src/slo-bot.ts
```diff
@@ -118,10 +118,14 @@ async function getSloFile(
 
   if (!sloRules) {
     path = 'issue_slo_rules.json';
+    logger.warn(
```
bcoe, line 121: With some of the other bots, we check for `CONFIGURATION_FILE_PATH` and if it doesn't exist the bot is a `noop`.

This allows us to enable the bot org-wide, without creating warnings.

What's our plan with this bot, do we think that we'll enable it globally with a `googleapis/.github` repository, or do we expect to turn it on for specific repositories.

PR #824, opened 2020-08-10
feat(data-processor): add better type checking, include updates from gcf-utils, and reduce duplicated code

azizsonawalla:
* Adds a centralized method to update documents in Firestore
* Adds method `getPrimaryKey` to Firestore schema to build the primary key for any document
* Update GitHub processor constant values to match updates from gcf-utils:
    * `Org` -> `Organization`
    * remove owner type from repository primary key
* Add type checking helpers

PR #825, opened 2020-08-10
feat(data-processor): add a centralized logger

azizsonawalla:
Adds a centralized Pino logger to be used across all modules

Depends on https://github.com/googleapis/repo-automation-bots/pull/824


packages/monitoring-system/data-processor/src/data-processors/cloud-tasks-data-processor.ts
```diff
@@ -87,7 +91,7 @@ export class CloudTasksProcessor extends DataProcessor {
   private async getBotNames(): Promise<string[]> {
     return new Promise<string[]>((resolve, reject) => {
       this.firestore
-        .collection('Bot')
+        .collection(FirestoreCollection.Bot)
```
bcoe, line 94: I like that you moved to an `enum` for this 👍 this should make renaming the collection easy ... speaking of which:

Combing from my relational database background, if I saw a table called `bot` or `bots`, I would assume it contains the bots themselves (_perhaps we serialize data bout the bots, and they're bootstrapped from this table).

I would be tempted to call the collection that collects metrics something like `metrics`, or `bot-metrics`, which speaks to the type of data being collected.

azizsonawalla: So the `Bot` collection in Firestore actually stored information on all our current bots - which as of now is limited to the bot name. In the future, this might include more information regarding the bot itself.

Execution related information is in a separate collection called `Bot_Execution`. You can see the entire schema [here](https://github.com/googleapis/repo-automation-bots/blob/6fcda088d41e8bdd9e8c5cde0c8929396e74eccb/packages/monitoring-system/data-processor/src/firestore-schema.ts#L218-L252)

PR #828, opened 2020-08-10
feat(blunderbuss): ignore draft PRs

JustinBeckwith:
Best viewed with whitespace changes hidden in the PR


packages/blunderbuss/src/blunderbuss.ts
```diff
@@ -54,158 +56,169 @@ export const sleep = (ms: number) => {
...
+    } catch (err) {
+      err.message = `Error reading configuration: ${err.message}`;
+      logger.error(err);
+    }
+    config = config || {};
+
+    const issue: Issue = context.payload.issue || context.payload.pull_request;
+
+    // If this is a PR, and it's in draft mode, don't assign it
+    if (issue.draft === true) {
```
kurtisvg, line 85: Are we sure we want to ignore draft PRs, even when triggered with the issue?

packages/blunderbuss/src/blunderbuss.ts
```diff
@@ -54,158 +56,169 @@ export const sleep = (ms: number) => {
...
+      config = (await context.config<Configuration>(
         CONFIGURATION_FILE_PATH,
         {}
-      )) as Configuration;
-      const issue = (context.payload.issue
-        ? context.payload.issue
-        : context.payload.pull_request) as Issue;
+      ))!;
+    } catch (err) {
+      err.message = `Error reading configuration: ${err.message}`;
```
bcoe, line 77: I believe this will lose the stack trace, I would perhaps instead do `${err.stack}`.

JustinBeckwith: Huh - i was modifying the message of the original error specifically so I *would* preserve the call stack!  I think this works right (please let me know if I'm wrong):

```js
try {
  throw new Error('this is busted!');
} catch (e) {
  e.message = `There was a problem: ${e.message}`;
  throw e;
}
```

Leads to:

```
/Users/beckwith/Code/repo-automation-bots/packages/blunderbuss/test.js:5
  throw e;
  ^

Error: There was a problem: this is busted!
    at Object.<anonymous> (/Users/beckwith/Code/repo-automation-bots/packages/blunderbuss/test.js:2:9)
    at Module._compile (internal/modules/cjs/loader.js:1201:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1221:10)
    at Module.load (internal/modules/cjs/loader.js:1050:32)
    at Function.Module._load (internal/modules/cjs/loader.js:938:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47
```

The call stack looks to point to the correct line of code, and I think it just modifies the message part.  Right?

packages/blunderbuss/src/blunderbuss.ts
JustinBeckwith: I mean, "sure" is a funny word :) This came up in the chat with @lesv and I thought it sounded reasonable.  Are there situations where you think users would wan to assign draft PRs?  I always looked at draft as a way to signal "don't look at me yet, just stashing stuff here"

kurtisvg: I mean "label" - if a user adds the label while the PR is in draft, should we assign it anyway?

lesv: Which "label" does that make sense for?  If it's draft, they are saying don't review, or they want a review from someone specific.  They might still want to use kokoro for a bit to test, they can add those labels if they are on the team.  We just don't get bothered until they think they are ready to merge.

JustinBeckwith: I could see the confusion there.  You apply the label specifically to get some blunderaction, and it does nothing.  That having been said - I think this is a reasonable restriction.

packages/blunderbuss/src/blunderbuss.ts
bcoe: This looks good to me, I guess updating the message preserves the stack trace, works for me.

PR #836, opened 2020-08-12
feat(slo-stat-bot): Lints for contradicting SLOs

hershey127:


packages/slo-stat-bot/src/slo-lint.ts
```diff
@@ -145,18 +147,57 @@ async function getFileShaContents(
...
+    errors: errors,
   } as ValidationResults;
 };
 
+/**
+ * Function checks to see if there is a label in both gitHubLabels and excludedGitHubLabels
+ * @param slo rule
+ * @returns boolean value if slo is self contradicting
+ */
+async function sloNotContradict(slo: SLORules): Promise<boolean | undefined> {
```
orthros, line 188: nit: I would change this to `sloIsContradictory` or `sloContradicts` so that it reads more naturally from callers. Having "nots" in your function names can lead to confusing calling code e.g. `if !sloNotContradict` has a double negative in it.

packages/slo-stat-bot/src/slo-lint.ts
```diff
@@ -31,7 +33,7 @@ type Conclusion =
 
 interface ValidationResults {
   isValid: boolean;
-  errors?: ErrorObject[] | null;
+  errors?: ErrorObject[] | JSON | null;
```
JustinBeckwith, line 36: What is `JSON` in this context?  If it's a generic property bag, there are specific types I could see us using like:
- `{}` - I know this is an object, and I won't pretend to guess what the properties are
- `any` - I don't know what is going on here, and it's impossible to know.  We really avoid this if we can :)
- `{[index: string]: string}` - I know these are string based key value pairs, but that's about it

packages/slo-stat-bot/src/slo-lint.ts
```diff
@@ -145,18 +147,57 @@ async function getFileShaContents(
...
+
+  if (validate.errors) {
+    return {
+      isValid: isValid,
+      errors: validate.errors,
+    } as ValidationResults;
+  }
+
+  //Check for contradicting SLOs
+  let errors = null;
```
JustinBeckwith, line 164: You typically don't need to explicitly assign to `null` like this!  Using `let errors: Error[];` will default to `undefined`, which in most cases here is perfectly ok

packages/slo-stat-bot/src/slo-lint.ts
```diff
@@ -145,18 +147,57 @@ async function getFileShaContents(
...
+  for (const slo of sloData) {
+    isValid = !(await sloContradicts(slo));
+
+    if (!isValid) {
+      const sloString = JSON.stringify(slo);
```
JustinBeckwith, line 169: This is going to provide some ugly output usually.  To get prettier formatting for display in a call stack, I use `JSON.stringify(slo, null, 2);`.  Try this out in the node repl!

packages/slo-stat-bot/src/slo-lint.ts
```diff
@@ -145,18 +147,57 @@ async function getFileShaContents(
...
 };
 
+/**
+ * Function checks to see if there is a label in both gitHubLabels and excludedGitHubLabels
+ * @param slo rule
+ * @returns boolean value if slo is self contradicting
+ */
+async function sloContradicts(slo: SLORules): Promise<boolean | undefined> {
+  const githubLabels = await convertToArray(slo.appliesTo.gitHubLabels);
+  const excludedGitHubLabels = await convertToArray(
```
JustinBeckwith, line 190: I don't know what's in `convertToArray`, but with a name like that I'm surprised it's an `async` function.

packages/slo-stat-bot/src/slo-lint.ts
```diff
@@ -145,18 +147,57 @@ async function getFileShaContents(
...
+    slo.appliesTo.excludedGitHubLabels
+  );
+
+  const isInValid = excludedGitHubLabels?.some((label: string) =>
```
JustinBeckwith, line 194: If `excludedGitHubLabels` is of type `string[]`, you shouldn't need to explicitly define the type in the arrow function.  `.some(label => githubLabels?.includes(label))` should do just fine.

packages/slo-stat-bot/src/slo-lint.ts
hershey127: That makes sense. I will fix it!

packages/slo-stat-bot/src/slo-lint.ts
hershey127: I tried doing that but the output prints with '\n' in the message. I think this might be caused by calling `JSON.stringify(slo, null, 4)`  in the `createCheck` method

packages/slo-stat-bot/src/slo-lint.ts
hershey127: I can create a type for displaying a message.

sofisl:
@orthros - is this OK to merge without @hershey127? It looks like she addressed all of the comments.

sofisl:
@hershey127, you may need to sign a CLA again!

PR #844, opened 2020-08-14
feat(snippet-bot): check matching region tags

### LINKED ISSUE #843: snippet-bot: implement simple check for mismatched region tags

tmatsuo:
* simplified config
* add tests

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/repo-automation-bots/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #843  🦕


packages/snippet-bot/package.json
```diff
@@ -30,7 +30,11 @@
   },
   "dependencies": {
     "@octokit/rest": "^18.0.0",
+    "@types/minimatch": "^3.0.3",
```
JustinBeckwith, line 33: You can move this into `devDependencies`

packages/snippet-bot/package.json
```diff
@@ -30,7 +30,11 @@
...
     "gcf-utils": "^5.3.0",
+    "install": "^0.13.0",
```
JustinBeckwith, line 35: You can delete `npm` and `install` from the list here

packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -12,28 +12,127 @@
...
 
-interface Configuration {
-  randomBoolean: boolean;
+class Configuration {
+  private options: ConfigurationOptions;
+  private minimatches: minimatch.IMinimatch[];
+
+  constructor(options: ConfigurationOptions) {
+    this.options = options;
+    this.minimatches = (options as ConfigurationOptions).ignoreFiles.map(
```
JustinBeckwith, line 41: I don't think you need the cast here `as ConfigurationOptions`

packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -12,28 +12,127 @@
...
+      return mm.match(filename);
+    });
+  }
 }
 
 export = (app: Application) => {
-  app.on(['issues.opened', 'pull_request.opened'], async context => {
-    const config = (await context.config(
+  app.on('pull_request', async context => {
+    const configOptions = (await context.config(
```
JustinBeckwith, line 57: It's important we wrap this call here in a `try`/`catch`.  See:
https://github.com/googleapis/repo-automation-bots/blob/master/packages/sync-repo-settings/src/sync-repo-settings.ts#L81

If the bot doesn't have code read access on setup, it will throw

packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -12,28 +12,127 @@
...
       CONFIGURATION_FILE_PATH,
       {}
-    )) as Configuration;
+    )) as ConfigurationOptions;
```
JustinBeckwith, line 60: The original code here was wrong.  Instead of casting to `ConfigurationOptions`, you can pass it as a generic parameter to the `context.config` method like:
```js
const configOptions = await context.config<ConfigurationOptions>( ...
```

packages/snippet-bot/test/fixtures/config/valid-config.yml
```diff
@@ -1 +1,2 @@
-randomBoolean: true
\ No newline at end of file
+ignoreFiles:
```
JustinBeckwith, line 1: Huh - I'm curious.  What's a situation where we'd detect a region tag, and then ignore it?

packages/snippet-bot/test/snippet-bot.ts
```diff
@@ -13,10 +13,13 @@
...
+/* eslint-disable @typescript-eslint/no-var-requires */
+
 import myProbotApp from '../src/snippet-bot';
 
 import {resolve} from 'path';
 import {Probot} from 'probot';
+import snapshot from 'snap-shot-it';
 import nock from 'nock';
 import * as fs from 'fs';
 import {expect} from 'chai';
```
JustinBeckwith, line 25: Let's remove `chai` here, and instead use `import assert from 'assert'`.  `assert` is baked into node.js, and doesn't require a third party dependency.

tmatsuo: done

packages/snippet-bot/test/fixtures/config/valid-config.yml
tmatsuo: I think there should be a way to opt out certain files.
For example, I want to exclude readme which contains an example of mismatched region tag.

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: Nah, it's still optional type

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: Done

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: done

packages/snippet-bot/package.json
tmatsuo: done

packages/snippet-bot/package.json
tmatsuo: done

packages/snippet-bot/test/fixtures/config/valid-config.yml
tmatsuo: As I did in the last commit

tmatsuo:
header-check is failing, but I think #846 will fix it

PR #847, opened 2020-08-14
build: include package-lock files again

JustinBeckwith:
Bringing lock files back.  Lock files are at best contentious for libraries, but best practices for executables (our bots are).  Adding this to protect us from weird octokit shenanigans.


stephenplusplus:
Does this bring them back to the library repos, (e.g. @google-cloud/storage)?

JustinBeckwith:
Nope! Just for this repository where we're shipping applications.  I'm curious - do you _want_ them for libraries?

stephenplusplus:
Woops, sorry I didn't pick up on that. I much prefer us not using them from libraries, I think I just wasn't sure from your post if we were going to be like "eh, we'll deal with it" :) Glad to hear we will keep them in the app space!

packages/snippet-bot/package-lock.json
```diff
@@ -0,0 +1,6498 @@
+{
```
tmatsuo, line 1: I've just updated `package.json` in this directory. Do you need to update the lock file with the `package.json` in master?

JustinBeckwith: Nope!  The robots will take care for us.

PR #848, opened 2020-08-14
chore(auto-label): clean up a few functions

JustinBeckwith:


packages/auto-label/src/auto-label.ts
```diff
@@ -261,9 +174,14 @@ handler.addLabeltoRepoAndIssue = async function addLabeltoRepoAndIssue(
...
+          .addLabels({
+            owner,
+            repo,
+            issue_number: issueNumber,
+            labels: [githubLabel],
+          })
+          .catch(logger.error);
         logger.info(
           `Label added to ${owner}/${repo} for issue ${issueNumber} is ${githubLabel}`
         );
```
tbpg, line 187: Should this just throw the error, rather than log the error then say, "label added"? Same below.

JustinBeckwith: Probably.  This was supposed to be a behavior consistent refactor, so I kept the behavior.  Will come back around and fix up a few of these odds and ends.

PR #850, opened 2020-08-15
chore(snippet-bot): remove problematic deps

tmatsuo:
also pin octokit version

Somehow the previous combination of the dependencies caused problems on Cloud Function environment. I'm still not 100% sure it's going to fix the problem, but I don't have local repro anyways, I think I need to try.

### DIFF
--- a/packages/snippet-bot/package.json
+++ b/packages/snippet-bot/package.json
@@ -29,9 +29,8 @@
     "lint": "gts check"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


tmatsuo:
@JustinBeckwith It fails to compile for me when I removed it. I'm going to merge this.

tmatsuo:
FYI, the bot works now :)

https://github.com/GoogleCloudPlatform/python-docs-samples/pull/4507

PR #853, opened 2020-08-17
fix(data-processor): add a `private` field to GitHub repositories to reduce unnecessary API calls

azizsonawalla:
* Adds a `private` field to GitHub repositories in Firestore.
* Only makes GitHub API calls for repositories that are not marked `private = true`
* If GitHub returns `Not Found` error for a repository, marks that repository as `private = true`

This avoids making lots of failed calls to GitHub and triggering the 'abuse detection mechanism'


packages/monitoring-system/data-processor/.gitignore
```diff
@@ -0,0 +1 @@
+package-lock.json
```
chingor13, line 1: These were explicitly enabled recently.

packages/monitoring-system/data-processor/src/data-processors/github-data-processor.ts
```diff
@@ -84,58 +89,104 @@ export class GitHubProcessor extends DataProcessor {
...
 
   /**
-   * Get all the publicly visible Events on the given repository
+   * Get recent 100 publicly visible Events on the given repository
    * @param repository repository for which to get events
+   * @returns A Promise with a list of GitHubEventDocuments. Promise
+   * will always resolve - if there is an error, it will resolve with
+   * an empty array and error will be logged.
+   *
+   * NOTE: We only fetch the first page (last 100) events given
```
chingor13, line 117: Will we be notified somehow if this happens? Or will we silently start missing events?

packages/monitoring-system/data-processor/.gitignore
azizsonawalla: Ah ok, I thought something changed in the config on my side. I'll remove the ignore.

packages/monitoring-system/data-processor/src/data-processors/github-data-processor.ts
azizsonawalla: No notification, but I think we'll notice if we start seeing bot executions without a mapped GitHub event on the dashboard.

The core issue here is that there's no definite 'ordering' to the events. In other words, if the latest event I have in Firestore has `timestamp = 1234`, I cannot guarantee that the next time I call the API I will have all the events with `timestamp < 1234` because the timestamp refers to the time of the event and not when the event was registered by the API. So an event with `timestamp=1134` may be registered at time `1344`. 

This leaves us with 2 possible solutions:

1. Make the change now to read all 300 events (this will triple the number of GitHub API calls)
2. Keep an eye on the dashboard and make the change later. At this point we'll probably also have to register the Data Processor as a GitHub app to increase the number of requests it can make.

azizsonawalla:
> What happens if a repository switches between private and public?

This is a good point, and right now I don't have an elegant solution for it. The only answer that comes to mind is to run a "refresh" job every night / every few nights to check the status of the repos again. I could create this job as a separate endpoint in Data Processor and schedule it for every few days.

For now, I can add a `TODO` to the code above for this.

PR #858, opened 2020-08-18
fix(release-please): clean up dependencies

### LINKED ISSUE #792: Support @octokit/webhooks versions > 7.1.1 by changing import strategy
[@octokit/webhooks.js](https://www.npmjs.com/package/@octokit/webhooks) changed the way they export the `Webhooks` type i[n a recent commit](https://github.com/octokit/webhooks.js/commit/12a266a34ad4781975038a8a78908481460f5fce#diff-41df78d86ed4de80aae9b0b9560feb2b) and this [breaks the release-please tests](https://github.com/googleapis/repo-automation-bots/pull/791/checks?check_run_id=939823269#step:8:1) for versions above 7.1.1. The fix should be fairly straightforward but it will require some testing to ensure nothing else is broken. 

https://github.com/octokit/webhooks.js/pull/113

### LINKED ISSUE #834: bug: refactor for octokit/webhooks
We either should refactor release-please to not use octokit/webhooks or add it to ignore modules
See #816

JustinBeckwith:
Fixes #792
Fixes #834

### DIFF
--- a/packages/release-please/.c8rc.json
+++ b/packages/release-please/.c8rc.json
@@ -1,0 +1,3 @@
+{
+  "exclude": [".mocharc.js", "build/test"]
+}
--- a/packages/release-please/.gitignore
+++ b/packages/release-please/.gitignore
@@ -1,0 +1,3 @@
+node_modules
+build/
+coverage/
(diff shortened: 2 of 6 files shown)


bcoe:
we've bumped into some nasty issues  in interdependencies  between probot and release-please. Landing this, but will keep a close eye and there's a chance we need to roll back.

PR #866, opened 2020-08-18
fix(header-checker-lint): catch configuration parse errors and log

### LINKED ISSUE #863: `Unidentified alias` error in header_checker_lint
Seeing this spammed in the logs:
```
unidentified alias "*/__snapshots__/*.js" at line 2, column 26:
      - **/__snapshots__/*.js
```

chingor13:
Turns out this is an error parsing the configuration yaml from our own repo. We now catch exceptions reading the configuration file and log an error/halt if the configuration is invalid).

This also fixes the bad configuration file (the `.github/header-checker-lint.yaml`) and removes the deprecated `.bots/header-checker-lint.json` file.

Fixes #863


packages/header-checker-lint/src/header-checker-lint.ts
```diff
@@ -74,6 +75,23 @@ class Configuration {
 
 export = (app: Application) => {
   app.on('pull_request', async context => {
+    let remoteConfiguration: ConfigurationOptions = DEFAULT_CONFIGURATION;
```
JustinBeckwith, line 78: Assuming `DEFAULT_CONFIGURATION` is of type `ConfigurationOptions`, the explicit type of `: ConfigurationOptions` is not needed.

packages/header-checker-lint/src/header-checker-lint.ts
```diff
@@ -74,6 +75,23 @@ class Configuration {
...
+    try {
+      const canditateConfiguration = await context.config(
```
JustinBeckwith, line 80: Instead of the `if` below to do the case, you should be able to do:
```js
remoteConfiguration = await context.config<ConfigurationOptions>( ...
```

I think you can skip the separate candidateConfiguration thingy
config = config ? 

Then outside of the catch:
```js
remote

Comment (chingor13) [reply packages/header-checker-lint/src/header-checker-lint.ts:80]: `context.config()` can return `null` so I think I still need to check the return

Comment (alexander-fenster) [inline packages/header-checker-lint/src/header-checker-lint.ts:80]: typo: should be `candiDate`

File: packages/header-checker-lint/src/header-checker-lint.ts (hunk 1 above)
Comment (chingor13) [reply packages/header-checker-lint/src/header-checker-lint.ts]: Done

File: packages/header-checker-lint/src/header-checker-lint.ts (hunk 2 above)
Comment (chingor13) [reply packages/header-checker-lint/src/header-checker-lint.ts]: Done.

File: packages/header-checker-lint/src/header-checker-lint.ts [hunk 3]
```diff
@@ -74,6 +75,23 @@ class Configuration {
 
 export = (app: Application) => {
   app.on('pull_request', async context => {
+    let remoteConfiguration = DEFAULT_CONFIGURATION;
+    try {
+      const candidateConfiguration = await context.config<ConfigurationOptions>(
+        WELL_KNOWN_CONFIGURATION_FILE
```
bcoe, line 81: Do you think it would be worth pulling this logic into a helper that all bots share at some point, seems like we're starting to collect a few lines of code.

Perhaps a tracking ticket to do this?

packages/header-checker-lint/src/header-checker-lint.ts
```diff
@@ -87,19 +105,11 @@ export = (app: Application) => {
     try {
       filesResponse = await context.github.pulls.listFiles(listFilesParams);
     } catch (err) {
-      app.log.error('---------------------');
-      app.log.error(err);
+      logger.error('---------------------');
```
bcoe, line 108: I don't sully understand how to add custom labels in the logs, I know @azizsonawalla was showing this functionality off today. Would this be an option to help this log jump out, rather than `-------`.

packages/header-checker-lint/test/header-checker-lint.ts
```diff
@@ -256,6 +256,18 @@ describe('HeaderCheckerLint', () => {
       requests.done();
     });
 
+    it('handles an invalid configuration file', async () => {
```
bcoe, line 259: 👍 thank you for test.

packages/header-checker-lint/src/header-checker-lint.ts
chingor13: Absolutely, but maybe not in this pull request. It might even be nice to validate against a schema like SLO bot does.

Opened https://github.com/googleapis/repo-automation-bots/issues/870

packages/header-checker-lint/src/header-checker-lint.ts
chingor13: I will remove the '-----' logging - it's not useful

Issue #842, opened 2020-08-13
bug (MOG): Let's find a better way of dealing with quota issues:


sofisl:
We're running up against quota and getting this: 
```You have triggered an abuse detection mechanism. Please wait a few minutes before you try again.```

In turn, MOG just fails quietly and doesn't do anything helpful. We should have MOG do something more helpful, at least removing the label if possible.

Issue #781, opened 2020-07-31
mog: use page iterator to iterate over status check & check run responses


sofisl:
https://octokit.github.io/rest.js/v18#pagination

PR #877, opened 2020-08-19
fix(mog): rate limit issues

### LINKED ISSUE #821: bug(MOG): MOG is returning empty lists when there is an error getting the branch protection
MOG is returning empty lists when there is an error getting branch protection. It should just fail instead.

sofisl:
Why is this PR so big?

1. I began by adding a rate limit block to have us not execute on PRs when we've hit the limit. This allows the PR to stay in circulation until we get back to a healthy state (#821 and #842).
2. Then, I tried running tests and realized tests were broken (#878). So, I wanted to get tests to a healthy state to pass. One of the changes that needed to be made was to use pagination in order to not get errors for empty pages (solving #781).
3. In order to limit some of our rate limits, I also moved our check for branch protection requirements to the wrapper logic around the MOG logic. So, MOG will only call this once, and store it in the database, lowering our overall calls. I still added Issue #918 to better improve this functionality.


packages/merge-on-green/package.json
```diff
@@ -28,7 +28,8 @@
   },
   "dependencies": {
     "@google-cloud/datastore": "^6.1.0",
-    "gcf-utils": "^5.5.1"
+    "gcf-utils": "^5.5.1",
+    "probot": "^9.13.0"
```
JustinBeckwith, line 32: Do we need to specifically bring in this version?  I've very much been trying to avoid diamond dependencies given our reliance here on `gcf-utils`.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -125,6 +125,14 @@ handler.addPR = async function addPR(wp: WatchPR, url: string) {
  */
 function handler(app: Application) {
   app.on(['schedule.repository'], async context => {
+    const rateLimit = (await context.github.rateLimit.get()).data.resources.core
```
JustinBeckwith, line 128: When this happens - do we think it's a regular "log" event, or an "error"?  Thinking out loud - I think this is an error.  It means we've built things in a way that it can't live within the API rate limits we're given and it's setting us up for trouble.  Not sure what the answer is to manage it moving forward, but I don't want to treat this as business as usual when it happens.

packages/merge-on-green/package.json
sofisl: I think we need it to use it to define a type.

packages/merge-on-green/src/merge-on-green.ts
sofisl: Great point. I made some changes to try to lower the calls, but I agree. Created tracking issue #918 for etags.

packages/merge-on-green/package.json
JustinBeckwith: Sure - but it's packaged in `gcf-utils` as a dependency, so technically you don't need to define it here.  Conventional wisdom says you are doing the right thing, but this is a bit of a weird situation.  Probot and Octokit have particularly bad diamond dependency issues.  With defining the version of probot here *and* in `gcf-utils`, we run the risk of them getting stuck on different, incompatible versions.  To ensure we only ever get a single version, we specifically don't include it in `package.json` for projects that already rely on `gcf-utils`.  Now, you will get an eslint warning.  It's ok to ignore the line where you import `probot` (like all the other bots)

sofisl:
@chingor13, we don't have a test case to check midway through, but I don't immediately see a use case unless we were going to put it in the logs - not sure if that's too noisy though. I also considered commenting on PRs when we're close to the end, but I also feel that's too noisy. 

This call does not have a quota!

PR #894, opened 2020-08-22
feat: migrate to node.js 12.x

JustinBeckwith:
We're at the stage where we want to reliably use `fs.promises`, the `recursive` option on `rmdir`, and a few other niceties.  Node.js 12 is technically beta on GCF, but that's really just for bake time as far as I'm being told.  The current version on GCF is `12.18.2`, so I'm just sticking to that as a bottom part of the range specifier.  Let's wait until Monday to merge this one 🙃


JustinBeckwith:
Throwing up the `do not merge` while we wait for publishbot to be replaced with kokoro again.

scripts/publish-function.sh
```diff
@@ -38,7 +38,7 @@ queueName=${botName//_/-}
 echo "About to publish function ${functionName}"
 gcloud functions deploy "${functionName}" \
   --trigger-http \
-  --runtime nodejs10 \
+  --runtime nodejs12 \
```
tmatsuo, line 41: How about to parameterize this like 

```
--runtime "${functionRuntime}" \
```

Then in the bot directory:

```
if [[ -f .env_deployment ]]; then
    source .env_deployment
fi

if [[ -z "${functionRuntime:-}" ]]; then
    # defaults to nodejs10
    functionRuntime=nodejs10
fi 
```
or something?

tmatsuo: Or maybe we can determine the nodejs runtime by grep against package.json .

tmatsuo:
#901 for parameterized version

bcoe:
Have opted for #901, the plan when we swap in new Node.js versions is that we'll test on a few then move forward with the  upgrade.

PR #910, opened 2020-08-26
build: automate deployment for dashboard

azizsonawalla:
Automate the deployment of the Metrics Dashboard to firebase hosting.

This will require us to give the Cloud Build service account access to read the secret `firebase-token-metrics-dashboard` in repo-automation-bots


packages/monitoring-system/metrics-dashboard/cloudbuild.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - '-c'
+      - "gcloud secrets versions access latest --secret=firebase-token-metrics-dashboard --format='get(payload.data)' | tr '_-' '/+' | base64 -d > decrypted-data.txt"
+
+  - name: gcr.io/cloud-builders/npm
+    id: "deploy-dashboard"
+    waitFor: ["get-firebase-token"]
+    entrypoint: bash
+    args:
+      - "-e"
+      - "./packages/monitoring-system/metrics-dashboard/scripts/deploy.sh  $(cat decrypted-data.txt)"
```
chingor13, line 30: I'd make the argument the file and not the actual secret. Rather not provide raw credentials via command line options (as well as if there are weird characters).

Additionally, you'll probably need to quote the value if you're using it on the command line

packages/monitoring-system/metrics-dashboard/cloudbuild.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - '-c'
+      - "gcloud secrets versions access latest --secret=firebase-token-metrics-dashboard --format='get(payload.data)' | tr '_-' '/+' | base64 -d > firebase-token.txt"
+
+  - name: gcr.io/cloud-builders/npm
+    id: "deploy-dashboard"
+    waitFor: ["get-firebase-token"]
+    entrypoint: bash
+    args:
+      - "-e"
+      - "./scripts/deploy.sh firebase-token.txt"
```
azizsonawalla, line 30: This assumes that the `pwd` for the Cloud Build is the directory where the `cloudbuild.yml` file is. Not sure if this assumption is correct?

chingor13: the default working directory is `/workspace` which is the root of the repository (for triggered cloud builds). You could provide a `dir` option: https://cloud.google.com/cloud-build/docs/build-config#dir

packages/monitoring-system/metrics-dashboard/scripts/deploy.sh
```diff
@@ -0,0 +1,17 @@
...
+  echo "Usage $0 <path-to-token-file>"
+  exit 1
+fi
+
+# compile
+npm install
+npm run compile
+
+# deploy to Firebase
+firebase deploy --token $(cat $1)
```
chingor13, line 17: Do we need to quote the token?

azizsonawalla: I don't believe it's necessary -- I've tested it with an actual token and seems to work fine -- but I can add it anyway

packages/monitoring-system/metrics-dashboard/scripts/deploy.sh
```diff
@@ -0,0 +1,17 @@
...
+  echo "Usage $0 <path-to-token-file>"
+  exit 1
+fi
+
+# compile
+npm install
+npm run compile
+
+# deploy to Firebase
+firebase deploy --token '$(cat $1)'
```
chingor13, line 17: I think you want double-quotes or it'll use the raw `$(cat $1)` as a string

packages/monitoring-system/metrics-dashboard/cloudbuild.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+steps:
+  - name: gcr.io/cloud-builders/gcloud
+    id: "get-firebase-token"
+    waitFor: ["-"]
+    dir: './packages/monitoring-system/metrics-dashboard'
+    entrypoint: 'bash'
+    args: 
+      - '-c'
+      - "gcloud secrets versions access latest --secret=firebase-token-metrics-dashboard --format='get(payload.data)' | tr '_-' '/+' | base64 -d > firebase-token.txt"
```
chingor13, line 23: You should be able to put each of these on its own line

packages/monitoring-system/metrics-dashboard/scripts/deploy.sh
```diff
@@ -0,0 +1,17 @@
+# build and deploy the dashboard from a fresh clone
```
chingor13, line 1: Needs license header

packages/monitoring-system/metrics-dashboard/cloudbuild.yaml
azizsonawalla: are you sure? I thought the whole line was passed as a string to bash because of the `-c` flag

packages/monitoring-system/metrics-dashboard/scripts/deploy.sh
azizsonawalla: yeah I wasn't sure about that. I'll use double!

packages/monitoring-system/metrics-dashboard/scripts/deploy.sh
azizsonawalla: good catch -- I guess support for `.sh` files still needs to be added to the license checker bot

packages/monitoring-system/metrics-dashboard/cloudbuild.yaml
azizsonawalla: I can move it to a separate `.sh` file if you want, that might be cleaner

packages/monitoring-system/metrics-dashboard/scripts/deploy.sh
```diff
@@ -0,0 +1,33 @@
...
+  echo "Usage $0 <path-to-token-file>"
+  exit 1
+fi
+
+# compile
+npm install
+npm run compile
+
+# deploy to Firebase
+firebase deploy --token "$(cat $1)"
```
bcoe, line 33: do we  need to give CloudBuild permission to perform a deploy?

azizsonawalla: the token covers that

packages/monitoring-system/metrics-dashboard/cloudbuild.yaml
azizsonawalla: PS: The official docs have the whole line in one arg - https://cloud.google.com/cloud-build/docs/securing-builds/use-encrypted-secrets-credentials

azizsonawalla:
@orthros would you be able to grant the Cloud Build service account (`service-856896688174@gcp-sa-cloudbuild.iam.gserviceaccount.com`) privileges to read the secret `firebase-token-metrics-dashboard` in `repo-automation-bots`? Thanks!

azizsonawalla:
Just tested this one locally, and as expected, it gets permission denied when trying to read the token:

```
Starting Step #0 - "get-firebase-token"
Step #0 - "get-firebase-token": Already have image (with digest): gcr.io/cloud-builders/gcloud
Step #0 - "get-firebase-token": ERROR: (gcloud.secrets.versions.access) PERMISSION_DENIED: Permission 'secretmanager.versions.access' denied for resource 'projects/repo-automation-bots/secrets/firebase-token-metrics-dashboard/versions/latest' (or it may not exist).
Finished Step #0 - "get-firebase-token"
ERROR
ERROR: build step 0 "gcr.io/cloud-builders/gcloud" failed: step exited with non-zero status: 1
```

JustinBeckwith:
@chingor13 I am going to go ahead and land this - let me know if there's anything we need to clean up!

Issue #905, opened 2020-08-25
chore: migrate bots to nodejs12 runtime


tmatsuo:
Cloud Function provides nodejs12 runtime. This is still Beta, but we think it's stable enough.
We want to migrate the bots to nodejs12 runtime.

This can be done by per bot basis.
- add "nodejs12" to your `cloudbuild.yaml` ([example](https://github.com/googleapis/repo-automation-bots/blob/bd938928481fbd537cd059983b10a062bde986d8/packages/release-please/cloudbuild.yaml#L38))
- change node version in package.json ([example](https://github.com/googleapis/repo-automation-bots/blob/bd938928481fbd537cd059983b10a062bde986d8/packages/release-please/package.json#L49))
- remove "prepare" script in `package.json`([example PR](https://github.com/googleapis/repo-automation-bots/pull/902)) if you have one

I think we should circulate this issue among bot owners and each bot owners do the migration.

- [x] auto-label
- [x] blunderbuss
- [x] buildcop
- [x] conventional-commit-lint
- [x] failurechecker
- [x] generate-bot
- [x] header-checker-lint
- [x] label-sync
- [x] merge-on-green
- [x] monitoring-system
- [x] release-please
- [ ] slo-stat-bot
- [x] snippet-bot
- [ ] sync-repo-settings
- [ ] trusted-contribution

I don't think we'll migrate `publish` to nodejs12. Let me know if I'm wrong.

PR #911, opened 2020-08-26
chore: opt in 4 bots to use nodejs12 runtime

tmatsuo:
- failurechecker
- header-checker-lint
- label-sync
- merge-on-green

Part of #905


tmatsuo:
Let's merge this tomorrow morning so that we can monitor the bots.

Issue #789, opened 2020-07-31
gcf-utils: github trigger info should include event type


chingor13:
We should be able to distinguish which type of event triggered the webhook.

PR #921, opened 2020-08-27
feat(gcf-utils): log event type with trigger information

azizsonawalla:
Logs the event type along with the trigger information.

PS: most of the lines in this PR are from fixtures

Resolves https://github.com/googleapis/repo-automation-bots/issues/789


packages/gcf-utils/src/logging/trigger-info-builder.ts
```diff
@@ -63,47 +66,72 @@ export function buildTriggerInfo(
...
+}
 
-  return triggerInfo;
+/**
+ * Returns a description of the GitHub Event type
+ * @param requestBody the body of the incoming GitHub Webhook request
+ */
+function getEventTypeDetails(eventName: string, actionValue: string): string {
+  eventName = eventName === '' ? 'UNKNOWN' : eventName;
+  return `${eventName}${actionValue ? `: ${actionValue}` : ''}`;
```
chingor13, line 129: I might make this concatenate using a `.` to match what others do (like probot). `issue.labeled` or `pull_request.merged` etc.

PR #922, opened 2020-08-27
testing: introduce `node` matrix

### LINKED ISSUE #892: testing: use the same nodejs version as the production
I think now we're using different nodejs versions in our tests and the production environment.

In https://github.com/googleapis/repo-automation-bots/pull/889, I tried to use `fs.rmdirSync(tmpDir.name, {recursive: true});`. The test passes locally, as well as in our CI builds.

`recursive` option was introduced in nodejs 12.10.0:
https://nodejs.org/en/blog/release/v12.10.0/

Our cloud function environment says it's nodejs 10.

I think we should use the same version in our CI builds.

tmatsuo:
Now we only have node=12 because majority of the bots are running on
nodejs12 runtime.

We can add a new job by adding an element in `include` section of the
matrix.

For example, if you want to run the test for snippet-bot on node 14, we
can do something like:

```
    include:
      - node: 14
	package: snippet-bot
```

Fixes #892

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,6 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        node: [12]
         package: [
           'auto-label',
           'blunderbuss',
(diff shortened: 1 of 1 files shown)


tmatsuo:
If we go this approach, we'll need to change the required checks on the master branch because the Checks have slight different names (including node version).

PR #925, opened 2020-08-28
fix(trusted-contribution): don't use Cloud Tasks

### LINKED ISSUE #920: Trusted Contribution bot running into Task size limit
Seeing the following in trusted contribution logs:

```
INVALID_ARGUMENT: Task size too large
```

We probably need to disable Cloud Tasks for the trusted contribution bot.

tbpg:
Some of the tasks are running over the Cloud Tasks size limit.

Fixes #920.


JustinBeckwith:
@bcoe sadly I think we need to address this more completely

PR #939, opened 2020-09-06
feat(sync-repo-settings): add config validation

JustinBeckwith:
This adds a PR status check for changes specifically to the config for this bot.

### DIFF
--- a/packages/sync-repo-settings/src/sync-repo-settings.ts
+++ b/packages/sync-repo-settings/src/sync-repo-settings.ts
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 // eslint-disable-next-line node/no-extraneous-import
-import {Application, Context} from 'probot';
+import {Application, Context, Octokit} from 'probot';
 import extend from 'extend';
 import {
   LanguageConfig,
@@ -22,6 +22,14 @@ import {
   PermissionRule,
 } from './types';
 import {logger} from 'gcf-utils';
+import Ajv from 'ajv';
+import yaml from 'js-yaml';
+
+export const configFileName = 'sync-repo-settings.yaml';
+
+// configure the schema validator once
+const schema = require('./schema.json');
+const ajv = new Ajv();
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function deepFreeze(object: any) {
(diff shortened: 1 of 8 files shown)


packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -64,6 +72,76 @@ const branchProtectionDefaults = deepFreeze({
...
+
+        const checkParams: Octokit.ChecksCreateParams = context.repo({
+          name: 'sync-repo-settings-check',
+          head_sha: context.payload.pull_request.head.sha,
+          conclusion: 'success',
+        });
+        if (!isValid) {
+          checkParams.conclusion = 'failure',
+          checkParams.output = {
+            title: 'Invalid sync-repo-settings.yaml schema 😱',
```
tbpg, line 130: Can we get some sort of nice error message from the validation? Looking at https://ajv.js.org/#validation-errors, perhaps the `message` field?

tbpg (changes requested):

bcoe (approved):
This seems like a good idea to me. If I'm understanding, this  actually extends on the bot to provide a status check on a PR, indicating whether or not the updated configuration matches the schema.

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -64,6 +72,76 @@ const branchProtectionDefaults = deepFreeze({
...
+          file_sha: file.sha,
+        });
+        const configYaml = Buffer.from(blob.data.content, 'base64').toString('utf8');
+        const config = yaml.safeLoad(configYaml, {
+
+        });
+        let isValid = false;
+        let errorText = '';
+        if (typeof config === 'object') {
+          const validateSchema = ajv.compile(schema);
```
bcoe, line 115: this could be a nice approach to use in some other places, for instance when  providing configuration for CHANGELOG generation in release-please.

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -64,6 +72,76 @@ const branchProtectionDefaults = deepFreeze({
...
+        }
+        const blob = await context.github.git.getBlob({
+          owner,
+          repo,
+          file_sha: file.sha,
+        });
+        const configYaml = Buffer.from(blob.data.content, 'base64').toString('utf8');
+        const config = yaml.safeLoad(configYaml, {
+
+        });
```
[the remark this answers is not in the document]
JustinBeckwith: Just missed this - good catch.

packages/sync-repo-settings/src/sync-repo-settings.ts
JustinBeckwith: Yes!  So the issue is there can be many errors.  For now I created a detail view with the JSON.stringified text of all errors, and use it as the full error text for folks that dig in.

JustinBeckwith:
@tbpg ready for another look :)

packages/sync-repo-settings/src/sync-repo-settings.ts
JustinBeckwith: I want to get this all happily working, then I'd like to up-level the support to `gcf-utils`.  There are an annoying number of moving parts here, and having a generic utility feels nice.

tbpg: +1 -- this would be great for other bots. I'm very tempted to copy it as-is.

tbpg (approved):

PR #946, opened 2020-09-09
feat(blunderbuss): expand GitHub teams

### LINKED ISSUE #284: Blunderbuss: expand teams to select a member
If a team is specified as an option for either assignees or reviewers, expand the list of users to so we it can assign to an individual.

This will make maintenance of repositories easier as you can assign a repository to a combination of user(s) and team(s).

tbpg:
This applies to issues and pull requests.
* Issues can't be assigned to teams (GitHub limitation). So, no change
  in behavior.
* PRs _can_ be assigned to teams. So, this will change the behavior of
  Blunderbuss for PRs (individuals instead of teams).

REST docs:
* https://developer.github.com/v3/teams/members/#list-team-members
* https://octokit.github.io/rest.js/v18#teams

I also fixed references to the repo name when logging. Before, it was
always undefined.

I also checked the response status when making an assignment. Sometimes,
it seems users aren't assigned, but the logs would say they were.

Fixes #284 🦕


kurtisvg (approved):
Thanks for taking the time to do this.

tmatsuo (approved):

PR #963, opened 2020-09-18
refactor(probot)!: upgrade gcf-utils to probot@10, @octokit/rest@18

bcoe:
First pass at an attempt to get us on the latest version of `probot`, this is going to be somewhat gnarly.


packages/gcf-utils/package.json
```diff
@@ -51,7 +52,7 @@
     "mocha": "^8.0.0",
     "nock": "^13.0.0",
     "sinon": "^9.0.0",
-    "sonic-boom": "^1.0.1",
+    "sonic-boom": "1.1.0",
```
bcoe, line 55: the point release of sonic-boom breaks types.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -133,7 +141,7 @@ export class GCFBootstrapper {
       return {...config, Octokit: LoggingOctokit} as Options;
     } else {
       logger.info('custom logging instance not enabled');
-      return config as Options;
+      return {...config, Octokit} as Options;
```
bcoe, line 144: always use the `Octokit` constructor, I believe this will  help us avoid diamond dependency issues.

packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -68,15 +68,15 @@ describe('GCFBootstrapper', () => {
...
-          });
+      sinon
+        .stub(bootstrapper, 'getAuthenticatedOctokit')
+        .returns(Promise.resolve(new Octokit()));
+      handler = bootstrapper.gcf(async app => {
         app.on('issues', spy);
-        app.on('schedule.repository', spy);
-        app.on('err', sinon.stub().throws());
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        app.on('schedule.repository' as any, spy);
```
bcoe, line 77: Octokit is now strict about event names, I've been talking to @gr2m about how we might approach this going forward.

packages/gcf-utils/test/logging-octokit-plugin.ts
```diff
@@ -357,18 +357,4 @@ describe('Logging-Octokit-Plugin', () => {
         assert.deepEqual(actual, expected);
       });
   });
-
-  it('does not log information for unknown actions', () => {
```
bcoe, line 361: I'm not quite sure what the intent of this test was @chingor13 @azizsonawalla? but I couldn't seem to get it working on the latest `@octokit/rest`.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -281,17 +291,16 @@ export class GCFBootstrapper {
       // Job was scheduled for a single repository:
       await this.scheduledToTask(body.repo, id, body, eventName, signature);
     } else {
-      const octokit = new Octokit({
-        auth: await this.getInstallationToken(body.installation.id),
-      });
+      const octokit = await this.getAuthenticatedOctokit(body.installation.id);
       // Installations API documented here: https://developer.github.com/v3/apps/installations/
-      const installationsPaginated = octokit.paginate.iterator({
-        url: '/installation/repositories',
```
bcoe, line 289: @gr2m does this look like a reasonable update to the logic?

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -306,10 +315,11 @@ export class GCFBootstrapper {
...
   }
 
-  async getInstallationToken(installationId: number) {
-    return await this.probot!.app!.getInstallationAccessToken({
-      installationId,
-    });
+  // TODO: How do we still get access to this installation token?
+  async getAuthenticatedOctokit(installationId: number): Promise<Octokit> {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const app = (this.probot as any).apps[0] as Application;
```
bcoe, line 321: @gr2m I think this is what we talked about earlier today, that can be used to give us an Octokit instance with permission to call `octokit.apps.listReposAccessibleToInstallation`?

bcoe (commented):
@sofisl/@gr2m take a look 😄

@gr2m thanks for your help planning this upgrade so far 👏

packages/gcf-utils/test/logging-octokit-plugin.ts
chingor13: I believe we had to explicitly maintain a map of octokit functions to the data that we log. I think this test uses an untracked endpoint and thus should not log any actions.

azizsonawalla: Yep, @chingor13 is right. We only wanted to log certain calls to Octokit that bots make (i.e. calls to Octokit that actually make a change on GitHub like updating an issue, rather than just 'get' information). So this test checks that a non-tracked action does not log any information.

packages/gcf-utils/src/gcf-utils.ts
gr2m: what are "diamond dependency issues"?

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -281,17 +291,16 @@ export class GCFBootstrapper {
...
-        headers: {
-          Accept: 'application/vnd.github.machine-man-preview+json',
-        },
-      });
+      const installationsPaginated = octokit.paginate.iterator(
+        octokit.apps.listReposAccessibleToInstallation,
+        {
+          headers: {
+            Accept: 'application/vnd.github.machine-man-preview+json',
+          },
```
gr2m, line 301: the `machine-man` preview was graduated ([announcement](https://developer.github.com/changes/2020-08-20-graduate-machine-man-and-sailor-v-previews/)). If you work with GHES, you can enable previews more easily like this
```suggestion
          mediaType: {
            previews: ['machine-man'],
          },
```

packages/gcf-utils/src/gcf-utils.ts
gr2m: yes 👍🏼

packages/gcf-utils/src/gcf-utils.ts
gr2m: yes, sorry for the need of this ugly hack, I hope make it obsolete soon

packages/gcf-utils/src/gcf-utils.ts
bcoe: we were bumping into an issue where probot wanted `@octokit/rest@16.x`, whereas our library `release-please` wanted `@octokit/rest@18.x`, I don't think we need to worry about two versions of this dependency in the tree any longer.

bcoe:
@gr2m thank you for the review.

packages/gcf-utils/src/gcf-utils.ts
gr2m: Probot v10 only upgraded to v17 to ease the transition from v16, as you'd only need to address all deprecation messages in order to upgrade. Probot v11 will have `@octokit/rest` v18. If that's a problem I can cut a Probot v11 release sooner, no big deal

bcoe: @gr2m I would love to get on `v18` ideally, in testing so far I've been able to use `v18` as the dependency.

packages/gcf-utils/test/logging-octokit-plugin.ts
bcoe: added test back.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -129,11 +138,13 @@ export class GCFBootstrapper {
     const config = JSON.parse(payload);
     if (logging) {
       logger.info('custom logging instance enabled');
-      const LoggingOctokit = Octokit.plugin(LoggingOctokitPlugin);
+      const LoggingOctokit = Octokit.plugin(LoggingOctokitPlugin).plugin(
```
bcoe, line 141: this was required to continue giving us support for `context.github.config`.

gr2m: Instead of using `Octokit` from `@octokit/rest`, you can `import { ProbotOctokit } from 'probot'`, it will come with all the plugins and defaults that Probot is using internally

gr2m: You might also run into compatibility issues. Because Probot v10 is compatible with `@octokit/rest` v17, while you install v18 in your package.json. I would recommend to get rid of the `@octokit/rest` dependency altogether

packages/gcf-utils/src/gcf-utils.ts
JustinBeckwith: Just so we don't lose track, @bcoe could I trouble you to file an upstream bug and leave it in a comment here so we can come clean up?

packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -68,15 +68,15 @@ describe('GCFBootstrapper', () => {
...
       enqueueTask = sinon.stub(bootstrapper, 'enqueueTask');
-      sinon.stub(bootstrapper, 'getInstallationToken');
-      handler = await bootstrapper.gcf(async app => {
-        app.auth = () =>
-          new Promise<GitHubAPI>(resolve => {
-            resolve(GitHubAPI());
-          });
+      sinon
+        .stub(bootstrapper, 'getAuthenticatedOctokit')
+        .returns(Promise.resolve(new Octokit()));
```
JustinBeckwith, line 73: Shorten to `.resolves(new Octokit())`

packages/release-please/src/release-please.ts
```diff
@@ -190,7 +197,8 @@ export = (app: Application) => {
     }
   });
 
-  app.on('schedule.repository', async context => {
+  // See: https://github.com/octokit/webhooks.js/issues/277
+  app.on('schedule.repository' as any, async context => {
```
JustinBeckwith, line 201: Just to keep the build clean, please address the warnings here

packages/release-please/test/release-please.ts
```diff
@@ -16,8 +16,18 @@ import myProbotApp from '../src/release-please';
 import {Runner} from '../src/runner';
 import {describe, it, beforeEach} from 'mocha';
 import {resolve} from 'path';
+
+// We provide our own GitHub instnace, similar to
```
JustinBeckwith, line 20: typo in instnace :)

packages/release-please/test/release-please.ts
```diff
@@ -114,9 +120,9 @@ describe('ReleasePleaseBot', () => {
       );
       const requests = nock('https://api.github.com')
         .get(
-          '/repos/chingor13/google-auth-library-java/contents/.github/release-please.yml'
+          '/repos/chingor13/google-auth-library-java/contents/.github%2Frelease-please.yml'
         )
-        .reply(200, {content: config.toString('base64')});
```
JustinBeckwith, line 119: This is gonna break stuff in other bots

JustinBeckwith (approved):

packages/release-please/test/release-please.ts
```diff
@@ -339,16 +346,17 @@ describe('ReleasePleaseBot', () => {
...
         .delete(
-          '/repos/Codertocat/Hello-World/issues/2/labels/release-please:force-run'
+          '/repos/Codertocat/Hello-World/issues/2/labels/release-please%3Aforce-run'
         )
         .reply(200);
 
       await probot.receive({
-        name: 'pull_request.labeled',
+        // see: https://github.com/probot/probot/issues/1367
+        name: 'pull_request.labeled' as any,
```
gr2m, line 359: Remove `.labeled` from the name. `labeled` is the action, it is part of the payload: https://github.com/googleapis/repo-automation-bots/blob/e9637b11eedb2fdecf746e99dffbee347bd052a0/packages/release-please/test/fixtures/pull_request_labeled.json#L2

```suggestion
        name: 'pull_request',
```

packages/release-please/test/release-please.ts
```diff
@@ -363,7 +371,8 @@ describe('ReleasePleaseBot', () => {
       };
 
       await probot.receive({
-        name: 'pull_request.labeled',
+        // see: https://github.com/probot/probot/issues/1367
+        name: 'pull_request.labeled' as any,
```
gr2m, line 375:
```suggestion
        name: 'pull_request',
```

PR #977, opened 2020-09-25
fix(buildcop): sort dup closed issues by closed_at

### LINKED ISSUE #975: build-cop-bot: doesn't reopen most recently closed issue, so it always thinks the issue was closed > 10 days ago
If there are duplicate issues, we currently pick the one that was opened first. But, if there are duplicate _closed_ issues, that means we pick the one that was probably closed the longest time ago.

When deciding whether to reopen an issue as flaky or to create a new issue, the bot looks at how long ago the issue was closed. So, if the first issue was closed a really long time ago, the bot will _always_ think it should file a new issue.

We should take the close time into account when thinking about reopening an issue.

Example of where this came up: https://github.com/GoogleCloudPlatform/golang-samples/issues/1726

tbpg:
When deciding whether to reopen an old issue or to open a new one, we
want to find the old issue that was most recently closed. That way, if
the most recently closed issue was closed within 10 days, we know to
reopen it.

Fixes #975 🦕

### DIFF
--- a/packages/buildcop/src/buildcop.ts
+++ b/packages/buildcop/src/buildcop.ts
@@ -385,12 +385,9 @@ buildcop.openIssues = async (
         // If the existing issue has been closed for more than 10 days, open
         // a new issue instead.
         //
-        // The type of closed_at is null. But, it is actually a string if the
-        // issue is closed. Convert to unknown then to string as a workaround.
         // If this doesn't work, we'll mark the issue as flaky.
-        const closedAtString = (existingIssue.closed_at as unknown) as string;
-        if (closedAtString) {
-          const closedAt = Date.parse(closedAtString);
+        const closedAt = parseClosedAt(existingIssue.closed_at);
+        if (closedAt) {
           const daysAgo = 10;
           const daysAgoDate = new Date();
           daysAgoDate.setDate(daysAgoDate.getDate() - daysAgo);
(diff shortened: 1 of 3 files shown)


packages/buildcop/src/buildcop.ts
```diff
@@ -614,13 +611,25 @@ buildcop.issueComparator = (
...
   if (buildcop.isFlaky(a) && !buildcop.isFlaky(b)) {
     return -1;
   }
   if (!buildcop.isFlaky(a) && buildcop.isFlaky(b)) {
     return 1;
   }
-  return a.number - b.number;
+  const aClose = parseClosedAt(a.closed_at);
+  const bClose = parseClosedAt(b.closed_at);
+  if (aClose && bClose) {
```
JustinBeckwith, line 629: Just be careful about truthiness here.  This will evaluate false for undefined, null, and 0.  That may be fine since it's a date :)  I tend to prefer more strict checks.

JustinBeckwith (approved):

packages/buildcop/src/buildcop.ts
tbpg: That seems OK for this particular case. This check would not be valid for any of those values. I suppose we know that the result is `number|undefined`, though.

I'm open here. Will defer to you for JS style.

JustinBeckwith: Nah this is fine, just shining light on it :)

PR #979, opened 2020-09-25
feat(label-sync): allow a repo to be ignored

bcoe:
This will allow [api-linter](https://github.com/googleapis/api-linter/labels) to ignore our label sync logic.

CC: @lukesneeringer

### DIFF
--- a/packages/label-sync/src/label-sync.ts
+++ b/packages/label-sync/src/label-sync.ts
@@ -13,13 +13,19 @@
 // limitations under the License.
 
 // eslint-disable-next-line node/no-extraneous-import
-import {Application, GitHubAPI, Octokit} from 'probot';
+import {Application, GitHubAPI, Octokit, Context} from 'probot';
 import {createHash} from 'crypto';
 import {Storage} from '@google-cloud/storage';
 import * as util from 'util';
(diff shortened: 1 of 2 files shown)


lesv (approved):

JustinBeckwith:
Possibly annoying feedback on the approach.  In `sync-repo-settings`, having centralized config started to cause a significant amount of pain.  To move away from that, we started moving towards repo-level config.  The nice thing about that, is that a central `.github` repository can keep a generalized config that applies to all.  

I could easily see us switching the model where:
- we introduce a `github.com/googleapis/.github/.github/label-sync.yaml`
- That file has a list of all labels we want to add
- We rip the code out of the bot that talks to GCS, and move that to a simple GitHub action that on a nightly basis reads the file from GCS, modifies the `yaml` file with the list of available labels, submits a PR, and auto-approves itself

This has the benefit of making the bot more generally useful beyond the specific scope of Cloud DPE, and likely simplifies the logic of the bot quite a bit.  

I recognize this is *a lot* more work than what's here.  I am ok with doing what's here to unblock Luke, but I also just want to admit it's not entire ideal.

JustinBeckwith (approved):

PR #983, opened 2020-09-29
feat(gcf-utils): write payload to tmp bucket

bcoe:
## Background

Cloud Tasks gives us the ability to retry failed bot operations. Observationally, this has contributed to the stability of automations:

* we've been able to identify edge-cases in logic, by observing queue depths.
* cloud functions have been known to periodically bump memory constraints, etc., leading to failures that can be retried.

## Problem

It seems like there are occasionally WebHooks that are delivered that hit the [100Kb quota of Cloud Tasks](https://cloud.google.com/tasks/docs/quotas).

## Solution

~This PR detects payloads that are over ~65,000 bytes (_I picked this number, because base64 increases byte size), and stores these payloads in a temporary bucket in Cloud Storage.~

We decided we'd start by always uploading to cloud storage.

Notes:

* ~This PR assumes that we will land the [upgrade to probot v10 first](https://github.com/googleapis/repo-automation-bots/pull/963) and forks from this branch.~
  * _I've been testing release-please against this branch, and can confirm the upgrade appears to be successful._
* There's a rule configured on the storage bucket that deletes objects daily.
 
- TODO: add a couple tests. ✅


packages/gcf-utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "gcf-utils",
-  "version": "5.7.1",
+  "version": "5.5.1",
```
JustinBeckwith, line 3: That doesn't seem like a number that is supposed to go backwards

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -233,10 +248,12 @@ export class GCFBootstrapper {
...
             payload = this.parsePubSubPayload(request);
           }
+          // TODO: find out the best way to get this type, and whether we can
+          // keep using a custom event name.
           await this.probot.receive({
-            name,
+            name: name as EventNames.StringNames,
             id,
-            payload: payload,
+            payload: await this.maybeDownloadOriginalBody(payload),
```
JustinBeckwith, line 256: From a readability perspective, I usually shy away from inline `await`s like this.  Would assignment outside of the receive call be ok?  Also a comment on why this is a "maaaaaaybe" would be nice 😆

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -306,10 +322,11 @@ export class GCFBootstrapper {
...
   }
 
-  async getInstallationToken(installationId: number) {
-    return await this.probot!.app!.getInstallationAccessToken({
-      installationId,
-    });
+  // TODO: How do we still get access to this installation token?
+  async getAuthenticatedOctokit(installationId: number): Promise<Octokit> {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const app = (this.probot as any).apps[0] as Application;
```
JustinBeckwith, line 328: Are the upstream types here just broken?  If so, do we have a tracking bug?

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -424,4 +441,57 @@ export class GCFBootstrapper {
...
+   * Storage. This allows us to circumvent the 100kb limit on Cloud Tasks.
+   *
+   * @param body
+   */
+  private async maybeWriteBodyToTmp(body: string): Promise<string> {
+    // Cloud tasks has a maximum payload size of 100kb, if a webhook payload is
+    // approaching this size, then serialize it to a temporary folder:
+    const MAX_PAYLOAD_SIZE = 65000;
+    if (process.env.WEBHOOK_TMP && body.length > MAX_PAYLOAD_SIZE) {
+      const tmp = `${Date.now()}-${v4()}.txt`;
```
JustinBeckwith, line 457: Honest question since you know more here - is `Date.now()` more safe than `uuid.v4()`?  Since there is no guarantee this is only process running at a time, I feel like having some kind of process based seed would be needed, no?

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -424,4 +441,57 @@ export class GCFBootstrapper {
...
+  private async maybeDownloadOriginalBody(payload: {
+    [key: string]: string;
+  }): Promise<object> {
+    if (payload.tmpUrl) {
+      if (!process.env.WEBHOOK_TMP) throw Error('no tmp directory configured');
+      const bucket = storage.bucket(process.env.WEBHOOK_TMP);
+      const file = bucket.file(payload.tmpUrl);
+      const readable = file.createReadStream();
+      const content = await getStream(readable);
+      await file.delete();
```
JustinBeckwith, line 490: Total nit, but maybe something we don't need to `await` and we could just let happen in a background?

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -424,4 +441,57 @@ export class GCFBootstrapper {
...
+   * Setting the process.env.WEBHOOK_TMP environment variable indicates
+   * that the webhook payload should be written to a tmp file in Cloud
+   * Storage. This allows us to circumvent the 100kb limit on Cloud Tasks.
+   *
+   * @param body
+   */
+  private async maybeWriteBodyToTmp(body: string): Promise<string> {
+    // Cloud tasks has a maximum payload size of 100kb, if a webhook payload is
+    // approaching this size, then serialize it to a temporary folder:
+    const MAX_PAYLOAD_SIZE = 65000;
```
tbpg, line 455: Can we do everything through the bucket? This adds an additional fork in the flow of data for the bots (tasks yes/no, storage yes/no). To keep it as simple as possible, should we always use a bucket?

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -424,4 +441,57 @@ export class GCFBootstrapper {
...
   }
+
+  /*
+   * Setting the process.env.WEBHOOK_TMP environment variable indicates
+   * that the webhook payload should be written to a tmp file in Cloud
+   * Storage. This allows us to circumvent the 100kb limit on Cloud Tasks.
+   *
+   * @param body
+   */
+  private async maybeWriteBodyToTmp(body: string): Promise<string> {
```
tbpg, line 452: Are there docs about how the bots work? Would you mind adding something short saying, "Payloads can/are stored in Cloud Storage. ..."?

tbpg (changes requested):

packages/gcf-utils/src/gcf-utils.ts
JustinBeckwith: To be honest I had same thought reading this ... having two data flow paths felt complex to me

packages/gcf-utils/src/gcf-utils.ts
bcoe: 👍 I will figure out a place to mention this.

packages/gcf-utils/src/gcf-utils.ts
bcoe: I've been including a `uuid`, which is guaranteed not to collide, and a timestamp.

The reason for the timestamp is so that if we ever need to perform cleanup, we can infer when the object was written by name.

packages/gcf-utils/src/gcf-utils.ts
bcoe: done.

packages/gcf-utils/src/gcf-utils.ts
JustinBeckwith: Oh gosh I read right over that 🙈 Apologies!

packages/gcf-utils/package.json
bcoe: fixed upstream.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -438,4 +448,53 @@ export class GCFBootstrapper {
...
+   * If body has the key tmpUrl, download the original body from a temporary
+   * folder in Cloud Storage.
+   *
+   * @param body
+   */
+  private async maybeDownloadOriginalBody(payload: {
+    [key: string]: string;
+  }): Promise<object> {
+    if (payload.tmpUrl) {
+      if (!process.env.WEBHOOK_TMP) throw Error('no tmp directory configured');
```
JustinBeckwith, line 489: I need to find a prettier rule for this :) Please use curly braces and new lines here.

JustinBeckwith (approved):

packages/gcf-utils/src/gcf-utils.ts
bcoe: @tbpg I created a tracking ticket, there's an architecture diagram that we simply need to update. I think we should wait until we've proven this approach out:

https://github.com/googleapis/repo-automation-bots/issues/986

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -438,4 +454,60 @@ export class GCFBootstrapper {
...
+  /*
+   * Setting the process.env.WEBHOOK_TMP environemtn variable indicates
+   * that the webhook payload should be written to a tmp file in Cloud
+   * Storage. This allows us to circumvent the 100kb limit on Cloud Tasks.
+   *
+   * @param body
+   */
+  private async maybeWriteBodyToTmp(body: string): Promise<string> {
+    if (process.env.WEBHOOK_TMP) {
+      console.info('gots in here');
```
sofisl, line 467: I think this is a comment from testing!

bcoe: nice catch 😄

packages/gcf-utils/src/gcf-utils.ts
bcoe: I actually landed on letting the bucket policy just cleanup files, because the delete makes it so we can't retry.

tbpg (approved):

PR #988, opened 2020-09-30
fix: upgrade MOG to Probot 10

sofisl:


packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -857,7 +851,7 @@ describe('merge-on-green', () => {
       ));
 
       await probot.receive({
-        name: 'pull_request.labeled',
+        name: 'pull_request.labeled' as any,
```
bcoe, line 854: I opened an issue related to this, and it seems like we should be able to work around the issue like this:

https://github.com/probot/probot/issues/1367#issuecomment-700416806

So I think hopefully the only actual issue is the `repository.scheduled`.

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -431,18 +425,14 @@ describe('merge-on-green', () => {
...
-          ],
+          name: 'Special Check',
+          conclusion: 'success',
         }),
         getCommentsOnPr([]),
       ];
 
       await probot.receive({
-        name: 'schedule.repository',
+        name: 'schedule.repository' as any,
```
bcoe, line 435: we might want to link to this:

https://github.com/octokit/webhooks.js/issues/277

to show that we're moving away from this any logic.

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -151,18 +152,10 @@ describe('merge-on-green', () => {
   const loggerStub = sandbox.stub(logger, 'error').throwsArg(0);
 
   beforeEach(() => {
-    probot = new Probot({
-      // eslint-disable-next-line node/no-extraneous-require
-      Octokit: require('@octokit/rest').Octokit,
+    probot = createProbot({
+      githubToken: 'abc123',
```
bcoe, line 156: I think in test we should get in the habit of providing an octokit constructor, like this:

https://github.com/googleapis/repo-automation-bots/blob/master/packages/release-please/test/release-please.ts#L53

Otherwise in production you might get a slightly different API surface than what probot provides by default.

You don't need to add this dependency, as it will come in by default with `gcf-util`.

bcoe (changes requested):
Left a few notes, and it might be good for us to publish the real version of the library later today.

But this is looking great.

packages/merge-on-green/src/merge-logic.ts
```diff
@@ -395,6 +397,7 @@ function cleanReviews(reviewsCompleted: Reviews[]): Reviews[] {
       }
     }
   }
+  console.log(cleanReviews);
```
bcoe, line 400: nit: I think we  could use @azizsonawalla's logger here

```
```

☝️ at the top.

```
logger.error(e);
logger.info('foo')
```

bcoe (approved):
Left a nit, it would be good to use the new logging approach. Doesn't need to be a blocker as long as we come back to this eventually

packages/merge-on-green/src/merge-logic.ts
sofisl: oops, that was a testing remnant. I can fix that, thanks!

PR #1000, opened 2020-10-02
feat: upgrade buildcop to probot 10

sofisl:


packages/buildcop/test/buildcop.ts
```diff
@@ -96,20 +100,11 @@ describe('buildcop', () => {
...
 
   beforeEach(() => {
-    probot = new Probot({
-      // use a bare instance of octokit, the default version
-      // enables retries which makes testing difficult.
-      // eslint-disable-next-line node/no-extraneous-require
-      Octokit: require('@octokit/rest').Octokit,
+    probot = createProbot({
+      githubToken: 'abc123',
+      Octokit: TestingOctokit as any,
```
tbpg, line 105: A bunch of `any` lint warnings in this file, too.

sofisl: Yeah, unfortunately the 'any' is a known issue since the Github API event names have been locked down; see here: https://github.com/octokit/webhooks.js/issues/277

packages/buildcop/src/buildcop.ts
```diff
@@ -75,6 +79,139 @@ const GROUPED_MESSAGE = `Many tests failed at the same time in this package.
 
 `;
 
+interface IssuesListForRepoResponseItem {
```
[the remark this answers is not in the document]
sofisl: Unfortunately, @octokit/types only defines the response to this method as an array, rather than a single item. We need to define this as well for those references.

Screen Shot 2020-10-14 at 12 29 40 AM

packages/buildcop/src/buildcop.ts
```diff
@@ -154,7 +291,9 @@ export function buildcop(app: Application) {
       labels: ISSUE_LABEL,
       state: 'all', // Include open and closed issues.
     });
-    let issues = await context.github.paginate(options);
+    let issues = (await context.github.paginate(
+      options
+    )) as IssuesListForRepoResponseData;
```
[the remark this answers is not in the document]
sofisl: This is because its expecting a type of `Promise<PaginationResults<unknown>>` rather than the array of data that it will be.

packages/buildcop/test/buildcop.ts
sofisl: I'll add this as a comment in the file, though!

sofisl:
@tbpg Thanks for the review! I cut down on the `IssuesListForRepoResponseItem` and added in comments to describe the 'any' situation as well as the need to declare the `IssuesListForRepoResponseItem` type.

tbpg (approved):

PR #1001, opened 2020-10-02
fix: upgrade conventional-commit-lint to probot 10

sofisl:


packages/conventional-commit-lint/src/conventional-commit-lint.ts
```diff
@@ -41,19 +46,15 @@ export = (app: Application) => {
...
-    // can be found here: https://unpkg.com/@octokit/rest@16.28.3/index.d.ts
-    let commitsResponse: Octokit.Response<
-      Octokit.PullsListCommitsResponseItem[]
-    >;
+
+    let commits: PullsListCommitsResponseData;
     try {
-      commitsResponse = await context.github.pulls.listCommits(commitParams);
+      commits = (await context.github.pulls.listCommits(commitParams)).data;
+      console.log(commits);
```
bcoe, line 53: i think we can get rid of this log line?

bcoe (approved):
This looks reasonable to me 👍 this is an easy one to test, we just need to edit a title on a PR.

PR #1004, opened 2020-10-02
feat: auto-label bot can label PR language

nicoleczhu:
Update: PR is ready to be reviewed
- [x] users can now turn on/off labeling features
     - product labeling defaults to on
     - language labeling defaults to off
- [x] users can now get PRs labeled by language and: 
    - Can config languages based on project directory structure, e.g. /sr/emailservice -> "lang: python"
    - Can config custom file extension to language mappings, e.g. "filename.ts" -> "lang: js"
    - Can config label copy e.g. "language: python" instead of "lang: python"
- [x] write tests  & refactor a bit
- [x] write readme

When deploying, verify permissions that may need to be updated: 
**Permissions**
    - **contents: read [NEW!]**
    - issues: write
    - metadata: read
    - pull_requests: write
**App events**
    - issues
    - pull_request

Finally, do we also need to notify users to update their permissions/configs?


packages/auto-label/src/extensions.json
```diff
@@ -0,0 +1,178 @@
+{
```
sofisl, line 1: Do you forsee much updating to this file? Just thinking ahead, every time we've had these files they've become a pain to upkeep. Maybe there's a place you can grab this info from is dynamic?

packages/auto-label/src/auto-label.ts
```diff
@@ -17,6 +17,19 @@ import {Storage} from '@google-cloud/storage';
 import {Application, Context} from 'probot';
 import {logger} from 'gcf-utils';
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const langlabler = require('./language');
+// Default app configs if user didn't specify a .config
+const LABEL_PRODUCT_BY_DEFAULT = true;
+const LABEL_LANGUAGE_BY_DEFAULT = false;
+const DEFAULT_CONFIGS = {
```
sofisl, line 25: This may just be something to think about, perhaps no action required. Sometimes we install bots on an org-level (like autolabel), and the config file at that point dictates the configuration for all the repos. It seems like this configuration is assuming a repo-by-repo level, but perhaps you'd want to support a configuration on an org-level as well?

sofisl (approved):

packages/auto-label/src/extensions.json
nicoleczhu: Good point! I do plan on updating this file, and I wasn't able to find an api/resource that dynamically provides file extension to language mappings. 

One of the items on my future todo is to improve this mapping to the point where it can be a canonical set of extension<>language mappings that users will be happy with as a default.

Pls let me know if you know if of a better mapping source. Open to suggestions!

packages/auto-label/src/auto-label.ts
nicoleczhu: I doubled checked: by default, probot imports configs from the org if it's not available at the repo level: 
`context.config supports sharing configs between repositories. If configuration for your app is not available in the target repository, it will be loaded from the .github directory of the target organization's .github repository.`

To your point, the only place where I'm not checking org level configs is in "installation" webhook. So atm, users can disable product labeling at the repo level but not a the org level. This is bc in the installation (custom) handler, I can't call context.config, so I manually check configs at each repo level. 

We might want to enable product-label disabling-by-config at the org level as well, I'll add this in as a next todo. :) Thanks!

packages/auto-label/src/auto-label.ts
```diff
@@ -314,15 +334,67 @@ export function handler(app: Application) {
...
+    if (!driftRepos) return;
+
     for await (const repository of repositories) {
       const [owner, repo] = repository.full_name.split('/');
 
+      // Looks for a config file, breaks if user disabled product labels
+      const response = await context.github.repos.getContent({
+        owner,
+        repo,
+        path: '.github/config.yml',
```
bcoe, line 385: let's make this something specific to auto-label, I gave the bot permission to access `.github/autolabel.yaml`

bcoe (approved):

Adawntoremember (approved):

PR #1012, opened 2020-10-05
feat(failure-checker): autorelease: tagged is not considered failure for some languages

bcoe:
Reads `release-please.yml`, and for some language flavors doesn't treat an old `autorelease: tagged` label as a failure.

### DIFF
--- a/packages/failurechecker/src/failurechecker.ts
+++ b/packages/failurechecker/src/failurechecker.ts
@@ -23,12 +23,10 @@ import {IssuesListForRepoResponseData} from '@octokit/types';
 type OctokitType = InstanceType<typeof Octokit>;
 
 import {logger} from 'gcf-utils';
+
 // labels indicative of the fact that a release has not completed yet.
-const RELEASE_LABELS = [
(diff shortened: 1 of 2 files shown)


packages/failurechecker/src/failurechecker.ts
```diff
@@ -23,12 +23,10 @@ import {IssuesListForRepoResponseData} from '@octokit/types';
...
 import {logger} from 'gcf-utils';
+
 // labels indicative of the fact that a release has not completed yet.
-const RELEASE_LABELS = [
-  'autorelease: pending',
-  'autorelease: tagged',
-  'autorelease: failed',
-];
+const RELEASE_LABELS = ['autorelease: pending', 'autorelease: failed'];
+const RELEASE_TYPE_NO_PUBLISH = ['go-yoshi'];
```
sofisl, line 29: I think this solution makes sense for right now. But, if this number expands, how do we account for the scale? Could we install it on a repo-level (vs. an org level)?

sofisl (approved):

PR #1013, opened 2020-10-06
fix(deps): upgrade SRS to probot 10

sofisl:


packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -150,7 +165,7 @@ export function handler(app: Application) {
     }
   );
 
-  app.on(['schedule.repository'], async (context: Context) => {
+  app.on(['schedule.repository' as any], async (context: Context) => {
```
JustinBeckwith, line 168: Let's please treat compiler warnings as errors.  Anytime we see one of these, we should mitigate the warning.  In this case - instead of `any`, I think you cast `as '*'` :)

packages/sync-repo-settings/test/test.ts
```diff
@@ -64,7 +68,7 @@ function nockUpdateBranchProtection(
   repo: string,
   contexts: string[],
   requireUpToDateBranch: boolean,
-  requireCodeOwners = false
```
JustinBeckwith, line 67: Curious, why change this?

packages/sync-repo-settings/test/test.ts
```diff
@@ -85,7 +89,7 @@ function nockUpdateBranchProtection(
 
 async function receive(org: string, repo: string, cronOrg?: string) {
   await probot.receive({
-    name: 'schedule.repository',
+    name: 'schedule.repository' as any,
```
JustinBeckwith, line 92: Same as above, I think we can cast this to `'*'`

packages/sync-repo-settings/test/test.ts
```diff
@@ -104,18 +108,11 @@ async function receive(org: string, repo: string, cronOrg?: string) {
 
 describe('Sync repo settings', () => {
   beforeEach(() => {
-    probot = new Probot({
-      // eslint-disable-next-line node/no-extraneous-require
-      Octokit: require('@octokit/rest').Octokit,
+    probot = createProbot({
+      githubToken: 'abc123',
+      Octokit: TestingOctokit as any,
```
JustinBeckwith, line 113: Why does this need to be `any`?

JustinBeckwith (changes requested):

packages/sync-repo-settings/test/test.ts
sofisl: The test 'should update settings for a known repository' needs this property to be true for the test to pass.

packages/sync-repo-settings/test/test.ts
sofisl: For some reason, in testing, I'm not allowed to cast this to a wildcard. I'll add a link to the issue though.

JustinBeckwith (approved):

packages/sync-repo-settings/test/test.ts
sofisl: I'll be honest, I took this part from @bcoe's upgrade of release-please. But, I think this is because this way, we can test using a separate instance of Octokit, as opposed to using the one that comes with Probot.

bcoe: we might be able to make it an `instanceof Octokit`, I believe I tried this and it was unhappy though.

PR #1014, opened 2020-10-06
fix: upgrade trusted-contribution to probot 10

sofisl:


packages/trusted-contribution/package.json
```diff
@@ -27,9 +27,10 @@
     "lint": "gts check"
   },
   "dependencies": {
-    "gcf-utils": "^5.3.0"
+    "gcf-utils": "^6.1.1"
   },
   "devDependencies": {
+    "@probot/octokit-plugin-config": "^1.0.0",
```
JustinBeckwith, line 33: I'm curious - are we 100% sure this is a dev dependency and not required as a regular dependency?

bcoe: This plugin is added in `gcf-utils`, so exists on the octokit instances passed to functions in production.

We could potentially expose the Octokit instance uses in `gcf-utils`.

JustinBeckwith (approved):

PR #1016, opened 2020-10-07
fix(failurechecker): use the logger!

JustinBeckwith:


packages/failurechecker/src/failurechecker.ts
```diff
@@ -43,14 +43,14 @@ interface ConfigurationOptions {
 }
 
 export function failureChecker(app: Application) {
-  app.on(['schedule.repository' as any], async context => {
+  app.on('schedule.repository' as '*', async context => {
```
tbpg, line 46: This change seems unrelated to logging?

tbpg (changes requested):

packages/failurechecker/src/failurechecker.ts
JustinBeckwith: This was just a little bit of type gymnastics to prevent eslint warnings where we didn't add 'ignore no-any' comments elsewhere.  It has no material change on the underlying code.  Happy to remove if it's problematic

tbpg (approved):

Issue #974, opened 2020-09-24
bug(merge-on-green): clean up errors


sofisl:
1. ensure that errors that should be bubbled up are not being caught.
2. distinguish between quota issues and server/validation errors.
3. (somewhat related): make sure rate limit is being checked before each PR, not before the entire list

sofisl:
This issue hasn't been directly addressed, but I think #1023 made some of these issues irrelevant by a large refactor. Feel free to reopen if we still face rate limiting issues.

Issue #918, opened 2020-08-27
feature request(mog): Help MOG not hit rate limits


sofisl:
MOG hits a lot of rate limits. Might be good to try to use something like etags:
https://developer.github.com/v3/

JustinBeckwith:
I took a somewhat closer look here.  Good news - there are some really low hanging watermelons 🍉   I saw we were getting pinged for abuse on the endpoint that gets PR info.  We also subscribe to PR events though!  One easy win would be using the `app.on('pull_request')`  we already have, and let that populate a table in the Datastore on every event.  We would also need to look for cases where PRs are closed, and delete them from the table.  Then, in the places where you would otherwise query for shallow data on the PR (I saw many), you can just use the Datastore table instead of reaching out to the API.  Happy to pair for an hour or two to try this approach out!

sofisl:
@JustinBeckwith, thanks for digging into this! The main reason we re-request the PR info is to make sure the PR has _not_ been closed since someone tagged the PR, or to check on statuses, etc.

If I'm understanding you correctly, maybe we could listen to PR events, and update our main Datastore table whenever these things change. I think that could work!

chingor13:
I think the problem here is that if we do hit rate limits, we aren't detecting it and we're removing the PR. Removing the PR does 2 things: remove it from the database and then remove the label from the PR. We're not removing the label so we have no signal that merge-on-green has stopped watching the pull request.

chingor13:
I'm going to change this to a bug - this is causing us a lot of pain to the point where I can trust bulk tagging automerge because I have to babysit the remaining PRs that still have the automerge label that won't be checked.

sofisl:
@chingor13, I want to let you know that I have this on my radar. I want to upgrade the bots before we continue making changes, since MOG is still using v16 of the octokit library. If I have to add any new methods, I'd rather not have to undo them after the upgrade.

Issue #882, opened 2020-08-20
feat: have MoG give some sort of indication it's watching a pull request


tbpg:
**Is your feature request related to a problem? Please describe.**

Sometimes, I'm not sure if MoG is watching a pull request. Did I add the right label (`automerge: exact`?)? A PR is green, but it hasn't been merged yet -- is MoG watching it and I just caught it at a bad time?

 **Describe the solution you'd like**

An indication the PR is being watched by MoG. For example, a :eyes: reaction on the PR.

 **Describe alternatives you've considered**

* Do nothing (works well as-is).
* A comment (too noisy).

 **Additional context**

I've thought this before, but this came up again because I wasn't 100% sure about `automerge: exact` in https://github.com/GoogleCloudPlatform/golang-samples/pull/1661.

tbpg:
Fixed by #1023. I noticed :eyes: on a PR! Thank you!

Issue #673, opened 2020-07-10
merge on green: store installation ID when merge on green requested


bcoe:
If we stored the installation ID in our tracking table when the merge on green label is applied, we would be able to use this identifier to create an instance of `@octokit/rest` authenticated to the given installation, rather than relying on a scheduled task for each installation.

If we eventually get other teams at Google interested in using merge on green, this would be a good way to approach things.

PR #1023, opened 2020-10-08
fix(merge-on-green): implement solution for quota issues

sofisl:
This PR aims at restructuring the information that is held in the Datastore table for MOG in order to decrease the requests made to the API. The following changes have been implemented:

- MOG now listens for PRs closed, merged, or unlabeled, to potentially remove the PRs from the Datastore table
- MOG now reacts to PRs with an 'eyes' emoji to let the user know it's watching the PR
- MOG saves the following pieces of information in the Datastore table: 
       - the author of the PR
       - the ID of the reaction MOG places on the PR
       - the label the PR is using
       - the installation ID of the bot so that bots can authenticate independently on each run (this part hasn't been implemented yet, right now we're just storing the installation ID)

Should address #918, third point on #974, #673, #882


packages/merge-on-green/src/merge-logic.ts
```diff
@@ -609,32 +542,29 @@ export async function mergeOnGreen(
...
-    await getLatestCommit(owner, repo, pr, github),
-  ]);
-  logger.info(`required checks: ${requiredChecks}`);
-
-  if (prInfo.state === 'closed') {
-    logger.info(`${owner}/${repo}/${pr} is closed`);
-    return true;
+  const rateLimit = (await github.rateLimit.get()).data.resources.core
+    .remaining;
+  if (rateLimit <= 0) {
```
chingor13, line 551: Do we know how quickly quota refills? Should we wait until we have a certain threshold of quota available?

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -164,11 +203,21 @@ function handler(app: Application) {
...
+              await handler.removeLabelAndReaction(
+                wp.owner,
+                wp.repo,
+                wp.number,
+                wp.label,
+                wp.reactionId,
+                github as any
+              );
             }
+            await handler.removePR(wp.url);
```
chingor13, line 220: Should this be in the `if(remove || wp.state === 'stop')` block?

chingor13 (commented):
Overall, I'd probably split out adding the watching reaction into a separate feature PR as it's unrelated to the quota issues.

I'm not fully understanding how this is fixing the quota issues.

sofisl:
@chingor13, this should help quota issues because we're no longer making requests for the PR at the beginning of the execution: https://developer.github.com/v3/pulls/#get-a-pull-request

Instead, we're storing most of the information in the Datastore table when we initially store the PR. We will eventually request this PR information to get info like `mergeable`, but we'll do this once tests and reviews have passed (which will hopefully reduce API requests). We also won't be checking to see if the label is on the PR itself, we'll update whether the PR exists in the Datastore table based on the `label` and `unlabel` events.

As to the reaction piece - I hear you. I *don't* think the reaction is breaking the test execution currently, it should be a pretty innocuous addition to the PR. I'm leaning towards keeping it just because it should be a relatively minor addition, but I'm happy to split this up into two PRs if you feel strongly. What's currently breaking the PR is using the installation ID. I can make that into its own PR.

packages/merge-on-green/src/merge-on-green.ts
sofisl: AWESOME catch! thanks!

packages/merge-on-green/src/merge-logic.ts
sofisl: It should take an hour to refill! I think if we just keep the PR in the Datastore table, we can just check back in an hour to see if the rate has been replenished (I might be misunderstanding your suggestion here though, feel free to clarify).

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -246,11 +309,58 @@ function handler(app: Application) {
...
+      )?.name;
+
+      // If the label is on the PR but the action was unlabeled, it means the PR had some other
+      // label removed. No action needs to be taken.
+      if (label && context.payload.action === 'unlabeled') {
+        logger.info(
+          `correct label ${label} is still on ${repo}/${prNumber}, will continue watching`
+        );
+        return;
+      }
```
chingor13, line 343: This seems to only matter for the pull_request.unlabeled event. Let's split into it's own handler block.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -246,11 +309,58 @@ function handler(app: Application) {
...
+
+      // Check to see if the PR exists in the table before trying to delete. We also
+      // need to do this to get the reaction id to remove the reaction when MOG is finished.
+      const PR: WatchPR = await handler.getPR(
```
chingor13, line 347: looks like we're using lower camel-case for variables. we should also try to avoid abbreviations if possible.
```suggestion
      const watchedPullRequest: WatchPR = await handler.getPR(
```

packages/merge-on-green/src/merge-logic.ts
chingor13: My question is more around if the quota trickles in -- say we are out of quota, then are given back 3 API calls, but a single m-o-g process takes 5 API calls. We won't be able to complete the m-o-g process and we'll exhaust our quota again.

chingor13 (changes requested):

packages/merge-on-green/src/merge-logic.ts
sofisl: per our conversation, making it ~10 (that's about the API calls it would take to finish)

chingor13 (approved):

PR #1024, opened 2020-10-09
fix: minor clean ups relating to #1023

sofisl:


packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -241,7 +242,7 @@ function handler(app: Application) {
       (label: Label) =>
         label.name === MERGE_ON_GREEN_LABEL ||
         label.name === MERGE_ON_GREEN_LABEL_SECURE
-    );
+    )?.name;
```
chingor13, line 245: Alternatively, you're ensuring label is present on the next line. You could confidently use `label.name` after that check.

chingor13 (approved):

PR #1039, opened 2020-10-16
feat: extended auto-label to accept user configs & label based on file directories

### LINKED ISSUE #978: auto label bot should label any PR that touches files in a samples folder with the samples label
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Python sample owners receive a lot of PRs from the library, but only need to pay attention when they have to do with samples. If possible, it would be super if we could label the PRs with the `samples` label to make it easier. We are also considering using blunderbuss to properly assign PRs, but I see blunderbuss as a solution that would complement this, not necessarily be an alternative.

nicoleczhu:
Fixes #978 🦕

Additions: 
- Users can now just turn on the path labeling feature without language labeling (i.e. directory based labeling can be use independently now)
- Language labeling doesn't default to a "lang: __" prefix anymore
- Test coverage for the above


packages/auto-label/src/auto-label.ts
```diff
@@ -363,18 +355,47 @@ export function handler(app: Application) {
...
-    if (language) {
-      logger.info('Labeling PR with: ' + language);
-      await context.github.issues.addLabels({
-        owner,
-        repo,
-        issue_number: pull_number,
-        labels: [language],
-      });
+
+    if (config.path && config.path.pullrequest) {
```
sofisl, line 359: Would you mind if we set comments above some of the 'if' statements just to explain the paths?

sofisl:
@nicoleczhu, I think this all makes sense, although it's a little difficult to follow the logic. Do you have access to the GCP account to monitor logs when it's deployed? It would be nice to test out the new functionality!

packages/auto-label/src/auto-label.ts
nicoleczhu: That's good point. I'll add more comments to this PR, since the various configs are now getting hard to follow

nicoleczhu:
> @nicoleczhu, I think this all makes sense, although it's a little difficult to follow the logic. Do you have access to the GCP account to monitor logs when it's deployed? It would be nice to test out the new functionality!

I'll add more comments to this PR, since the various configs are now getting hard to follow. 

I do have access to monitoring and will watch it if/when this PR is merged in the future!

sofisl:
Thank you! This looks awesome!

I'm happy to merge this, but just looking through it, I'm wondering if we should (for future work) consider consolidating the language- and product-labeling workflows a bit more, since they seem a bit disparate right now (although maybe that's for the best). For example, you have a helper function to check to see if the label is already on a PR, and the product-labeling code does that as well.  I'm not super worried about it given that the language labeling seems to only apply to PRs, and the product labeling seems to apply to issues. But, I guess it might be worth considering if you want to expand the language-labeling part of auto-label to do more? For example, we could add it to the bit of code that runs on a cron job to backfill repos with labels? Perhaps you're interested in having it respond to issues?

These are more just thoughts in case we consider having both of the workflows overlap a bit more, since I think they do similar functionalities, just for different names. We can take it day by day and see if you're interested in increasing the language-labeling scope!

nicoleczhu:
> Thank you! This looks awesome!
> 
> I'm happy to merge this, but just looking through it, I'm wondering if we should (for future work) consider consolidating the language- and product-labeling workflows a bit more, since they seem a bit disparate right now (although maybe that's for the best). For example, you have a helper function to check to see if the label is already on a PR, and the product-labeling code does that as well. I'm not super worried about it given that the language labeling seems to only apply to PRs, and the product labeling seems to apply to issues. But, I guess it might be worth considering if you want to expand the language-labeling part of auto-label to do more? For example, we could add it to the bit of code that runs on a cron job to backfill repos with labels? Perhaps you're interested in having it respond to issues?
> 
> These are more just thoughts in case we consider having both of the workflows overlap a bit more, since I think they do similar functionalities, just for different names. We can take it day by day and see if you're interested in increasing the language-labeling scope!

@sofisl definitely! I created an issue here #1045 to track all of these work items. Can you assign me to it? I don't have permissions to edit the filed issue.

I definitely plan to maintain this package going forward as a 20% project. There's still work to be done on it ~~~

sofisl (approved):

nicoleczhu:
@sofisl is it possible for you to merge this one? I don't have write access to do so. Thank you!

PR #1062, opened 2020-10-28
feat(auto-label): never label with api: docs or api: build

### LINKED ISSUE #955: auto-label: distinguish between 'docs' as a type of issue and an API
Screen Shot 2020-09-14 at 10 36 52 AM

Auto-label thinks this is related to the docs API, but it's actually for the Drive docs. This is a pretty minor inconvenience to fix, but just a heads up.

tbpg:
The docs prefix is almost always for Conventional Commits, rather than
the API.

Fixes #955 🦕


nicoleczhu (approved):

packages/auto-label/src/auto-label.ts
```diff
@@ -112,6 +112,12 @@ export function autoDetectLabel(
   firstPart = firstPart.toLowerCase(); // Convert to lower case.
   firstPart = firstPart.replace(/\s/, ''); // Remove spaces.
 
+  // The Conventional Commits "docs:" prefix is far more common than "api: docs".
+  // So, never label with "api: docs".
+  if (firstPart === 'docs') {
```
JustinBeckwith, line 117: This happens with `build` too 😆

JustinBeckwith (approved):

bcoe (approved):

packages/auto-label/src/auto-label.ts
bcoe: I agree it might be worth adding `build` too.

tbpg: Done! I also updated the logic to short-circuit for titles that start with `docs:` or `build:`.

PR #1075, opened 2020-11-03
chore(deps): update gts  to 3.0.0

sofisl:
Addresses #1019


package.json
```diff
@@ -17,5 +17,8 @@
     "smee-client": "^1.1.0",
     "@microsoft/api-documenter": "^7.8.10",
     "@microsoft/api-extractor": "^7.8.10"
+  },
+  "dependencies": {
```
JustinBeckwith, line 21: Not sure if we need this, since I don't think it gets run specifically in CI.  This should be delegated to all the sub-directories.

JustinBeckwith (approved):

PR #1099, opened 2020-11-10
feat(snippet-bot): check if region tags have product prefix

### LINKED ISSUE #1098: feat(snippet-bot): check if region tags have product prefix
For getting product names, maybe we can use `gs://devrel-prod-settings/apis.json`. I'm still wondering how accurate this file is.

tmatsuo:
Fixes #1098


tbpg (approved):
LGTM assuming we are sure this is the right set of prefixes.

tmatsuo:
@tbpg Thanks, this file should contain all the product names and it should work.

PR #1105, opened 2020-11-11
fix(snippet-bot): use products.json for region_tag_prefix

### LINKED ISSUE #1101: bug(snippet-bot): cloudrun prefix is considered as a violation
snippet-bot started to check product prefixes. Now it is using product short names for product prefix, but in some cases, the product prefix is different from the product short names (e.g. cloudrun v.s. run).

We should use `region_tag_prefix` instead of `api_shortname`. Now we're waiting for the `api.json` to have that new field.

tmatsuo:
fixes #1101

Locally it works fine with using the dev bucket. Currently the new `products.json` file is not yet deployed to the prod bucket, so we can not merge this yet.

### DIFF
--- a/packages/snippet-bot/src/violations.ts
+++ b/packages/snippet-bot/src/violations.ts
@@ -24,18 +24,24 @@ export interface Violation {
   violationType: violationTypes;
 }
 
+let dataBucket: string | undefined = process.env['DEVREL_SETTINGS_BUCKET'];
+if (dataBucket === undefined) {
+  // Default prod bucket
+  dataBucket = 'devrel-prod-settings';
+}
+
 export const checkProductPrefixViolations = async (
   changes: ChangesInPullRequest
 ): Promise<Array<Violation>> => {
   const ret: Violation[] = [];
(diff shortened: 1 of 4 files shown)


packages/snippet-bot/src/violations.ts
```diff
@@ -24,18 +24,24 @@ export interface Violation {
   violationType: violationTypes;
 }
 
+let dataBucket: string | undefined = process.env['DEVREL_SETTINGS_BUCKET'];
```
JustinBeckwith, line 27: This could be simplified to `const dataBucket = process.env.DEVREL_SETTINGS_BUCKET || 'devrel-prod-settings'`.

JustinBeckwith (approved):

packages/snippet-bot/src/violations.ts
tmatsuo: Done

tmatsuo:
Now the `products.json` file is in the prod bucket. Merging.

Issue #1088, opened 2020-11-06
merge on green: 'not-mergeable' comment left on merged PR


busunkim96:
https://github.com/googleapis/python-billingbudgets/pull/53

Yesterday (11/04/2020)
- I opened the above PR.
- I added (and then removed) the `automerge` label.
- I merged the PR.

Today (11/05/2020)
- About 24 hours after the merge, gcf-merge-on-green commented with the below:

> Merge-on-green attempted to merge your PR for 6 hours, but it was not mergeable because either one of your required status checks failed, or one of your required reviews was not approved. Learn more about your required status checks here: https://help.github.com/en/github/administering-a-repository/enabling-required-status-checks. You can remove and reapply the label to re-run the bot.

There is no impact to my workflow, but I wanted to report this in case it was indicative of some other issue.

busunkim96:
Similar behavior obsereved https://github.com/GoogleCloudPlatform/python-docs-samples/pull/4946. On this PR the bot is commenting much more frequently.

bcoe:
@sofisl if you're so inclined, please feel free to make a case for this being `p2` 😄

sofisl:
@busunkim96, thanks for bringing this up! I will look into this ASAP.

busunkim96:
Closed by #1091. Thanks @sofisl!

PR #1112, opened 2020-11-12
fix(merge-on-green): add cleanup cron job, fix blocking issue for removing

sofisl:
See [java-bigtable/498](https://github.com/googleapis/java-bigtable/pull/498), and [this issue](https://github.com/googleapis/repo-automation-bots/issues/1088).

I did some further digging into this multiple commenting issue, that did not seem to be solved by #1091. I think the real issue here is when people add an 'automerge' label and then either remove that label, close, or merge that PR soon after. I think a race condition is happening where the PR is not yet recorded in the Datastore table when the PR is closed or removed, making so that it doesn't actually get removed. 

I don't know how to solve this problem categorically. What I ended up doing is creating a clean-up cron job that runs every 30 minutes. Every 30 minutes, if the cron job is the 'cleanup' job, the app will go through the PRs in the Datastore table and clean up ones that are either closed, merged, or do not have an automerge label on them anymore.


packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -148,6 +148,59 @@ handler.cleanUpPullRequest = async function cleanUpPullRequest(
     .catch(logger.error);
 };
 
+/**
+ * Check if PR has been merged, closed, or unlabeled, then remove
```
tbpg, line 152: Missing end of sentence?

tbpg (changes requested):
Design question: could we do this cleanup "check" before doing anything with a PR in the normal loop/cron? That way, there won't be a delay between the problematic state and when it gets fixed by the next cleanup cron iteration.

Also, I saw the multiple-comment behavior on https://github.com/googleapis/repo-automation-bots/pull/1111, which doesn't seem like the same issue?

sofisl:
Thanks for the review @tbpg!
> Design question: could we do this cleanup "check" before doing anything with a PR in the normal loop/cron? That way, there won't be a delay between the problematic state and when it gets fixed by the next cleanup cron iteration.

I think what you're asking is if we can check to see if the PR is closed/merged/unlabeled before doing something in the regular workflow. This was the previous implementation of the bot, but I changed it in #1023 because checking this info was causing us to run into rate limits very quickly. Now, we wait for the unlabeled/closed/merged event to update the Datastore table. But, now we're running into a separate issue on whether or not the PR is in the table before it gets deleted if it gets added and deleted in the table too quickly.

But, maybe we run the cleanup cron job every 15 minutes, instead of every 30? It is an expensive task to do, but it might be worth to address these issues. I'll leave that suggestion up to you and/or anyone else who has thoughts on it :)

> Also, I saw the multiple-comment behavior on #1111, which doesn't seem like the same issue?

Funny you mention that PR, it is actually the same issue! I was testing it last night. I quickly added and removed the automerge label. 

Screen Shot 2020-11-12 at 12 19 47 PM

If you see there, MOG actually ended up merging that PR, even though it didn't have an automerge label on it. I am not exactly clear on what causes the multiple commenting issue, but I do know that this fix should solve it, since that PR shouldn't have even been in the Datastore table to begin with.

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -26,7 +26,7 @@ import Ajv from 'ajv';
 import yaml from 'js-yaml';
 import {PullsListFilesResponseData} from '@octokit/types';
 
-export const configFileName = 'sync-repo-settings.yaml';
+export const configFileName = 'sync-repo-settings.yml';
```
tmatsuo, line 29: There seem to be some changes unrelated?

sofisl: oof, thanks for the heads up. Need to resync to main branch.

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -340,12 +340,6 @@ async function updateRepoTeams(
   logger.info(`Update team access for ${repo}`);
   const [owner, name] = repo.split('/');
```
tmatsuo, line 342: Is this also an irrelevant change?

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -798,6 +810,93 @@ describe('merge-on-green', () => {
...
+          name: 'schedule.repository' as any,
+          payload: {org: 'testOwner', cleanUp: true},
+          id: 'abc123',
+        });
+
+        scopes.forEach(s => s.done());
+        assert(removePRStub.called);
+      });
+
+      it('deletes a PR if label is not found when cleaning up repository', async () => {
```
bcoe, line 868: awesome to see the increased coverage 👍

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -210,15 +297,19 @@ function handler(app: Application) {
               github as any
             );
             if (remove || wp.state === 'stop') {
+              await handler.removePR(wp.url);
+              try {
```
bcoe, line 301: spacing looks a little weird here, might need to run `npm run fix`.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -188,6 +252,29 @@ function handler(app: Application) {
   //open issue for a fix: https://github.com/octokit/webhooks.js/issues/277
   app.on('schedule.repository' as any, async context => {
     const watchedPRs = await handler.listPRs();
+    if (context.payload.cleanUp === true) {
+      while (watchedPRs.length) {
```
bcoe, line 256: I might be tempted to pull this into a helper function.

bcoe (changes requested):
The logic in `schedule.repository` is getting pretty expansive, perhaps we could split it into two helpers that get called depending on the type of payload:

`runCleanUp`/`runMergeOnGreen` or something like that?

tmatsuo:
This is kind of meta comment. Take this a grain of solt.

The PR title should be something like:
```
fix(merge-on-green): add cleanup cron job
```
1. Maybe we should use package name in the paren
2. More concrete title is better in my opinion

bcoe (approved):

tbpg (approved):
A few small comments. Otherwise, LGTM.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -135,17 +135,82 @@ handler.cleanUpPullRequest = async function cleanUpPullRequest(
...
+  try {
+    pr = (
+      await github.pulls.get({
+        owner,
+        repo,
+        pull_number: prNumber,
+      })
+    ).data;
+  } catch (err) {
+    pr = undefined;
```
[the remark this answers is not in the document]
sofisl: I might leave this just in case the github API is unavailable for some reason, but we can confirm by checking to see if a label is on the PR in the next function.

PR #1116, opened 2020-11-13
feat(snippet-bot): give warnings for removing region tags in use

### LINKED ISSUE #1115: feat(snippet-bot): give warnings for removing region tags in use

tmatsuo:
fixes #1115


packages/snippet-bot/test/snippet-bot.ts
```diff
@@ -83,6 +86,43 @@ describe('snippet-bot', () => {
...
+            branch: 'master',
+          },
+        ],
+      });
+      testSnippets.set('datastore_named_key', {
+        title: '',
+        description: '',
+        languages: languageMap2,
+      });
+      getSnippetsStub = sandbox.stub(snippetsModule, 'getSnippets');
```
bcoe, line 124: mind adding an assertion that specifically describes the new behavior, e.g.,

"should add warning when this condition is met".

tmatsuo: Sure, done

tmatsuo:
@bcoe Thanks, PTAL

tmatsuo:
Rebased.
The build failed with the following error:

```
     Uncaught Error: Cannot store new snapshot value
in "/home/runner/work/repo-automation-bots/repo-automation-bots/packages/snippet-bot/build/test/snippet-bot.js"
for snapshot called "snippet-bot responds to PR gives warnings about removing region tag in use"
test key "snippet-bot responds to PR gives warnings about removing region tag in use 1"
when running on CI (opts.ci = 1)
```

tmatsuo:
The build failure should be fixed by the new snapshot (I fixed the typo in the test description so that I needed to update the snapshot :)

packages/snippet-bot/test/snippet-bot.ts
```diff
@@ -406,6 +412,102 @@ describe('snippet-bot', () => {
 
       requests.done();
     });
+
+    it('gives warnings about removing region tag in use', async () => {
```
sofisl, line 416: I wonder if it would be useful to put another type of test as well asserting the opposite, i.e., doesn't warn if region tag isn't being removed or isn't in use, etc.

tmatsuo: This is tested by other assertion :)

bcoe (approved):
Thanks for the test, I'm a 👍 as soon if @sofisl's concerns are addressed.

packages/snippet-bot/test/snippet-bot.ts
tmatsuo: Maybe I can add an explicit description for that :)

packages/snippet-bot/src/snippets.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export interface SnippetLanguage {
+  status: SnippetStatus;
+  current_locations: Array<SnippetLocation>;
+}
+
+export interface Snippet {
+  title: string;
+  description: string;
+  languages: Map<string, SnippetLanguage>;
```
JustinBeckwith, line 61: Generally you'll see things like typed as `{ [index: string]: SnippetLanguage }`.  When you do a `JSON.parse`, it will not provide a `Map` object like you would expect.

packages/snippet-bot/src/snippets.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export const getSnippets = async (
+  dataBucket: string
+): Promise<Map<string, Snippet>> => {
+  const snippets = await storage
+    .bucket(dataBucket)
+    .file('snippets.json')
+    .download();
+  const parsedResponse = JSON.parse(snippets[0].toString());
```
JustinBeckwith, line 71: Yeah I'm finding the mapping code here somewhat confusing, and I suspect you want to represent the types using the method I mentioned above instead of remapping it.

packages/snippet-bot/src/violations.ts
```diff
@@ -27,6 +33,79 @@ export interface Violation {
...
+export const checkRemovingUsedTagViolations = async (
+  changes: ChangesInPullRequest,
+  config: Configuration,
+  parseResults: Map<string, ParseResult>,
+  baseRepositoryPath: string, // googleapis/java-accessapproval
+  baseBranch: string
+): Promise<Map<string, Array<Violation>>> => {
+  const removeUsedTagViolations = new Array<Violation>();
+  const removeConflictingTagViolations = new Array<Violation>();
+  const snippets = await getSnippets(dataBucket as string);
```
JustinBeckwith, line 45: You shouldn't have to cast this to string - what type does the compiler think it has?

packages/snippet-bot/test/snippet-bot.ts
tmatsuo: Ok, I changed the test description. @sofisl PTAL

packages/snippet-bot/src/snippets.ts
tmatsuo: Oh nice, I'll try that. Thanks.

packages/snippet-bot/src/snippets.ts
tmatsuo: Agreed

packages/snippet-bot/src/violations.ts
tmatsuo: It was a leftover because I declared it as `string | undefined`, so now it's not needed.

tmatsuo:
@JustinBeckwith @sofisl Thanks, PTAL

JustinBeckwith (approved):
Much better.  Thank you!

sofisl (approved):

PR #1120, opened 2020-11-15
chore(auto-label): clean up types

JustinBeckwith:


packages/auto-label/src/auto-label.ts
```diff
@@ -31,8 +52,7 @@ const DEFAULT_CONFIGS = {
   },
 };
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const colorsData = require('./colors.json');
+import colorsData from './colors.json';
```
tbpg, line 55: Move to be next to the other imports at the top? Or is that not a thing?

packages/auto-label/package.json
```diff
@@ -18,7 +18,7 @@
     "probot-app"
   ],
   "scripts": {
-    "compile": "tsc -p . && cp ./src/colors.json ./build/src/colors.json && cp ./src/extensions.json ./build/src/extensions.json",
+    "compile": "tsc",
```
tbpg, line 21: Intentional update?

tbpg (approved):

packages/auto-label/src/auto-label.ts
JustinBeckwith: We could, just didn't move it since it was already here

packages/auto-label/package.json
JustinBeckwith: Yes!  I updated the `tsconfig.json` to include JSON in it's compilation, allowing us to do the import style require like I did above.  This has the benefit of also copying the JSON files imported in source automatically into the `build/src` folder, negating the need for the `cp` commands.  Also `-p .` is the default behavior of `tsc`, so you know, saving some bits here and there 😛

PR #1124, opened 2020-11-15
fix(blunderbuss): handle assign_issues_by only config

### LINKED ISSUE #1123: Investigate "usernames is not iterable" from the logs
We're seeing the following error in the logs:
```
TypeError: usernames is not iterable
at expandTeams (/workspace/build/src/blunderbuss.js:190)
at (/workspace/build/src/blunderbuss.js:160)
at runMicrotasks (<anonymous>)
```

JustinBeckwith:
Fixes #1123 - dug through the logs, and it was pointing towards a run where this issue was opened:
https://github.com/googleapis/google-cloud-dotnet/issues/5576

The repo uses the config here:
https://github.com/googleapis/google-cloud-dotnet/blob/master/.github/blunderbuss.yml

There was a bit of code where I think `undefined` was getting passed around, because we thought we were smarter than the TypeScript compiler (we weren't)


packages/blunderbuss/test/fixtures/config/assign_by.yml
```diff
@@ -0,0 +1,5 @@
+assign_issues_by:
+- labels:
+  - 'api: logging'
+  to:
+  - tormund
```
tbpg, line 5: This is a real user, so it might be good to use another username. :dog:

tbpg (approved):

PR #1131, opened 2020-11-16
feat(gcf-utils): change the method signature of addOrUpdateIssueComment

### LINKED ISSUE #1130: make `addOrUpdateIssueComment` more accesible
Now it requires `probot.Context`, but it only needs Octokit.

tmatsuo:
fixes #1130 

Now it only requires Octokit, instead of Context.
The intention is to allow us to use this function from the cron hander.


tmatsuo:
Rebased. This is technically a breaking change, but this function is just added and we haven't released gcf-utils yet, so it should be fine to merge in in my opinion.

bcoe (approved):

PR #1141, opened 2020-11-16
feat(snippet-bot): cache Snippets data

### LINKED ISSUE #1140: feat(snippet-bot): cache Snippets
The json file for Snippets is somewhat big, so it might be worth caching. When the bot is triggered by `snippet-bot:force-run`, it should invalidate the cache.

tmatsuo:
fixes #1140


packages/snippet-bot/src/snippets.ts
```diff
@@ -65,11 +67,41 @@ export interface Snippets {
   [index: string]: Snippet;
 }
 
+let cachedSnippets: Snippets;
+let cacheTimestamp = 0;
+const lock = new AwaitLock();
```
tbpg, line 72: Why do we need the lock? A single Cloud Function instance only processes a single request at a time.

tbpg (commented):
One question, otherwise LGTM.

packages/snippet-bot/src/snippets.ts
tmatsuo: Now it's fine, but we may have another async function that uses `getSnippets` in the future.

tmatsuo: or, we may want to run the code in parallel.

bcoe: As long as we're always awaiting `getSnippets`, we shouldn't need a lock, because Node.js is [single threaded](https://nodejs.dev/learn/the-nodejs-event-loop), and a single cloud function (at this time) should only be handling one request at a time.

There's a chance in the future cloud functions support multiple concurrent requests, but in practice I've still found it's really rare to have contention on a resource, since these multiple requests will still be bottlenecked on a single processing thread -- basically, all the concurrency happens at the operating system level using [libuv](https://github.com/libuv/libuv), but a single process only ever has one thread.

JustinBeckwith: I don't think this will work the way you want it to :( Why Cloud Functions are single threaded in a sense, there can be multiple instances of a Cloud Function running at the same time.  If we truly need a global lock here (I very much think we should avoid that), we need a database that natively supports that feature.  Uh.... @chingor13 should probably be involved here.

JustinBeckwith (changes requested):
We need to have a broader discussion about the implication of a global lock with @bcoe and @chingor13

packages/snippet-bot/src/snippets.ts
bcoe: If the goal is caching between all cloud function invocations, redis is an incredible tool for the job, and has incredible libraries for Node.js.

For locks like this, I always like to use a lock with a TTL, so that you never end up with deadlock. This is built in to all of redis' data structures.

tmatsuo: The goal is caching the data in the single cloud function instance.

Currently, subsequent requests to the same instance will benefit from the cache.

However, if we want to introduce concurrent execution in the bot logic, I still think we need a lock.

tmatsuo: Re: deadlock

Doesn't it suffice to have the `finally` block? If `finally` ensures the code execution, we won't have deadlock.

tmatsuo: Maybe I can remove the usage of the lock with a consensus of us running this app on Cloud Function in a foreseeable future.

However, how about the other platform? If there are platforms that supports multiple threads, or concurrent requests, then it's safe to use the lock now.

What's the downside of using a lock besides a slight performance penalty?

bcoe: > What's the downside of using a lock besides a slight performance penalty?

In a single Node.js process, I think it just adds more complexity than is needed. Even with multiple requests hitting the same process, there's no danger that there will be contention on a resource (due to the single event loop model). So, the only thing this would potentially protect against is if two requests hit the cache step at an identical time, in which case one will take precedence.

tmatsuo:
Removed the lock usage. PTAL

tbpg (approved):

JustinBeckwith (approved):
Thank you :)

PR #1144, opened 2020-11-17
cleanup(snippet-bot): reduce memory footprint for large diffs

### LINKED ISSUE #1143: bug(snippet-bot): reduce memory footprint for large diffs
Related to #1134 

Before we increase the memory limit to 1GB from 500MB, the bot died for 2 PRs with large diffs (75MB and 180MB). The bot create a string in an early stage, then passes that large string to `parseRegionTagsInPullRequest`. This creates adiitional copy of the string.

I think we should reduce at least one copy of this string instance.

tmatsuo:
fixes #1143

### DIFF
--- a/packages/snippet-bot/src/region-tag-parser.ts
+++ b/packages/snippet-bot/src/region-tag-parser.ts
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import parseDiff from 'parse-diff';
-import * as nodeFetch from 'node-fetch';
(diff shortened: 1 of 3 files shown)


tmatsuo (commented):
Well, I think we should move the fetch part to that function too ...

tmatsuo:
Now it's ready! PTAL

bcoe (approved):

PR #1146, opened 2020-11-17
fix(snippet-bot): disable warning for removal of region tags in use

### LINKED ISSUE #1145: bug(snippet-bot): disable warnings for removal of region tags in use
It turned out that the data we have now is not the right one. We should disable this feature for now.

tmatsuo:
fixes #1145


packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -425,20 +421,9 @@ ${bodyDetail}`
         result,
         configuration
       );
-      // Check for removing region tags in use.
-      const removeViolations = await checkRemovingUsedTagViolations(
```
JustinBeckwith, line 429: It makes me slightly nervous that we're able to delete this code with no test changes :)  LGTM assuming CI is passing.  Can we get an issue in the tracker that outlines what's broken, and reminds us to add this back though?

JustinBeckwith (approved):

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: Ya, there was no `calledOnce` type assert ;)

Will open a new issue about enabling this.

tmatsuo: https://github.com/googleapis/repo-automation-bots/issues/1147

Issue #1148, opened 2020-11-17
cleanup(gcf-utils): use `new Probot(options)` instead of `createProbot(options)`


tmatsuo:
I see the following warnings:

``` 
W2020-11-17T22:31:47.688Zsnippet_botjboarmlhl8bc Deprecation: [probot] "createProbot(options)" is deprecated, use "new Probot(options)" instead
2020-11-17T22:31:47.688Zsnippet_botjboarmlhl8bc at Object.exports.createProbot (/workspace/node_modules/probot/lib/index.js:237:22)
2020-11-17T22:31:47.688Zsnippet_botjboarmlhl8bc at GCFBootstrapper.loadProbot (/workspace/node_modules/gcf-utils/build/src/gcf-utils.js:56:36)
2020-11-17T22:31:47.688Zsnippet_botjboarmlhl8bc at async /workspace/node_modules/gcf-utils/build/src/gcf-utils.js:135:33ARN (Deprecation): [probot] "createProbot(options)" is deprecated, use "new Probot(options)" instead
```

I also get the following warnings:

```
2020-11-17T22:31:47.689Zsnippet_botjboarmlhl8bc Deprecation: [probot] "cert" option is deprecated. Use "privateKey" instead
2020-11-17T22:31:47.689Zsnippet_botjboarmlhl8bc at new Probot (/workspace/node_modules/probot/lib/index.js:45:27)
2020-11-17T22:31:47.689Zsnippet_botjboarmlhl8bc at Object.exports.createProbot (/workspace/node_modules/probot/lib/index.js:238:12)
2020-11-17T22:31:47.689Zsnippet_botjboarmlhl8bc at GCFBootstrapper.loadProbot (/workspace/node_modules/gcf-utils/build/src/gcf-utils.js:56:36)
2020-11-17T22:31:47.689Zsnippet_botjboarmlhl8bc at async /workspace/node_modules/gcf-utils/build/src/gcf-utils.js:135:33```

PR #1149, opened 2020-11-17
cleanup(gcf-utils): clean up deprecations

tmatsuo:
related to #1148

I think we'll need to update the secret too.


bcoe:
as you say, let's coordinate updating the secret in secret manager; perhaps we can test both upgrading to the latest version of probot (which we've been blocked on) and dropping these deprecation warnings.

SurferJeffAtGoogle:
This PR has seen no activity for 90+ days.

PR #1150, opened 2020-11-18
chore(data-processor): use @google-cloud/functions

JustinBeckwith:
This swaps out the 3rd party cloud functions module for our delightful new in-house one.  @chingor13 can I trust the unit tests here, or should I do some further testing?  

@bcoe I found a mistake in our CI config already :) Included that fix in here for good measure.


chingor13 (approved):

chingor13:
Types FTW here. I think we can trust the tests if we believe we can trust the types.

Issue #1160, opened 2020-11-19
feat(gcf-utils): provide a helper to display checkbox for refreshing an issue comment


tmatsuo:
It is great if we can display a checkbox to trigger the bot to refresh the issue comment.

Example:

- [ ] Refresh this comment

I'm going to prototype with snippet-bot and then create a PR on gcf-utils.

PR #1161, opened 2020-11-19
feat(snippet-bot): provide a checkbox for refreshing

tmatsuo:
a part of #1160

This feature requires a new permission:
- Listen to webhooks for Issue comment

We should merge this after setting the above permission for googleapis and googlecloudplatform.


tmatsuo:
Here is an example of the checkbox:
https://github.com/tmatsuo/python-docs-samples/pull/20

packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -450,7 +451,61 @@ To update this comment, add \`${REFRESH_LABEL}\` label.
...
+// This should synced with the body of the comment. When we move
+// `onRefreshCheckbox` to gcf-utils, we can have the single source and
+// automatically add the checkbox in `addOrUpdateIssueComment`.
+const REFRESH_STRING = '- [x] Refresh this comment';
+
+function onRefreshCheckbox(
+  app: Application,
+  callback: (context: Context) => Promise<void>
+): void {
+  app.on('issue_comment.edited', async context => {
```
tbpg, line 463: API question: I kind of like having all `app.on` calls in the main application function. That way, you can easily see what event types lead to what code being run. But, if we call `app.run` here (or in gcf-utils), this `issue_comment.edited` event is hidden. Perhaps we should lift the `app.on` call to the caller of `onRefreshCheckbox`?

This would increase the verbosity of the API, but I believe it will make it easier to understand and debug.

tbpg (commented):
This is awesome!

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: Sure, in gcf-utils, maybe we can only provide a function for creating commentMark `getCommentMark(instanceId: string)`, so that the user handler can determine if the comment is made by the bot.

The logic after that might defer bot by bot (one bot wants to "refresh the comment", but another bot might want to "rebase", or "disable the feature", or whatever).

I'll update this PR accordingly.

tmatsuo: Greatly simplified the API. PTAL

packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -450,7 +465,54 @@ To update this comment, add \`${REFRESH_LABEL}\` label.
...
+    const commentMark = getCommentMark(context.payload.installation.id);
+
+    // If the comment is made by bots, and the comment has the refresh
+    // checkbox checked, we'll proceed.
+    if (
+      !context.payload.comment.body.includes(commentMark) ||
+      !context.payload.comment.body.includes(REFRESH_STRING)
+    ) {
+      return;
+    }
```
tbpg, line 499: Move this to the beginning of this block? No need to fetch the config if the comment isn't right (which doesn't require a network call).

tbpg (commented):
Much easier to follow. Thanks.

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: Thanks, it's a good call, done.

tmatsuo:
@tbpg Thanks, PTAL

tbpg (approved):

PR #1182, opened 2020-11-24
fix(snippet-bot): exits early for PRs with null head

### LINKED ISSUE #1180: Snippet bot: TypeError: Cannot read property 'owner' of null
We are seeing instances of this in the logs for snippet-bot:

```
TypeError: Cannot read property 'owner' of null
        at scanPullRequest (/workspace/build/src/snippet-bot.js:226:289)
        at /workspace/build/src/snippet-bot.js:464:15
        at runMicrotasks (<anonymous>)
        at async Promise.all (index 0)
        at async Promise.all (index 0)
        at async /workspace/node_modules/gcf-utils/build/src/gcf-utils.js:203:21
```

tmatsuo:
fixes #1180


tmatsuo:
Should I also check against `null`?

packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -286,6 +286,17 @@ async function scanPullRequest(
   const owner = context.payload.repository.owner.login;
   const repo = context.payload.repository.name;
 
+  // If the head repo is null, we can not proceed.
+  if (
```
JustinBeckwith, line 290: It's nice that this is explicit, but in the future you can say `if (context.payload.pull_request.head.repo) { ...` and it will do a falsey check that looks for both null and undefined.

packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -286,6 +286,17 @@ async function scanPullRequest(
   const owner = context.payload.repository.owner.login;
   const repo = context.payload.repository.name;
 
+  // If the head repo is null, we can not proceed.
```
JustinBeckwith, line 289: I suspect, in the `app.on('pull_request')` style code somewhere in this project, you could check for `pull_request.state` (or something like that), check if it's closed, and just short circuit all of the code paths if this PR is closed far before getting here.

JustinBeckwith (approved):
LGTM, with minor nits/suggestions

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: I actually prefer exiting early because I'll have less indentation in the code.

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: I think that's another good check. I created #1183

PR #1188, opened 2020-11-26
feat(do-not-merge): create Do Not Merge bot

### LINKED ISSUE #942: New Bot: Do-not-merge bot
When a PR is created/edited and the title includes a variant of "do not merge", or a "do not merge" label is added, create a blocking pull request review until the title is fixed and the "do not merge" label is removed.

When do not merge status is removed, dismiss the review, but do not approve.

tbpg:
The bot adds a failing check review when the do not merge label is on a
PR. When the label is removed, the check passes.

If the label is never added, there is no check at all to avoid being
spammy.

Fixes #942.


packages/do-not-merge/src/do-not-merge.ts
```diff
@@ -0,0 +1,126 @@
...
+  title: 'OK to merge, label not found',
+  summary: 'OK to merge, label not found',
+};
+
+export = (app: Application) => {
+  app.on(
+    [
+      'pull_request.labeled',
+      'pull_request.unlabeled',
+      'pull_request.synchronize', // To run the check on every commit.
```
JustinBeckwith, line 39: Curious - why synchronize?

packages/do-not-merge/src/do-not-merge.ts
```diff
@@ -0,0 +1,126 @@
...
+    ],
+    async (
+      context: Context<Webhooks.EventPayloads.WebhookPayloadPullRequest>
+    ) => {
+      const owner = context.payload.repository.owner.login;
+      const repo = context.payload.repository.name;
+      const sha = context.payload.pull_request.head.sha;
```
JustinBeckwith, line 46: Before any of the real logic, make sure to exit if the PR is currently closed.  I think `pull_request.head` may be null if the branch gets deleted, which sometimes happens right after the merge.

packages/do-not-merge/src/do-not-merge.ts
```diff
@@ -0,0 +1,126 @@
...
+      const existingCheck = await findCheck(context, owner, repo, sha);
+
+      if (!labelFound) {
+        logger.info('Do not merge label not found');
+        // If the check already exists, but it's not a success, make it a success.
+        if (existingCheck && existingCheck.conclusion !== 'success') {
+          logger.info(
+            `Updating check on ${context.payload.pull_request.url} to success`
+          );
+          await context.github.checks.update({
```
JustinBeckwith, line 61: It feels like this will be super noisy, creating another status check on every PR in every repository across the org.  Would it possibly make more sense to only create the check if it's a failing PR?  That does of course come with the problem that there's no way to tell GitHub "don't allow merging of a PR of any check fails". 🤔

packages/do-not-merge/test/fixtures/config/valid-config.yml
```diff
@@ -0,0 +1 @@
+randomBoolean: true
```
JustinBeckwith, line 1: Do we need this?

JustinBeckwith (commented):
This is really close

packages/do-not-merge/src/do-not-merge.ts
tbpg: Done. Good catch!

packages/do-not-merge/src/do-not-merge.ts
tbpg: This only updates to success when there is already an existing check, specifically for that reason. A check is only _added_ when it fails. However, we have to update the check to success if the label is removed. I don't think there is a way to dismiss a check entirely. See https://docs.github.com/en/free-pro-team@latest/rest/reference/checks#check-runs.

packages/do-not-merge/test/fixtures/config/valid-config.yml
tbpg: Nope. Deleted.

tbpg (commented):
Thanks! Please take another look.

README.md
```diff
@@ -12,6 +12,7 @@ by Google on GitHub.
 | [blunderbuss] | Assigns issues and PRs randomly to a specific list of users | [install][blunderbuss-app] |
 | [buildcop] | Listen on PubSub queue for broken builds, and open corresponding issues | [install][buildcop-app] |
 | [conventional-commit-lint] | PR checker that ensures that the commit messages follow conventionalcommits.org style | [install][conventional-commit-lint-app] |
+| [do-not-merge] | PR checker that ensures the do not label is not present | [install][do-not-merge-app]
```
bcoe, line 15: I believe this should read:

```
ensures the do not merge label is not present
```

packages/do-not-merge/src/do-not-merge.ts
```diff
@@ -0,0 +1,133 @@
...
+            conclusion: 'success',
+            check_run_id: existingCheck.id,
+            owner,
+            repo,
+            output: SUCCESS_OUTPUT,
+          });
+        }
+        return;
+      }
+      logger.info('Do not merge label found');
```
bcoe, line 78: some additional debugging in the `logger.info` might be useful, if you have `owner`/`repo`, it becomes easier to find  specific run.

packages/do-not-merge/test/test.ts
```diff
@@ -0,0 +1,172 @@
...
+
+import myProbotApp from '../src/do-not-merge';
+import {resolve} from 'path';
+import {Probot, createProbot, ProbotOctokit} from 'probot';
+import {config} from '@probot/octokit-plugin-config';
+import nock from 'nock';
+import {describe, it, beforeEach} from 'mocha';
+import snapshot from 'snap-shot-it';
+
+const TestingOctokit = ProbotOctokit.plugin(config);
```
bcoe, line 25: What if we pulled this `TestingOctokit` into `gcf-utils/test-helpers` or something, wondering how we could avoid this boiler plate in all our tests.

Not a blocker on this PR.

bcoe (approved):

README.md
tbpg: Done.

packages/do-not-merge/src/do-not-merge.ts
tbpg: Added.

packages/do-not-merge/test/test.ts
tbpg: I don't mind a bit of copying. But, seems good to look into.

tbpg (commented):
Thanks!

JustinBeckwith (approved):

PR #1197, opened 2020-12-01
build: use HEAD~1 if on main branch

### LINKED ISSUE #1156: GitHub Actions CI flakiness
See https://github.com/googleapis/repo-automation-bots/runs/1424133746?check_suite_focus=true.

```
Error: Command failed: git diff --name-only origin/
fatal: ambiguous argument 'origin/': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

Seems to be related to the new setup from #626.

cc @JustinBeckwith @bcoe @tmatsuo

bcoe:
When merging to the main branch, there is no GITHUB_BASE_REF.
Because we use a squash based workflow, we can use HEAD~1 instead.

Fixes #1156

### DIFF
--- a/.github/workflows/interrogate.js
+++ b/.github/workflows/interrogate.js
@@ -22,8 +22,7 @@ if (baseRef) {
 } else {
   // If we're on the main branch, run tests based on last commit:
   console.log(`running against last commit`);
-  execSync(`git reset --hard HEAD~1`, { encoding: 'utf-8'});
-  status = execSync(`git diff --name-only origin/master`, { encoding: 'utf-8'});
+  status = execSync(`git diff --name-only HEAD~1`, { encoding: 'utf-8'});
 }
 console.log(status);
 const changes = status.split('\n');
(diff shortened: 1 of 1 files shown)


.github/workflows/interrogate.js
```diff
@@ -15,8 +15,16 @@
...
-const status = execSync(`git diff --name-only origin/${baseRef}`, { encoding: 'utf-8'});
+let status;
+if (baseRef) {
+  console.log(`base ref: ${baseRef}`);
+  status = execSync(`git diff --name-only origin/${baseRef}`, { encoding: 'utf-8'});
+} else {
+  // If we're on the main branch, run tests based on last commit:
+  console.log(`running against last commit`);
+  execSync(`git reset --hard HEAD~1`, { encoding: 'utf-8'});
+  status = execSync(`git diff --name-only origin/master`, { encoding: 'utf-8'});
```
tbpg, line 26: Seems odd to `reset` then `diff`. Can we `diff` in a single command? `git diff --name-only HEAD~`?

(I'm nervous about the `reset` accidentally messing up other commands we add in the future, but only triggering the bug on the main branch.)

tbpg (changes requested):
Thank you!

tbpg (approved):

PR #1198, opened 2020-12-01
build: fetch the main branch

bcoe:
It looks like the checkout library doesn't fetch the main branch by default (it's probably a shallow clone?).


.github/workflows/interrogate.js
```diff
@@ -16,6 +16,11 @@ const {execSync} = require('child_process');
 execSync('git fetch origin master');
 const baseRef = process.env.GITHUB_BASE_REF;
 let status;
+
+execSync('git fetch origin master');
```
tbpg, line 20:
```suggestion
execSync('git fetch --depth 2 origin master');
```

tbpg (changes requested):

.github/workflows/interrogate.js
tmatsuo, line 20: It will break once github changes the default branch from `master` to something else.

bcoe: we re no longer using a backed in vlue.

.github/workflows/ci.yaml
```diff
@@ -14,6 +14,8 @@ jobs:
       requiredJobs: ${{ steps.interrogate.outputs.requiredJobs }}
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
tbpg, line 18: Why not 2 or some small number? Seems like this does the entire repo?

bcoe: @tbpg I tried a variety of other values and was consistently getting an error that HEAD~ was ambiguous. The docs indicate that setting `0` "Fetch all history for all tags and branches".

tmatsuo (approved):

JustinBeckwith (approved):

tbpg (approved):
Thank you for fixing this!

PR #1214, opened 2020-12-04
fix: update to latest @octokit/* dependencies

gr2m:
closes https://github.com/googleapis/repo-automation-bots/pull/1195

- [x] buildcop
- [x] conventional-commit-lint
- [x] failurechecker
- [x] header-checker-lint
- [x] label-sync


gr2m:
@googlebot I signed it!

packages/failurechecker/package.json
```diff
@@ -28,11 +28,10 @@
     "lint": "gts check"
   },
   "dependencies": {
-    "@octokit/rest": "^18.0.6",
     "gcf-utils": "^6.2.0"
   },
   "devDependencies": {
-    "@octokit/types": "^5.5.0",
+    "@octokit/types": "^6.0.0",
```
JustinBeckwith, line 34: Elsewhere this was replaced with `@octokit/openapi-types` - I'm curious - why keep this package here?  Is it still needed?

packages/monitoring-system/data-processor/package.json
```diff
@@ -29,7 +29,7 @@
     "@google-cloud/pubsub": "^2.6.0",
     "@google-cloud/tasks": "^2.1.2",
     "@octokit/rest": "^18.0.9",
-    "@octokit/types": "^5.5.0",
+    "@octokit/types": "^6.0.0",
```
JustinBeckwith, line 32: Same question as above - why are we keeping this here, but migrating elsewhere?

JustinBeckwith (approved):

packages/failurechecker/package.json
gr2m: I think `@octokit/types` can be removed altogether, it doesn't seem to be used at all

packages/monitoring-system/data-processor/package.json
gr2m: You use 

```ts
import {OctokitResponse} from '@octokit/types';
```

which is specific to the Octokit libraries and goes beyond the scope of GitHub's OpenAPI spec, so you need `@octokit/types` in this case

Issue #1159, opened 2020-11-19
Merge-on-Green: sometimes misses pull requests


chingor13:
We've noticed (especially when tagging in bulk) that merge-on-green sometimes misses PRs:

Some examples:
googleapis/java-scheduler#277
googleapis/java-redis#260
googleapis/java-pubsub#430
googleapis/java-dialogflow#366 (comment)
googleapis/java-cloudbuild#321 (comment)
googleapis/java-asset#446 (review)

I suggest we introduce a second cron that looks for issues tagged with the automerge labels and adds them to the db to watch if they are not already watched. This will catch any missed pull request (due to quota or possibly failed webhook delivery)

See more information: https://github.com/googleapis/java-cloud-bom/issues/735

bcoe:
Calling this a `p2`, since there's the workaround of manually landing; but we will work to prioritize this (in less than the 180 days `p2` implies).

chingor13:
This is causing a bunch of pain. In the interim, can we change the "eyes" reaction to a separate label like `automerge: watching`. This would allow us to use the API to find PRs tagged `automerge`, but not tagged with the `automerge: watching` label. AFAIK, you cannot query for the reaction.

chingor13:
#1217 should fix this, closing.

PR #1217, opened 2020-12-05
fix: add cron job to pick up hanging prs

sofisl:
This addresses #1159 

Following Jeff's suggestion, this PR adds functionality for a cron job that will run once an hour. 

I *believe* that, if we format the cron job correctly, we will be able to trigger the gcf-utils mechanism of rotating through all of the repositories under which MOG is installed. 

Taking a lead from the SRS cron job, I think we can just check if the `cron_org` exists in the payload to trigger the rotation through all of the repos, which is different from the current MOG cron job in Cloud Scheduler.

I am cautious about this approach - we may trigger an abuse mechanism or hit our rate limit faster since we'd be checking all PRs in all our repos. I feel like we should try this solution, but if it doesn't work, we may want to write another bot that just picks up PRs.


packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -331,6 +331,98 @@ handler.checkPRMergeability = async function checkPRMergeability(
...
+handler.pickUpPR = async function pickUpPR(
+  owner: string,
+  repo: string,
+  github: Context['github']
+) {
+  const prs = await github.paginate(await github.pulls.list, {
+    owner,
+    repo,
+    state: 'open',
+  });
```
chingor13, line 350: You can probably reduce the # of queries, by using the search API: https://docs.github.com/en/free-pro-team@latest/rest/reference/search#search-issues-and-pull-requests

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -331,6 +331,98 @@ handler.checkPRMergeability = async function checkPRMergeability(
...
+      } catch (err) {
+        err.message = `Error in getting branch protection\n\n${err.message}`;
+        await github.issues.createComment({
+          owner,
+          repo,
+          issue_number: pr.number,
+          body:
+            "Your PR doesn't have any required checks. Please add required checks to your master branch and then re-add the label. Learn more about enabling these checks here: https://help.github.com/en/github/administering-a-repository/enabling-required-status-checks.",
+        });
+        logger.error(err);
```
chingor13, line 391: Should this logic share the same code as the webhook handling?

chingor13 (changes requested):

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -331,6 +331,98 @@ handler.checkPRMergeability = async function checkPRMergeability(
...
+      }
+    );
+
+    const labelFound = labels.find(
+      (label: Label) =>
+        label.name === MERGE_ON_GREEN_LABEL ||
+        label.name === MERGE_ON_GREEN_LABEL_SECURE
+    );
+
+    if (labelFound) {
```
bcoe, line 368: we should only do this if there's not already an entry in `datastore` for the given PR right?

packages/merge-on-green/src/merge-on-green.ts
bcoe: 👍 I agree, I think the logic could be simplified if we:

1. check to see if the PR is already tracked in datastore (_if it is tracked do nothing._).
2. if the PR is not tracked, but an automerge label exists, make a synthetic call to `app.on('pull_request.labeled', async context`; I think you could do this by looking at the fields you use from context, and then making sure you populate these in a fake context object you pass.

bcoe (changes requested):
Added a couple notes. I agree with @chingor13, I think you could be clever and make it so that finding a labeled PR that isn't tracked hits the same code path as adding a label.

packages/merge-on-green/src/merge-on-green.ts
chingor13: I'm not sure about faking an webhook - we could probably extract that common logic into a function to execute from both paths.

sofisl: I simplified the logic here!

sofisl:
@chingor13 and @bcoe, made some adjustments, thanks for the review!

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -331,6 +432,77 @@ handler.checkPRMergeability = async function checkPRMergeability(
...
 
+//TODO: change the search query if expanding the allowlist to search for more authors
+/**
+ * For a given repository, looks through all the PRs and checks to see if they have a MOG label
+ * @param owner the owner of the repo
+ * @param repo the repo name
+ * @param context the context of the webhook payload
+ * @returns void
+ */
+handler.pickUpPR = async function pickUpPR(
```
chingor13, line 443: nit: can we call this `scanForMissingPullRequests` or something like that?

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -214,29 +288,56 @@ handler.checkIfPRIsInvalid = async function checkIfPRIsInvalid(
...
-      reactionId: wp.reactionId,
-      installationId: wp.installationId,
-    },
-    method: 'upsert',
-  };
-  await datastore.save(entity);
+handler.addPR = async function addPR(
+  wp: WatchPR,
+  url: string,
+  github: Context['github']
```
chingor13, line 300: Other bots are using this type: https://github.com/googleapis/repo-automation-bots/blob/78950c856ed7205341efd8908b4cca935ab1c5ab/packages/gcf-utils/src/gcf-utils.ts#L41

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -39,12 +39,12 @@ interface WatchPR {
...
   owner: string;
   state: 'continue' | 'stop';
-  branchProtection: string[];
+  branchProtection?: string[];
   label: string;
   author: string;
   url: string;
-  reactionId: number;
-  installationId: number;
+  reactionId?: number | undefined;
```
chingor13, line 46: nit: is `undefined` implicitly possible with the `?`

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -350,6 +522,22 @@ function handler(app: Application) {
...
+      if (context.payload.cron_org !== owner) {
+        logger.info(`skipping run for ${context.payload.cron_org}`);
+        return;
+      }
+      await handler.pickUpPR(owner, repo, context.github);
+      return;
+    }
+
     const start = Date.now();
     await handler.checkPRMergeability(watchedPRs, app, context);
```
chingor13, line 542: Is it worth trying to do these both async and awaiting them at the end? (it's going to complete all of `pickUpPR` before trying to do the merges.)

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -331,6 +432,77 @@ handler.checkPRMergeability = async function checkPRMergeability(
...
+  repo: string,
+  github: Context['github']
+) {
+  // Github does not support searching the labels with 'OR'.
+  // The searching for issues is considered to be an "AND" instead of an "OR" .
+  const issuesAutomergeLabel = await github.paginate(
+    await github.search.issuesAndPullRequests,
+    {
+      q:
+        'is:open is:pr user:googleapis user:GoogleCloudPlatform label:"automerge"',
```
chingor13, line 454: use the constant?

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -331,6 +432,77 @@ handler.checkPRMergeability = async function checkPRMergeability(
...
+    }
+  );
+
+  const issuesAutomergeExactLabel = await github.paginate(
+    await github.search.issuesAndPullRequests,
+    {
+      q:
+        'is:open is:pr user:googleapis user:GoogleCloudPlatform label:"automerge: exact"',
```
chingor13, line 462: use the constant?

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -331,6 +432,77 @@ handler.checkPRMergeability = async function checkPRMergeability(
...
+    if (!pullRequestInDatastore) {
+      logger.info('before adding PR');
+      await handler.addPR(
+        {
+          number: issue.number,
+          owner,
+          repo,
+          state: 'continue',
+          url: issue.html_url,
+          label: 'automerge',
```
chingor13, line 477: use the constant?

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -331,6 +432,77 @@ handler.checkPRMergeability = async function checkPRMergeability(
...
+    const pullRequestInDatastore: WatchPR = await handler.getPR(issue.html_url);
+    if (!pullRequestInDatastore) {
+      await handler.addPR(
+        {
+          number: issue.number,
+          owner,
+          repo,
+          state: 'continue',
+          url: issue.html_url,
+          label: 'automerge: exact',
```
chingor13, line 496: use the constant?

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -331,6 +432,77 @@ handler.checkPRMergeability = async function checkPRMergeability(
...
+ * @returns void
+ */
+handler.pickUpPR = async function pickUpPR(
+  owner: string,
+  repo: string,
+  github: Context['github']
+) {
+  // Github does not support searching the labels with 'OR'.
+  // The searching for issues is considered to be an "AND" instead of an "OR" .
+  const issuesAutomergeLabel = await github.paginate(
```
chingor13, line 450: These 2 searches can probably be started at the same time then awaited before we want to start executing on them.

packages/merge-on-green/src/merge-on-green.ts
sofisl: Oh, I'm expecting this to run on a separate cron job - we won't try merging on the same run as we try picking up the PRs.

sofisl:
@chingor13, really appreciate the in-depth review: it's super helpful to make sure we stay consistent within and outside the package. Thank you!

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -960,6 +1226,7 @@ describe('merge-on-green', () => {
 
       //This function is supposed to respond with an error
       it('does not add a PR if branch protection errors and comments on PR', async () => {
+        addPRStub.restore();
```
bcoe, line 1229: one thought here, I think we could use Sinon's sandbox functionality and just always restart the sandbox in a `beforeEach`?

bcoe (approved):
This is looking good to me, thanks for DRY(don't repeat yourself)ing things up 👍

packages/merge-on-green/test/merge-on-green.ts
sofisl: I think we are, line 852.

We need to restore sometimes in order to provide specific functionality for some specific tests.

chingor13: If you need different setup scenarios, you can split describe blocks with separate beforeEach implementations.

chingor13 (approved):

PR #1240, opened 2020-12-11
build: enable automatic publishing to npm

bcoe:
This should get publishing working again, using a similar approach to what I've been developing for other mono-repos.


.github/workflows/release.yaml
```diff
@@ -0,0 +1,83 @@
...
+      # Forking token does not have permission to add label, so we use
+      # this script to do so:
+      - id: label
+        if: ${{ steps.release-pr.outputs.pr }}
+        uses: actions/github-script@v3    
+        with:
+            github-token: ${{secrets.GITHUB_TOKEN}}
+            script: |
+              const [owner, repo] = process.env.GITHUB_REPOSITORY.split('/');
+              await github.issues.addLabels({
```
JustinBeckwith, line 29: We've already seen frequent cases where Octokit just kinda breaks their API.  This ... makes me nervous.  How do we know what version of Octokit this uses?  Is it pinned inside of `actions/githubb-script`?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,83 @@
...
+      - uses: actions/setup-node@v1
+        if: ${{ steps.tag-release.outputs.release_created }}
+        with:
+          node-version: 14
+          registry-url: 'https://wombat-dressing-room.appspot.com/'
+      # Actually publish to npm:
+      - name: publish
+        if: ${{ steps.tag-release.outputs.release_created }}
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```
JustinBeckwith, line 59: Is this a token from wombat dressing room?  This would be the first place where we use a token like this in GitHub Secrets, right?

JustinBeckwith (commented):
I'm not strictly against it, but there's a lot of ceremony in here, and I'm a little nervous about the npm token in github secrets.  Let's maybe talk this one through a bit?

.github/workflows/release.yaml
bcoe: @JustinBeckwith Octokit has much more frequently broken their types than their actual API, which I've rarely needed to make an update for. GitHub script is managed by the actions organization itself, and I'm fairly confident in its stability.

.github/workflows/release.yaml
bcoe: This is a token from wombat dressing room. This is the same approach we're using for _googleapis_, to manage the monorepo, and for [release-please-action](https://github.com/google-github-actions/release-please-action).

Given that the token has 2FA tied to GitHub releases, and that wombat dressing room means we don't need to provide our actual OTP secrets in actions, I think this risk is worth the ability we gain of moving publication away from kokoro, which I've found limiting in its functionality compared to actions.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,83 @@
...
+        run: |
+          npm install
```
bcoe, line 61: this should cd into packages/gcf-utils, prior to publishing.

bcoe:
Need to update to publish from correct folder, also we can use a gcf-utils specific token.

.github/workflows/release.yaml
bcoe: It's also crossed my mind that perhaps I could deliver an easy to configure testing harness for GitHub script; then we could use it as a hammer for small tasks like adding labels, but have the confidence of tests.

bcoe:
I have started a security audit related to this.

PR #1243, opened 2020-12-11
feat(snippet-bot): add FYI comment for sample browser page removal

### LINKED ISSUE #1242: feat(snippet-bot): handle sample browser pages differently
snippet-bot started to warn when you remove a region tag used on devsite pages.

However, if the page is created by sample browser, that page will be removed automatically, so snippet-bot should not warn in that case.

Instead snippet-bot can give a FYI comment about removing sample browser page.

tmatsuo:
fixes #1242

This change is safe to deploy without the new json file.


JustinBeckwith (changes requested):
This needs tests :)

tmatsuo:
@JustinBeckwith It's actually tested with the new snapshot :)

JustinBeckwith (approved):
If snapshots cover, LGTM.

PR #1245, opened 2020-12-11
feat(snippet-bot): allow `UNTRACKED` SnippetStatus

### LINKED ISSUE #1244: feat(snippet-bot): allow `UNTRACKED` SnippetStatus
Drift team thinks we should also add `UNTRACKED` region tags. The bot should allow that value.

tmatsuo:
fixes #1244


tmatsuo:
> Here is the summary of possible violations 😱

@snippet-bot It's ok friend :)

chingor13 (commented):
Can we add a test for this so we can show what the bot is or isn't doing for these?

tmatsuo:
@chingor13 I updated the fixture and snapshot to do that.
We already have cases with and without warnings. I added an additional warning to the existing snapshot.

Technically we can have another test case, but I think the current test suffice. Let me know if you think differently.

chingor13 (approved):

PR #1250, opened 2020-12-14
refactor: remove publish bot

JustinBeckwith:
@bcoe I wasn't sure if you were ready for this quite yet, but I would very much like to get the renovate automatic lock file updates working soon, and this is blocking it :)

Also, this may require a little admin override to handle, since the current CI scripts don't really account for entire bots getting deleted.


bcoe:
I feel comfortable getting rid of this, sorry that this has died on the vine over the holidays.

bcoe (approved):

PR #1261, opened 2020-12-23
fix: hanging cron pr

sofisl:


packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -518,16 +518,9 @@ function handler(app: Application) {
 
     //because we're searching for the PRs, and not getting the installation ID, we have to use
     //the bot's installation ID to call the API. So, we need to make sure it matches the repo owner
-    if (context.payload.cron_org) {
+    if (context.payload.pickUp === true) {
```
bcoe, line 521: is the `pickUp` variable set in the cron? I love this design of using the search API, but don't love the variable name `pickUp`.

When would `pickUp` not be set?

sofisl: I just made up a random variable name that we could add to the body. I just changed it to find_hanging_prs, and I was thinking of only setting it as true in that cron job!

bcoe (approved):

PR #1266, opened 2021-01-08
fix(conventional-commit): Allow longer body line length

Neenu1995:
Changed body line length limit from 100 to 256 to accommodate longer line length


bcoe (approved):
Thank you for the contribution 👍

PR #1269, opened 2021-01-09
fix(merge-on-green): second attempt to add cron to pick up hanging PRs

sofisl:
Attempt to readdress #1159, fixing the issue that caused reverting in #1264.

I believe the issue was that reaction IDs were not being added to stored PRs, causing the main (merge-logic) to fail. This was because I was attempting to make reaction and reaction IDs optional. To fix, I've essentially copied the exact format in which PRs are currently being added, and made it into its own function (addPR). 

I have also been doing some manual testing in the GCF function just to confirm this works. I was able to spot another potential bug, which is that we need to perform scans for missing PRs on different cron jobs with different installation IDs depending on the org, otherwise we won't have access to info like branch protection, etc.

I've also rescoped tests, which was a hanging comment left from #1217

### DIFF
--- a/packages/merge-on-green/src/merge-on-green.ts
+++ b/packages/merge-on-green/src/merge-on-green.ts
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2021 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/packages/merge-on-green/src/merge-on-green.ts
+++ b/packages/merge-on-green/src/merge-on-green.ts
@@ -513,10 +513,11 @@ function handler(app: Application) {
 
         return;
       }
-      const start = Date.now();
-      await handler.checkPRMergeability(watchedPRs, app, context);
-      logger.info(`mergeOnGreen check took ${Date.now() - start}ms`);
     }
+
+    const start = Date.now();
+    await handler.checkPRMergeability(watchedPRs, app, context);
+    logger.info(`mergeOnGreen check took ${Date.now() - start}ms`);
   });
   app.on('pull_request.labeled', async context => {
     const prNumber = context.payload.pull_request.number;
--- a/packages/merge-on-green/src/merge-on-green.ts
+++ b/packages/merge-on-green/src/merge-on-green.ts
@@ -18,7 +18,7 @@ import {Datastore} from '@google-cloud/datastore';
 import {mergeOnGreen} from './merge-logic';
 import {logger} from 'gcf-utils';
 
-type GitHubType = InstanceType<typeof ProbotOctokit>
+type GitHubType = InstanceType<typeof ProbotOctokit>;
 
 const TABLE = 'mog-prs';
 const datastore = new Datastore();
--- a/packages/merge-on-green/src/merge-on-green.ts
+++ b/packages/merge-on-green/src/merge-on-green.ts
@@ -504,19 +504,12 @@ function handler(app: Application) {
       logger.info('Entering job to pick up any hanging PRs');
       // we cannot search in an org without the bot installation ID, so we need
       // to divide up the cron jobs based on org
-      if (context.payload.org === 'googleapis') {
-        await handler.scanForMissingPullRequests(context.github, 'googleapis');
-      }
-      if (context.payload.org === 'GoogleCloudPlatform') {
-        await handler.scanForMissingPullRequests(
-          context.github,
-          'GoogleCloudPlatform'
-        );
-
-        return;
-      }
+      await handler.scanForMissingPullRequests(
+        context.github,
+        context.payload.org
+      );
+      return;
     }
-
     const start = Date.now();
     await handler.checkPRMergeability(watchedPRs, app, context);
     logger.info(`mergeOnGreen check took ${Date.now() - start}ms`);
--- a/packages/merge-on-green/src/merge-on-green.ts
+++ b/packages/merge-on-green/src/merge-on-green.ts
@@ -18,6 +18,8 @@ import {Datastore} from '@google-cloud/datastore';
 import {mergeOnGreen} from './merge-logic';
 import {logger} from 'gcf-utils';
 
+type GitHubType = InstanceType<typeof ProbotOctokit>
+
 const TABLE = 'mog-prs';
 const datastore = new Datastore();
 const MAX_TEST_TIME = 1000 * 60 * 60 * 6; // 6 hr.
@@ -144,7 +146,7 @@ handler.cleanUpPullRequest = async function cleanUpPullRequest(
   prNumber: number,
   label: string,
   reactionId: number,
-  github: InstanceType<typeof ProbotOctokit>
+  github: GitHubType
 ) {
   await github.issues.removeLabel({
     owner,
@@ -176,7 +178,7 @@ handler.checkIfPRIsInvalid = async function checkIfPRIsInvalid(
   label: string,
   reactionId: number,
   url: string,
-  github: InstanceType<typeof ProbotOctokit>
+  github: GitHubType
 ) {
   let pr;
   let labels;
@@ -235,7 +237,7 @@ handler.checkForBranchProtection = async function checkForBranchProtection(
   owner: string,
   repo: string,
   prNumber: number,
-  github: InstanceType<typeof ProbotOctokit>
+  github: GitHubType
 ): Promise<string[] | undefined> {
   let branchProtection: string[] | undefined;
   // Check to see if branch protection exists
@@ -274,7 +276,7 @@ handler.checkForBranchProtection = async function checkForBranchProtection(
 handler.addPR = async function addPR(
   incomingPR: IncomingPR,
   url: string,
-  github: InstanceType<typeof ProbotOctokit>
+  github: GitHubType
 ) {
   let branchProtection: string[] | undefined;
   try {
@@ -420,7 +422,7 @@ handler.checkPRMergeability = async function checkPRMergeability(
  * @returns void
  */
 handler.scanForMissingPullRequests = async function scanForMissingPullRequests(
-  github: InstanceType<typeof ProbotOctokit>,
+  github: GitHubType,
   org: string
 ) {
   // Github does not support searching the labels with 'OR'.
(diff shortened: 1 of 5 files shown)


packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -213,30 +224,102 @@ handler.checkIfPRIsInvalid = async function checkIfPRIsInvalid(
...
+ * @param owner type string
+ * @param repo type string
+ * @param prNumber type number
+ * @param github type githup API surface from payload
+ */
+handler.checkForBranchProtection = async function checkForBranchProtection(
+  owner: string,
+  repo: string,
+  prNumber: number,
+  github: InstanceType<typeof ProbotOctokit>
```
bcoe, line 238: nit, you could potentially do this once by going:

```
type GitHubType = InstanceType<typeof ProbotOctokit>
```

towards the top of the file.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -213,30 +224,102 @@ handler.checkIfPRIsInvalid = async function checkIfPRIsInvalid(
...
-  };
-  await datastore.save(entity);
+handler.addPR = async function addPR(
+  incomingPR: IncomingPR,
+  url: string,
+  github: InstanceType<typeof ProbotOctokit>
+) {
+  let branchProtection: string[] | undefined;
+  try {
+    branchProtection = await handler.checkForBranchProtection(
```
bcoe, line 281: It feels a little strange to me that we're checking `branchProtection` before adding the PR, since this is something we check in `merge-logic.ts` right?

From a separation of functionality point of view, I'd be tempted to try to keep all the logic for if something _should_ be merged in `merge-logic.ts`, whereas `merge-on-green.ts` just handles:

* receiving Webhooks from GitHub 
* adding PRs to the DB (either based on webhook or your periodic cron job).

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -331,6 +414,70 @@ handler.checkPRMergeability = async function checkPRMergeability(
...
+  org: string
+) {
+  // Github does not support searching the labels with 'OR'.
+  // The searching for issues is considered to be an "AND" instead of an "OR" .
+  const [issuesAutomergeLabel, issuesAutomergeExactLabel] = await Promise.all([
+    github.paginate(github.search.issuesAndPullRequests, {
+      q: `is:open is:pr user:${org} label:"${MERGE_ON_GREEN_LABEL}"`,
+    }),
+    github.paginate(github.search.issuesAndPullRequests, {
+      q: `is:open is:pr user:${org} label:"${MERGE_ON_GREEN_LABEL_SECURE}"`,
```
bcoe, line 433: I love the search approach that you came up with for this.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -350,11 +497,28 @@ function handler(app: Application) {
       await handler.cleanDatastoreTable(watchedPRs, app, context);
       return;
     }
+
+    if (context.payload.find_hanging_prs === true) {
+      logger.info('Entering job to pick up any hanging PRs');
+      // we cannot search in an org without the bot installation ID, so we need
+      // to divide up the cron jobs based on org
+      if (context.payload.org === 'googleapis') {
+        await handler.scanForMissingPullRequests(context.github, 'googleapis');
```
bcoe, line 506: rather than using an `if` statement here, you could just do:

`await handler.scanForMissingPullRequests(context.github, context.payload.org);`

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -831,142 +845,76 @@ describe('merge-on-green', () => {
...
 
-      it('does not delete a PR if it is not invalid', async () => {
-        const scopes = [getPRCleanUp('open', false), getLabels('automerge')];
-
-        await probot.receive({
-          name: 'schedule.repository' as '*',
-          payload: {org: 'testOwner', cleanUp: true},
-          id: 'abc123',
+      it('adds a PR if branch protection when PR labeled', async () => {
+        //addPRStub.restore();
```
bcoe, line 886: we can drop this commented out bit I believe. I'd be tempted to use Sinon's sandbox mode, and just always reset the sandbox in a before filter.

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -831,142 +845,76 @@ describe('merge-on-green', () => {
...
-
         await probot.receive({
           name: 'pull_request',
           payload,
           id: 'abc123',
         });
 
         scopes.forEach(s => s.done());
-        assert(addPRStub.called);
+        addPRStub.restore();
```
bcoe, line 909: I think a `beforeEach` that does the sandbox dance would simplify this.

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -981,109 +929,268 @@ describe('merge-on-green', () => {
         });
 
         scopes.forEach(s => s.done());
-
-        assert(!addPRStub.called);
-
-        logger.info('stub called? ' + addPRStub.called);
+      });
+    });
+    describe('stubbing all GCP functions', () => {
```
bcoe, line 934: I wouldn't make the describe block describe "how you're going to write your tests" (_which is just an implementation detail), rather it should describe what you're testing.

If I'm following, we're moving away from `nock` based tests with these new tests towards using sinon more?

What I might do is go so far as splitting it across two test files:

* integration.js <-- integration(ish) tests with all our nock tests.
* unit.js <-- unit tests using sinon mianly for mocking.

This could be follow up work, not a blocker 👍

bcoe (changes requested):
This is looking really solid to me, a few things jumped out at me:

* it feels a bit strange tome that we're checking branch protection in `merge-on-green.ts`, because I thought our, _"should we merge it?", logic lived elsewhere. I see this logic isn't new, so it might just be I'm not understanding a valuable optimization that's been added.
* **non blocker**: It jumped out at me that we should perhaps have two sets of tests, those using `nock`, and those new ones you're adding using `sinon` (_I like the movement towards sinon._)
* I think we could cleanup tests quite a bit using sinon's sandbox API, and think this would be worth considering.

packages/merge-on-green/src/merge-on-green.ts
sofisl: Yeah! I think that was the thinking behind our original logic. The reason I changed where we check for branch protection is because it's a way of reducing API requests during the main` merge-logic`, especially since branch protection isn't really the kind of information that changes within the life of a PR (so it's not the kind of thing we need to check multiple times before merging a PR). The way we currently have it set up allows us to only check for branch protection once, and then we save it in our Datastore table and don't have to make that API call again. Another reason (though this one is weaker) is that it kind of makes sense for us to only add a PR to our queue of PRs to check if it has branch protection. In other words, instead of: (a) adding a PR to datastore, (b) checking if it has branch protection, (c) deleting the PR, we could save ourselves two steps and just never add the PR to Datastore if it doesn't have branch protection. Open to hearing your thoughts on this process!

packages/merge-on-green/src/merge-on-green.ts
sofisl: We could... we're just going to end up with a lot of errors in the logs and additional API calls that we don't need to make. I'm not sure what is better here, less lines of code and more logs or vice versa!

packages/merge-on-green/test/merge-on-green.ts
sofisl: Oops! Thanks for catching that! I was testing to make sure the sandboxing worked.

packages/merge-on-green/test/merge-on-green.ts
sofisl: I ended up dividing the functions into `merge-logic` vs. `merge-on-green` wrapper logic, which I think does a better job of describing the split (although, to your point, maybe it's worth adding more integration tests to have its own separate file)

packages/merge-on-green/src/merge-on-green.ts
bcoe: My temptation would be to still have the logic for checking branch protection inside of the `merge-logic.js` class, and you could just call that method here (rather than having the logic in two places).

This doesn't need to be a blocker for this PR, but perhaps we could make a tracking ticket to do this?

packages/merge-on-green/src/merge-on-green.ts
bcoe: what I"m getting at is that `context.payload.org` is going to be one of `googleapis` or `GoogleCloudPlatform` I believe?

---

Why do we return early for `GoogleCloudPlatform` but not `googleapis`, out of curiosity?

bcoe (commented):
This is mostly looking good to me 👍 

I'm still confused by why we need to have this:

` if (context.payload.org === 'googleapis') {`

When we could just pass that string as input to `scanForMissingPullRequests`?

---

Otherwise I just had a future refactoring nit.

packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -194,658 +122,110 @@ describe('merge-on-green', () => {
...
+        payload,
+        id: 'abc123',
       });
+
+      scopes.forEach(s => s.done());
     });
   });
-
-  describe('merge-on-green wrapper logic', () => {
+  suite('stubbing all GCP functions', () => {
```
bcoe, line 212: This seems like a good middle-ground, compared to breaking out the tests into multiple files, or completely restructuring tests.

sofisl:
Thanks Ben!

1. For this comment: `if (context.payload.org === 'googleapis') `
You were right, I was confused! I fixed it to just `await handler.scanForMissingPullRequests(context.github, context.payload.org);`

2. Also, we were not supposed to return early for GoogleCloudPlatform vs. googleapis. The return statement should have been outside of the if-blocks. Another mistake, good catch!

3. As for moving branch protection to `merge-logic` vs. keeping it where it is, my only concern would be that it's not really in our merge logic: we're not checking it every time to see if it's there in order to merge, we only check it once to see if it should be added to Datastore.
Perhaps a future ticket might be to create a `github-utils` file that contains functions that call to the API, and another one for functions that call to GCP? I'm not sure if this solution would address your concern.

bcoe (approved):

sofisl:
I've tested pre-merging, it is working!

PR #1275, opened 2021-01-12
chore(deps): update lock-file maintenance

sofisl:
Fixes broken tests in #1270


packages/auto-label/package-lock.json
```diff
@@ -1,8 +1,8118 @@
 {
   "name": "auto-label",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
JustinBeckwith, line 4: We want to avoid making this specific change.  This happens when npm v7 is used to generate the lock file, and is exactly what we asked renovate to fix in https://github.com/renovatebot/renovate/issues/8253 (looks like they rolled out a fix).  For now I would drop all of the lock file updates broadly, and just try to get the one project that was busted healthy so the renovatebot can do this update moving forward.

JustinBeckwith (changes requested):

packages/auto-label/package-lock.json
sofisl: Oh I see, sorry for the confusion. PR #1276 should update those tests.

PR #1279, opened 2021-01-14
feat: bazel bot is bash script

SurferJeffAtGoogle:


bcoe:
To get tests passing, I think we'll need to add an exception to the `.github/workflows/changeFinder`, so that it doesn't try to run `npm test` inside the folder with a shell script.

SurferJeffAtGoogle:
> To get tests passing, I think we'll need to add an exception to the `.github/workflows/changeFinder`, so that it doesn't try to run `npm test` inside the folder with a shell script.

Which raises the question, is this the most appropriate directory structure?  How would you lay it out?

JustinBeckwith:
The layout we have today makes the assumption that `packages/` is full of tiny node.js applications that all match the same patterns of supporting `npm install`, `npm test`, etc.  It assumes that they're each a bot.  The path of least resistance would be to make this a node.js program, but I also recognize that may be overkill :) 

Unless we're turning this into a node.js app, I would recommend moving it either to a top level directory, or creating a directory like 'cloudbuild-jobs' at the top level that contains these kinds of Docker based non-nodejs-ey tools.

packages/bazel-bot/cloudbuild.yaml
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+steps:
+  - name: 'gcr.io/repo-automation-bots/bazel-bot'
+
```
bcoe, line 17: nit: I think this is an extra newline, we should make linters catch this so that we don't waste people's time in review.

packages/bazel-bot/docker-image/docker-main.sh
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+git clone https://github.com/googleapis/googleapis.git
+git clone git@github.com:googleapis/googleapis-gen.git
+
+export GOOGLEAPIS=googleapis
```
bcoe, line 19: could we default to `googleapis`, but allow other variables to be passed in:

`${GOOGLEAPIS:="googleapis"}`

This might make it easier to configure things for testing.

packages/bazel-bot/docker-image/docker-main.sh
```diff
@@ -0,0 +1,26 @@
...
+export GOOGLEAPIS_GEN=googleapis-gen
+export BAZEL_REMOTE_CACHE=https://storage.googleapis.com/repo-automation-bots-bazel-cache
+
+# TODO: install git credentials.
```
bcoe, line 23: Since the container runs longer than 10 minutes, I think we will want to pass in a GitHub app's private key as an argument, which would then be used to create a short lived (10 minute) JWT, to perform operations on GitHub as needed.

I think we shouldn't build this into the container, as it would make leaking the container a security risk.

packages/bazel-bot/docker-image/generate-googleapis-gen.sh
```diff
@@ -0,0 +1,107 @@
...
+    # Some api always fails to build.
+    # TODO: file a bug when something fails to build.
+    set +e
+    # Invoke bazel build.
+    (cd "$GOOGLEAPIS" && bazel build -k \
+          --remote_cache=$BAZEL_REMOTE_CACHE \
+          --google_default_credentials $targets)
+    
+    # Clear out the existing contents of googleapis-gen.
+    rm -rf "$GOOGLEAPIS_GEN/external" "$GOOGLEAPIS_GEN/google" "$GOOGLEAPIS_GEN/grafeas"
```
bcoe, line 75: Do we have  plan to serialize builds, so that we don't end up with potential race conditions on this step?

packages/bazel-bot/docker-image/generate-googleapis-gen.sh
```diff
@@ -0,0 +1,107 @@
...
+    if [[ -z "$BUILD_TARGETS" ]] ; then
+        targets=$(cd "$GOOGLEAPIS" && bazel query 'filter(".*\.tar\.gz$", kind("generated file", //...:*))')
+    else
+        targets="$BUILD_TARGETS"
+    fi
+    # Clean out all the source packages from the previous build.
+    rm -f $(find -L "$GOOGLEAPIS/bazel-bin" -name "*.tar.gz")
+    # Some api always fails to build.
+    # TODO: file a bug when something fails to build.
+    set +e
```
bcoe, line 68: I think this line would be worth a comment, why don't we want to fail immediately here?

packages/bazel-bot/docker-image/test-generate-googleapis-gen.sh
```diff
@@ -0,0 +1,50 @@
...
+    sha=$(git -C googleapis log -3 --format=%H | tail -1)
+
+    # Create a fake googleapis-gen with the sha tag.
+    rm -rf googleapis-gen googleapis-gen-clone
+    mkdir googleapis-gen
+    git -C googleapis-gen init
+    echo hello > googleapis-gen/hello.txt
+    git -C googleapis-gen add -A
+    git -C googleapis-gen commit -m "Hello world."
+    git -C googleapis-gen tag "googleapis-$sha"
```
bcoe, line 34: @SurferJeffAtGoogle @JustinBeckwith I propose a compromise, where we keep this code in `packages/`, but have `npm` kick off these unit tests.

bcoe (changes requested):
Left a few nits and architectural suggestions.

Rather than moving the bash script into another directory, my preference would be that we make `npm test` kick off the tests you've written in bash -- you don't even need to write JavaScript to do this, it could be something like:

**package.json:**

```json
{
  "scripts": {
    "test": "bash test-generate-googleapis-gen.sh"
  }
}
```

packages/bazel-bot/docker-image/docker-main.sh
SurferJeffAtGoogle: Done.

packages/bazel-bot/docker-image/docker-main.sh
SurferJeffAtGoogle: Yep.  Let's do it in the next pull request.

packages/bazel-bot/docker-image/generate-googleapis-gen.sh
SurferJeffAtGoogle: The current algorithm is designed to work correctly with race conditions.  If two builds trigger at the same time, one will win and the other will lose, and the repo will be left in a good state.

packages/bazel-bot/docker-image/test-generate-googleapis-gen.sh
SurferJeffAtGoogle: Done.

packages/bazel-bot/docker-image/generate-googleapis-gen.sh
SurferJeffAtGoogle: I expanded the comment.

SurferJeffAtGoogle:
Ready for another look.

JustinBeckwith:
As this is written today, I worry the build/deploy scripts hiding in scripts/ will be triggered, and we'll try deploying cloud functions with this content.  Could I trouble you to double check those workflows?

SurferJeffAtGoogle:
> As this is written today, I worry the build/deploy scripts hiding in scripts/ will be triggered, and we'll try deploying cloud functions with this content. Could I trouble you to double check those workflows?

@bcoe  investigated this yesterday, and there was already a similar case.  Ben, can you confirm?

bcoe:
@JustinBeckwith @SurferJeffAtGoogle I was able to merge the `owl-bot-post-processor` without breaking builds, even though it does not yet have an app associated with it.

If we find that changes in bazel bot do end up causing annoying failures when merging, we can add an exception to the build scripts.

JustinBeckwith (approved):
Please let @bcoe give an LGTM before merging

bcoe (changes requested):
I notice that the test output is just the word `Error: `

Otherwise things are looking good to me.

SurferJeffAtGoogle:
> I notice that the test output is just the word `Error: `

I looked for that in https://github.com/googleapis/repo-automation-bots/runs/1736821574 and didn't see it.  Where do you see it?

bcoe:
> I looked for that in https://github.com/googleapis/repo-automation-bots/runs/1736821574 and didn't see it. Where do you see it?

@SurferJeffAtGoogle it looks good now, I must have been looking at a prior build.

bcoe (approved):

PR #1280, opened 2021-01-15
fix(deps): upgrade to Probot 11

sofisl:
If this looks okay, I can publish and start upgrading the rest of the bots!


JustinBeckwith (approved):

packages/gcf-utils/package.json
```diff
@@ -40,7 +40,7 @@
     "get-stream": "^6.0.0",
     "into-stream": "^6.0.0",
     "pino": "^6.3.2",
-    "probot": "10.9.5",
+    "probot": "^11.0.1",
```
JustinBeckwith, line 43: Only worry - the floating `^` tends to bite us, and we used to directly pin.  Should we keep doing that?

sofisl: Oh, right, that makes sense! I'm fine keeping it at 11, then just manually upgrading when they move to 12.

PR #1290, opened 2021-01-20
feat(release-please): enable configuring multiple base branches

### LINKED ISSUE #1287: release-please: allow releasing from multiple branches
Branches should be configurable to optionally set different rules from the default branch.

This will allow supporting backported releases more easily.

chingor13:
Adds a new configuration option for `branches`. This option lets you list more target branches that can each have their own configuration.

When listening to push/label events, we search all the available branch configurations for the settings to apply to that branch.

When running with cron, we iterate over the default branch and any additional configured branches.

Fixes #1287


packages/release-please/src/release-please.ts
```diff
@@ -84,37 +94,30 @@ function releaseTypeFromRepoLanguage(language: string | null): string {
...
+  // look at primaryBranch first
+  if (branch === config.primaryBranch) {
+    return {
+      ...config,
+      ...{branch},
+    };
   }
 
-  await Runner.runner(ReleasePRFactory.build(releaseType, buildOptions));
+  if (!config.branches) {
```
bcoe, line 109: If no `primaryBranch` was set, but a config file existed, our behavior was to assume `master`, does this logic remain the same?

packages/release-please/src/release-please.ts
```diff
@@ -146,11 +149,51 @@ async function createGitHubRelease(
   await Runner.releaser(ghr);
 }
 
+async function doRelease(
```
bcoe, line 152: this is rather than `createReleasePR`?

packages/release-please/test/release-please.ts
```diff
@@ -345,6 +411,38 @@ describe('ReleasePleaseBot', () => {
       assert(executed, 'should have executed the runner');
     });
 
+    it('should try to create a release on an alternate branch', async () => {
```
bcoe, line 414: Do we have a test for config [like this](https://github.com/googleapis/nodejs-secret-manager/blob/master/.github/release-please.yml)

```
releaseType: node
```

Where there's a releaser type, but no branch? If not let's add one.

bcoe (approved):
A couple suggestions and questions.

packages/release-please/src/release-please.ts
chingor13: In the additional branches configuration, `branch` is required.

packages/release-please/src/release-please.ts
chingor13: Originally, `doRelease` called both `createReleasePR` and `creatGitHubRelease`, but was later refactored to keep them separate. I'll rename it back

packages/release-please/test/release-please.ts
chingor13: Added a test in https://github.com/googleapis/repo-automation-bots/pull/1290/commits/0448967ffbbdc72e8c6692df313d7704c376aba2

bcoe (approved):

PR #1300, opened 2021-01-21
fix: authenticate with github app secret

SurferJeffAtGoogle:


packages/bazel-bot/docker-image/Dockerfile
```diff
@@ -14,16 +14,29 @@
 
 # build with:
 # docker build -t gcr.io/repo-automation-bots/bazel-bot .
+
+##########################################################
+# Build the command-line tool that lets us sign JWTs.
+FROM rust:1.49-buster
+RUN git clone https://github.com/mike-engel/jwt-cli.git
```
bcoe, line 21: why don't we clone this at a release tag or sha, so that we know the API surface won't change on us.

packages/bazel-bot/docker-image/install-credentials.sh
```diff
@@ -0,0 +1,34 @@
...
+    -H "Authorization: Bearer $JWT" \
+    -H "Accept: application/vnd.github.v3+json" \
+    https://api.github.com/app/installations/$GITHUB_APP_INSTALLATION_ID/access_tokens \
+    | jq -r .token)
+
+# According to https://cli.github.com/manual/gh_auth_login
+echo "$GITHUB_TOKEN" | gh auth login --with-token
+
+# According to https://docs.github.com/en/developers/apps/authenticating-with-github-apps#http-based-git-access-by-an-installation
+git -C "$GOOGLEAPIS_GEN" remote set-url origin \
```
bcoe, line 33: I used the credentials helper to setup the authentication, I'm wondering if that makes it less likely that a URL shows up in logs with the token?

https://github.com/googleapis/repo-automation-bots/blob/master/packages/owl-bot-post-processor/cloud-build/update-pr.yaml#L51

packages/bazel-bot/test-generate-googleapis-gen.sh
```diff
@@ -23,10 +23,10 @@ cd "$TEST_WORKDIR"
 
 # Clone googleapis
 if [ ! -d "googleapis" ] ; then
-    set +e  # Git clone returns an error code because config user.email is not set.
+    # Git clone returns an error code because config user.email is not set.
     # There's no way to set it for the repo and clone it at the same time.  :-/
-    git clone https://github.com/googleapis/googleapis
-    set -e
+    # || true effectively ignares the error.
```
bcoe, line 28:
```suggestion
    # || true effectively ignores the error.
```

packages/bazel-bot/cloudbuild.yaml
```diff
@@ -13,4 +13,7 @@
 # limitations under the License.
 
 steps:
-  - name: 'gcr.io/repo-automation-bots/bazel-bot'
\ No newline at end of file
+  - name: 'gcr.io/repo-automation-bots/bazel-bot'
+    env:
+      - 'BAZEL_FLAGS=--google_default_credentials --remote_cache=https://storage.googleapis.com/repo-automation-bots-bazel-cache'
```
bcoe, line 18: Is this refactor related? or just cleanup.

bcoe (changes requested):
Just a few nits mainly, and a question.

packages/bazel-bot/docker-image/install-credentials.sh
SurferJeffAtGoogle: I tried that too.  However, I got a permission denied issue.  I think I'm using it in a different context, maybe because I clone with https and later switch the url to ssl.

packages/bazel-bot/cloudbuild.yaml
SurferJeffAtGoogle: An unrelated improvement.

packages/bazel-bot/docker-image/Dockerfile
SurferJeffAtGoogle: Done.

packages/bazel-bot/test-generate-googleapis-gen.sh
SurferJeffAtGoogle: Fixed.

SurferJeffAtGoogle:
Ready for another look.

packages/bazel-bot/docker-image/install-credentials.sh
bcoe: 👍 we should just make sure that the Cloud Build logs are anywhere publicly accessible, and that the token doesn't show up in the logs.

bcoe (approved):

bcoe (approved):
This looks good to me, my only concern with using the URL based approach for authentication is if it shows up in logs anywhere -- it would be good to confirm once this is live that we don't see the token.

Since the tokens only live for 10 minutes, an accidental leak is lower risk than leaking an app token.

PR #1308, opened 2021-01-22
DNM: fix(blunderbuss): upgrade to probot 11/gcf-utils 7

sofisl:
This upgrade took a major refactor of the code because Probot does not allow for ambiguous types. The previous version of this bot allowed for either issues or PRs to come in, and then it would treat each webhook appropriately. With the new version of gcf-utils, this receives an error:

```
      const issue: Issue =
        context.payload.issue || context.payload.pull_request;
```

```
Property 'pull_request' does not exist on type 'WebhookPayloadIssues | WebhookPayloadPullRequest'.
  Property 'pull_request' does not exist on type 'WebhookPayloadIssues'.
```

I tried a couple of different solutions, but ultimately I found it was necessary to divide the hooks into PRs and Issues.

### DIFF
--- a/packages/blunderbuss/src/blunderbuss.ts
+++ b/packages/blunderbuss/src/blunderbuss.ts
@@ -13,9 +13,13 @@
 // limitations under the License.
 
 // eslint-disable-next-line node/no-extraneous-import
-import {Application, Context} from 'probot';
+import {Probot, Context} from 'probot';
 import * as util from 'util';
 import {logger} from 'gcf-utils';
+import {EventPayloads} from '@octokit/webhooks';
+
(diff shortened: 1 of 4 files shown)


packages/blunderbuss/src/blunderbuss.ts
```diff
@@ -117,124 +239,94 @@ export function blunderbuss(app: Application) {
...
+  app.on(
+    [
+      'pull_request.opened',
+      'pull_request.reopened',
+      'pull_request.edited',
+      'pull_request.labeled',
+    ],
+    async context => {
+      const config = await getConfig(context);
+      console.log(config);
```
kurtisvg, line 277: Should this still be here?

kurtisvg (approved):

PR #1318, opened 2021-01-25
fix(label-sync): auto-delete old buildcop labels

JustinBeckwith:


bcoe (approved):

packages/label-sync/src/label-sync.ts
```diff
@@ -281,6 +281,9 @@ async function reconcileLabels(
     'kokoro:force-ci',
     'kokoro: force-run',
     'kokoro: run',
+    'buildcop: issue',
+    'buildcop: quiet',
+    'buildcop: flaky',
```
tbpg, line 286: We should also replace these with the corresponding new label.

tbpg (changes requested):

packages/label-sync/src/label-sync.ts
JustinBeckwith: That's already done!  We added the new `flakybot: ` labels in #1306, and I manually just went through and replaced all instances of `buildcop: ` labels with the new auto created `flakybot: ` labels.

tbpg: Ah, forgot we were doing that manually. Thanks!

tbpg (approved):

PR #1345, opened 2021-02-03
fix(gcf-utils): upgrade to Probot 11, with e2e deployment test

sofisl:


packages/gcf-utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "gcf-utils",
-  "version": "7.0.0",
+  "version": "7.1.0",
```
bcoe, line 3: this will actually be updated automagically ⭐  when we merge this fix.

bcoe (changes requested):
Great digging @sofisl \o/

One small nit, automation should handle bumping the version for us I believe.

packages/gcf-utils/package.json
```diff
@@ -39,8 +39,9 @@
     "gaxios": "^4.0.0",
     "get-stream": "^6.0.0",
     "into-stream": "^6.0.0",
+    "octokit-auth-probot": "^1.2.2",
     "pino": "^6.3.2",
-    "probot": "11.0.5",
+    "probot": "^11.0.5",
```
JustinBeckwith, line 44: I would suggest us keeping probot pinned.  They have proven to be untrustworthy with semver.

sofisl: I always forget this...thanks for keeping me honest :)

packages/gcf-utils/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "gcf-utils",
-  "version": "7.0.0",
+  "version": "7.1.0",
```
bcoe, line 3:
```suggestion
  "version": "7.0.0",
```

bcoe (approved):

PR #1350, opened 2021-02-03
chore: move OwlBotLock into config-files.ts

SurferJeffAtGoogle:
I need to re-use OwlBotLock's definition in the database code, and I intend to define .OwlBot.yaml stuff here too.


packages/owl-bot/src/config-files.ts
```diff
@@ -0,0 +1,41 @@
...
+    digest: string;
+  };
+}
+
+// The default path where .OwlBot.lock.yaml is expected to be found.
+export const owlBotLockPath = '.github/.OwlBot.lock.yaml';
+
+// Throws an exception if the object does not have the necessary structure.
+// Otherwise, returns the same object as an OwlBotLock.
+export function owlBotLockFrom(o: any): OwlBotLock {
```
JustinBeckwith, line 30: Do we really need to accept `any` here?  The typing has no runtime behavior changes in TypeScript, so unless you're passing in arbitrary objects intentionally from other parts of the code, using `OwlBotLock` here should be ok.  

Side note - please treat all lint warnings as errors.  We need to fix this in `gts` to report them as errors, but if we use `any` anywhere in the code base, it needs an eslint exception (and a good reason usually)

bcoe: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/js-yaml/index.d.ts#L12

js-yaml seems to defines its type as:

`object | string | number | null | undefined`

So perhaps we could use this rather than `any`?

SurferJeffAtGoogle: I removed the `any`.  I'd like to keep this file intentionally ignorant of yaml types.

SurferJeffAtGoogle:
Ready for another look.

JustinBeckwith (approved):

PR #1358, opened 2021-02-04
fix(auto-label): second attempt upgrading to Probot 11

sofisl:


packages/auto-label/src/auto-label.ts
```diff
@@ -191,14 +199,18 @@ handler.addLabeltoRepoAndIssue = async function addLabeltoRepoAndIssue(
   return wasNotAdded;
 };
 
+function isFile(file: File | any): file is File {
```
JustinBeckwith, line 202: Please treat any gts warning as an error.  Since you're doing a cast below to `File`, and you're using the type guards, you can use `unknown` instead of `any` here without consequence :).

packages/auto-label/package.json
```diff
@@ -40,6 +41,7 @@
     "gts": "^3.0.2",
     "mocha": "^8.2.1",
     "nock": "^13.0.5",
+    "octokit-auth-probot": "^1.2.2",
```
JustinBeckwith, line 44: Why does probot like hurting us

JustinBeckwith (approved):

packages/auto-label/src/auto-label.ts
sofisl: Oh great, I was wondering how to address that. Thanks!

PR #1374, opened 2021-02-05
fix(blunderbuss): second attempt to upgrade blunderbuss to probot 11

sofisl:


JustinBeckwith (approved):

packages/blunderbuss/package.json
```diff
@@ -32,6 +32,7 @@
   },
   "devDependencies": {
     "@octokit/webhooks": "^7.24.2",
+    "@octokit/webhooks-definitions": "^3.57.2",
```
JustinBeckwith, line 35: huh - is this used anywhere?

packages/blunderbuss/src/blunderbuss.ts
```diff
@@ -150,7 +150,7 @@ export function blunderbuss(app: Probot) {
         // Don't check all labels to avoid updating an old issue when someone
         // changes a random label.
         const assigneesForNewLabel = findAssignees(byConfig, [
-          context.payload.label?.name,
```
JustinBeckwith: these do very different things!  The `?` will cause this to return `undefined` if `label` is undefined, but the `!` says "trust me, this will never be undefined or null".  Are you certain this is what you wanted?

JustinBeckwith (approved):

packages/blunderbuss/src/blunderbuss.ts
sofisl: I am expecting there is a value in label given that we've waited to enter this block only if the flag `isLabeled` is true, which means that it was kicked off if the PR or issue was labeled. But, just to be on the safe side, I included an if-block instead to confirm this value exists.

packages/blunderbuss/package.json
sofisl: I think it's needed when pulling in the `EventPayloads` type from octokit/webhooks.

PR #1385, opened 2021-02-06
fix(snippet-bot): upgrade snippet bot to Probot11

sofisl:


packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -584,11 +592,13 @@ export = (app: Application) => {
...
     const prNumber = context.payload.issue.number;
-    const prResponse = await context.github.pulls.get({
+    const prResponse = await context.octokit.pulls.get({
       owner: owner,
       repo: repo,
       pull_number: prNumber,
     });
+    // This line is giving us errors because the payload is
+    // a comment, I'm not sure how to fix this
     context.payload.pull_request = prResponse.data;
```
sofisl, line 602: @tmatsuo, this line is erroring because the payload should be a comment. I'm not sure what you're trying to do here, how can we fix this?

packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -367,14 +375,14 @@ async function scanPullRequest(
...
+      const blob = await context.octokit.repos.getContent({
         owner: context.payload.pull_request.head.repo.owner.login,
         repo: context.payload.pull_request.head.repo.name,
         path: file,
         ref: context.payload.pull_request.head.sha,
       });
-      if (blob.data.content === undefined) {
-        continue;
+      if (!isFile(blob.data)) {
+        return;
```
tmatsuo, line 385: Can we `continue` instead of `return`?

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: It is to pass the pull request data into `scanPullRequest` function.

I presume octokit now has stricter types.

Maybe you can refactor `scanPullRequest`, to accept `prResponse.data` as well?

tmatsuo:
@sofisl I made some changes to the snippet-bot and now it has a conflict. 

I'll create a new PR.

tmatsuo:
superseded by #1428

PR #1396, opened 2021-02-08
feat: implement ConfigsStore with firestore

SurferJeffAtGoogle:


SurferJeffAtGoogle:
Ready for review.

packages/owl-bot/package.json
```diff
@@ -47,11 +47,13 @@
...
     "nock": "^13.0.6",
     "sinon": "^9.2.3",
-    "typescript": "~4.1.3"
+    "typescript": "~4.1.3",
+    "uuid": "^8.3.2"
   },
   "dependencies": {
     "@google-cloud/cloudbuild": "^2.0.6",
     "@octokit/rest": "^18.1.0",
+    "firebase-admin": "^9.4.2",
```
JustinBeckwith, line 56: In most cases, we would want to use `@google-cloud/firestore` here ... but I will read on!

packages/owl-bot/src/configs-store.ts
```diff
@@ -0,0 +1,90 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {OwlBotLock, OwlBotYaml} from './config-files';
+
+export interface Configs {
+  // The body of .Owlbot.lock.yaml.
+  lock: OwlBotLock | undefined;
```
JustinBeckwith, line 19: It's a little more conventional to list these as `lock?: OwlBotLock`

packages/owl-bot/src/database.ts
```diff
@@ -0,0 +1,138 @@
...
+interface UpdateLockPr {
+  pullRequestId: string;
+}
+
+/**
+ * When firebase sees a / in a doc id, it interprets it as a collection name.
+ * So, we have to escape them.  Also escape +s because we use them for combining
+ * strings into keys.  And use a format that can be decoded by decodeURIComponent().
+ */
+export function encodeId(s: string): string {
```
JustinBeckwith, line 29: lol I ran into this too!  Straight up not a good time.

packages/owl-bot/src/configs-store.ts
```diff
@@ -0,0 +1,90 @@
...
+  yaml: OwlBotYaml | undefined;
+  // The commit hash from which the config files were retrieved.
+  commitHash: string;
+  // The branch name from which the config files were retrieved.
+  branchName: string;
+  // The installation id for our github app and this repo.
+  installationId: number;
+}
+
+export interface ConfigsStore {
```
JustinBeckwith, line 30: I don't see another implementor of this interface - is there a reason to have it as an interface instead of just using the single exported class?  Is there a potential to have multiple config stores?  Trying to avoid Java-esque INTERFACE ADAPTER FACTORY PATTERN stuff.

packages/owl-bot/src/database.ts
```diff
@@ -0,0 +1,138 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import admin from 'firebase-admin';
+import {OwlBotLock} from './config-files';
+import {Configs, ConfigsStore} from './configs-store';
+
+export type Db = admin.firestore.Firestore;
```
JustinBeckwith, line 19: I honestly have no idea which of these packages (this vs @google-cloud/firestore) we're supposed to use

packages/owl-bot/src/database.ts
```diff
@@ -0,0 +1,138 @@
...
+    this.db = db;
+    this.yamls = collectionsPrefix + 'yamls';
+    this.lock_update_prs = collectionsPrefix + 'lock-update-prs';
+  }
+
+  async getConfigs(repo: string): Promise<Configs | undefined> {
+    const docRef = this.db.collection(this.yamls).doc(encodeId(repo));
+    const doc = await docRef.get();
+    // Should we verify the data?
+    return doc.data() as Configs;
```
JustinBeckwith, line 59: I'm surprised there's no generic interface on `data<Configs>()` that lets you do this instead of a cast.  I try to avoid the casts when at all possible.

packages/owl-bot/test/database.ts
```diff
@@ -0,0 +1,159 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {describe, it, before} from 'mocha';
+import admin from 'firebase-admin';
+import {FirestoreConfigsStore, encodeId, decodeId} from '../src/database';
+import {Configs} from '../src/configs-store';
+import {v4 as uuidv4} from 'uuid';
+import * as assert from 'assert';
+import {resolve} from 'path';
```
JustinBeckwith, line 21: Let's try to clean up any of the lint warnings (here and below)

packages/owl-bot/test/database.ts
```diff
@@ -0,0 +1,159 @@
...
+    // to handle that case too.
+    const timeOutPromise = new Promise((resolve, reject) => {
+      setTimeout(() => reject('Timed out!'), 10000);
+    });
+    admin.initializeApp({
+      credential: admin.credential.applicationDefault(),
+    });
+    const db = admin.firestore();
+    try {
+      await Promise.race([
```
JustinBeckwith, line 36: Please, please make sure this is filed as a bug, and a link to the bug is included here. If we're experiencing this, so are customers.

packages/owl-bot/package.json
SurferJeffAtGoogle: I followed the instructions here:
https://firebase.google.com/docs/firestore/quickstart

packages/owl-bot/src/configs-store.ts
SurferJeffAtGoogle: Done.

packages/owl-bot/src/configs-store.ts
SurferJeffAtGoogle: This interface was useful for splitting the work between Ben and I.

packages/owl-bot/src/database.ts
SurferJeffAtGoogle: I followed the instructiosn here:
https://firebase.google.com/docs/firestore/quickstart

packages/owl-bot/src/database.ts
SurferJeffAtGoogle: I found no way to avoid the cast.

packages/owl-bot/test/database.ts
SurferJeffAtGoogle: https://github.com/googleapis/nodejs-firestore/issues/1418

packages/owl-bot/src/database.ts
```diff
@@ -0,0 +1,138 @@
...
+  pullRequestId: string;
+}
+
+/**
+ * When firebase sees a / in a doc id, it interprets it as a collection name.
+ * So, we have to escape them.  Also escape +s because we use them for combining
+ * strings into keys.  And use a format that can be decoded by decodeURIComponent().
+ */
+export function encodeId(s: string): string {
+  return s.replace('%', '%25').replace('/', '%2F').replace('+', '%2B');
```
bcoe, line 30: could you use `encodeURIComponent('foo/bar')`.

packages/owl-bot/src/database.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+function makeUpdateLockKey(repo: string, lock: OwlBotLock): string {
+  return [repo, lock.docker.image, lock.docker.digest].map(encodeId).join('+');
+}
+
+export class FirestoreConfigsStore implements ConfigsStore {
+  private db: Db;
+  readonly yamls: string;
+  readonly lock_update_prs: string;
```
bcoe, line 44: I would use camel case here, to be consistent with JavaScript coding conventions.

packages/owl-bot/test/database.ts
```diff
@@ -0,0 +1,158 @@
...
+describe('database', () => {
+  before(async function () {
+    // When checks run on github, the call to db.collection().doc().get()
+    // doesn't return an error; it never completes.  So, we need a timeout
+    // to handle that case too.
+    const timeOutPromise = new Promise((resolve, reject) => {
+      setTimeout(() => reject('Timed out!'), 10000);
+    });
+    admin.initializeApp({
+      credential: admin.credential.applicationDefault(),
```
bcoe, line 31: I find this a little strange, we don't have any credentials in our test environment do we?

packages/owl-bot/test/database.ts
```diff
@@ -0,0 +1,158 @@
...
+    });
+    const db = admin.firestore();
+    try {
+      await Promise.race([
+        timeOutPromise,
+        db.collection('hello').doc('world').get(),
+      ]);
+    } catch (e) {
+      // No connection to firestore.  Not possible to test.
+      this.skip();
```
bcoe, line 41: If I'm understanding correctly, are these tests only running locally then?

packages/owl-bot/test/database.ts
```diff
@@ -0,0 +1,158 @@
...
+      credential: admin.credential.applicationDefault(),
+    });
+    const db = admin.firestore();
+    try {
+      await Promise.race([
+        timeOutPromise,
+        db.collection('hello').doc('world').get(),
+      ]);
+    } catch (e) {
+      // No connection to firestore.  Not possible to test.
```
bcoe, line 40: Perhaps create a `TODO` comment suggesting we mock out some of the `firestore` API, so that we can test in our actions?

bcoe (approved):
LGTM with a few nits.

I think it might be worth considering mocking some of the firestore interface, so that we can test the logic in our actions -- I don't think that work needs to block this PR though.

packages/owl-bot/src/database.ts
SurferJeffAtGoogle: Yes, that's an option.  But for debugging purposes, I prefer to see strings like 'шеллы' rather than "%D1%88%D0%B5%D0%BB%D0%BB%D1%8B" in the database.

packages/owl-bot/test/database.ts
SurferJeffAtGoogle: Right.  In our test environment, we skip() the test.  It has a lot of potential to be flaky, given it's a system test that relies on a real firebase backend.

packages/owl-bot/test/database.ts
SurferJeffAtGoogle: Yep.

packages/owl-bot/src/database.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * @param collectionsPrefix should only be overridden in tests.
+   */
+  constructor(db: Db, collectionsPrefix = 'owl-bot-') {
```
bcoe, line 49: I think we need to pass in a project name somehow, since our Firestore database is on a different project.

Perhaps you could read from `process.env.FIREBASE_PROJECT`?

bcoe (changes requested):
Had a thought reading through a second time, we should make it so the project can be set with an environment variable, so that we can configure talking to the alternate instance.

packages/owl-bot/test/database.ts
SurferJeffAtGoogle: I think we have more valuable things to do with our limited time.  I wouldn't trust the results anyway.  For example, I never would have discovered the slashes-in-doc-ids issue had I not tested against the real backend.

packages/owl-bot/src/database.ts
SurferJeffAtGoogle: Done.

bcoe (approved):

packages/owl-bot/src/database.ts
bcoe: Talked out of band, this will be done in a follow up PR.

PR #1397, opened 2021-02-08
feat: add logic to open pull request with updated  .OwlBot.lock.yaml

bcoe:


packages/owl-bot/src/handlers.ts
```diff
@@ -0,0 +1,126 @@
...
+  // Opens a pull request with any files represented in changes:
+  const prNumber = await createPullRequest(
+    octokit as Octokit,
+    changes,
+    {
+      upstreamOwner: owner,
+      upstreamRepo: repo,
+      // TODO(rennie): we should provide a context aware commit
+      // message for this:
+      title: 'Update OwlBot.lock',
```
SurferJeffAtGoogle, line 103: 'chore: update OwlBot.lock with new version of post-processor'

packages/owl-bot/src/handlers.ts
```diff
@@ -0,0 +1,126 @@
...
+      branch: 'owl-bot-lock-1',
+      // TODO(bcoe): come up with a funny blurb to put in PRs.
+      description:
+        '🦉 flying is not merely some crude, mechanical process. It is a delicate art',
```
SurferJeffAtGoogle, line 107: 'Version {version} was published at {timestamp}.'

I think that requires timestamp to passed in as an additional argument.  Pubsub can provide the timestamp for when the message was published, which is close enough for our purposes.

SurferJeffAtGoogle (commented):
Looking good.  Two changes to the text of the pull request requested.

packages/owl-bot/test/handlers.ts
```diff
@@ -0,0 +1,99 @@
...
+      const lock = {
+        docker: {
+          image: 'foo-image',
+          digest: 'sha256:abc123',
+        },
+      };
+      const expectedYaml = dump(lock);
+      let recordedURI = '';
+      // Mock the database helpers used to check for/update existing PRs:
+      const fakeConfigStore = ({
```
SurferJeffAtGoogle, line 44: Suppose later, another argument is added to the method `recordPullRequestForUpdatingLock()`.

These tests will continue to compile and pass.  We won't find the bug until runtime in production.  Compare that to creating a `class FakeConfigsStore implements ConfigsStore { }` and pressing `Ctrl+.` to let VSCode implement all the methods for you.

SurferJeffAtGoogle (approved):

chingor13 (approved):

PR #1398, opened 2021-02-08
feat(redirect): add redirect for container subscription

bcoe:
Add an additional endpoint for redirecting PubSub messages from the container registry to owl-bot.

Authentication will be handled within Owl Bot itself, so there is no need for an installation token.

### DIFF
--- a/serverless-scheduler-proxy/main.go
+++ b/serverless-scheduler-proxy/main.go
@@ -53,6 +53,7 @@ func main() {
 	mux.Handle("/v0", botCronProxy(cfg))
 	mux.Handle("/v0/cron", botCronProxy(cfg))
 	mux.Handle("/v0/pubsub", botPubSubProxy(cfg))
+	mux.Handle("/v0/container", botContainerProxy(cfg))
 
 	addr := ":" + port
 	log.Printf("starting to listen on %s", addr)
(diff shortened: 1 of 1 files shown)


serverless-scheduler-proxy/main.go
```diff
@@ -99,6 +100,21 @@ func rewriteBotPubSubURL(c botConfig) func(*http.Request) {
 	}
 }
 
+func rewriteBotContainerURL(c botConfig) func(*http.Request) {
+	return func(req *http.Request) {
+		req.Header.Add("x-github-event", "pubsub.message")
+		parser := func(bodyBytes []byte) (string, string) {
+			var pay PubSubMessage
+			json.Unmarshal(bodyBytes, &pay)
```
orthros, line 108: would change to 

```
if err := json.Unmarshal(bodyBytes, &pay); err != nil {
  log.Printf("error occurred parsing pubsub message: %v\n", err)
}
```

orthros (changes requested):

serverless-scheduler-proxy/main.go
orthros: Just so that if we do ever get an error unmarshaling the bytes we can diagnose the issue

orthros (approved):

PR #1402, opened 2021-02-09
fix(snippet-bot): distinguish region tags with commit hash

### LINKED ISSUE #1283: snippet-bot's region tag check should consider commit pinning
Currently, `snippet-bot` flags PRs that delete snippets even if those snippets' include statements are pinned to a specific commit hash.

Ideally, we'd still warn the user about these - but make it clear (in the bot's responses to PRs) which tags are (or aren't) pinned.

### Example
Consider [this comment](https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1259#issuecomment-761298168) by `snippet-bot`. The cited use of `move_uploaded_file` tag is pinned to a particular branch, so removing the existing sample from the main branch will not affect the docs. (That said, it's still useful to know that unpinning that sample may break the docs!)

tmatsuo:
This PR makes the bot a little bit wiser. The bot can distinguish
region tags which are pinned to a commit hash, then gives different
message to the users.

fixes #1283


packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -158,6 +158,32 @@ function formatSampleBrowserFyi(violations: Violation[]): string {
   return formatExpandable(summary, detail);
 }
 
+/**
+ * It formats an array of Violation for informing removal of frozen region tags.
+ */
+function formatFrozenRegionTagFyi(violations: Violation[]): string {
+  let summary = 'You are about to delete the following frozen region tag';
```
bcoe, line 165: I would be tempted to pull this into a `util.ts`, so that you can test it in isolation (without all the setup required using nock).

packages/snippet-bot/test/snippet-bot.ts
```diff
@@ -542,6 +542,64 @@ describe('snippet-bot', () => {
       requests.done();
       diffRequests.done();
     });
+
+    it('gives fyi message for removing frozen region tag', async () => {
+      sandbox.restore();
```
bcoe, line 547: Given that most of the logic is added in `violations.ts` which has already been broken out into a helper, I wonder if it would be better to add some tests for `violations.ts`, vs., end to end tests using `nock`?

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: Done

packages/snippet-bot/test/snippet-bot.ts
tmatsuo: I added unit test for violations.ts.

tmatsuo:
@bcoe Thanks, PTAL

packages/snippet-bot/test/violations.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+  it('should detect removals of used region tags', async () => {
+    const snippets = require(resolve(
+      fixturesPath,
+      './violations-test-snippets'
+    ));
+    getSnippetsStub = sandbox.stub(snippetsModule, 'getSnippets');
```
bcoe, line 75: I like that we're moving towards sinon.

bcoe (approved):
Thanks for putting this work in 👏

PR #1404, opened 2021-02-10
fix: escape all the special characters in firestore ids

SurferJeffAtGoogle:
the code was just escaping just the first instance of each special character


JustinBeckwith (approved):
I ❤️  the test covering this

PR #1413, opened 2021-02-11
fix(owl-bot): fix two timing related bugs

bcoe:
There was no await during loading in gcf-utils, this meant that our async
load step would not have loaded routes yet, before requests are handled.

I also found a missing await in handlers, which could have lead to perf
issues

### DIFF
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -151,7 +151,7 @@ export class GCFBootstrapper {
       this.probot = createProbot({overrides: cfg});
     }
 
-    this.probot.load(appFn);
+    await this.probot.load(appFn);
 
     return this.probot;
   }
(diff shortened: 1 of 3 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -151,7 +151,7 @@ export class GCFBootstrapper {
       this.probot = createProbot({overrides: cfg});
     }
 
-    this.probot.load(appFn);
+    await this.probot.load(appFn);
```
bcoe, line 154: this took quite a while to find.

packages/owl-bot/src/handlers.ts
```diff
@@ -66,7 +66,7 @@ export async function onPostProcessorPublished(
         installation: configs.installationId,
       });
       // TODO(bcoe): switch updatedAt to date from PubSub payload:
-      createOnePullRequestForUpdatingLock(configsStore, octokit, repo, lock);
+      await createOnePullRequestForUpdatingLock(configsStore, octokit, repo, lock);
```
bcoe, line 69: this could have lead to hanging requests.

SurferJeffAtGoogle (approved):

chingor13 (approved):

PR #1417, opened 2021-02-11
refactor: require an installationId for scanGithubForConfigs()

SurferJeffAtGoogle:
Greatly simplifies the logic.


bcoe (approved):
You're thinking that we'll just bake the installation ID into the cron for now?

SurferJeffAtGoogle:
> You're thinking that we'll just bake the installation ID into the cron for now?

Yep.

PR #1420, opened 2021-02-11
fix(generate-bot): update generated code to compile

chingor13:
A newly generated bot does not compile (also it's e2e test fails similarly). It looks like the generated code does not match the probot/octokit versions being used.


packages/generate-bot/templates/src/{{programName}}.ts
```diff
@@ -12,26 +12,42 @@
...
-  app.on(['issues.opened', 'pull_request.opened'], async context => {
+export = (app: Probot) => {
+  app.on(['pull_request.opened'], async context => {
     const config = (await context.config(
       CONFIGURATION_FILE_PATH,
       {}
     )) as Configuration;
 
     if (
-      (context.payload.pull_request || context.payload.issue) &&
```
chingor13, line 31: Typescript was complaining about types here. In general, we probably don't want to mix these types of events in the same handler function -- I broke the issue event into a separate, nearly identical handler to demonstrate to users that you could have more than one.

JustinBeckwith (approved):

PR #1432, opened 2021-02-18
feat(m-o-g): use base branch for branch protection

sofisl:
support MOG finding branch protection for other branches


packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -464,11 +477,21 @@ handler.scanForMissingPullRequests = async function scanForMissingPullRequests(
       const ownerAndRepoArray = issue.repository_url.split('/');
       const owner = ownerAndRepoArray[ownerAndRepoArray.length - 2];
       const repo = ownerAndRepoArray[ownerAndRepoArray.length - 1];
+      let baseBranch;
+      try {
+        baseBranch = (
+          await github.pulls.get({owner, repo, pull_number: issue.number})
+        ).data.base.ref;
+      } catch (err) {
+        baseBranch = 'master';
```
bcoe, line 486: I would be tempted to drop this `try`/`catch` unless there's a known configuration where we expect the error to happen.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -442,11 +445,21 @@ handler.scanForMissingPullRequests = async function scanForMissingPullRequests(
       const ownerAndRepoArray = issue.repository_url.split('/');
       const owner = ownerAndRepoArray[ownerAndRepoArray.length - 2];
       const repo = ownerAndRepoArray[ownerAndRepoArray.length - 1];
+      let baseBranch;
+      try {
+        baseBranch = (
+          await github.pulls.get({owner, repo, pull_number: issue.number})
```
bcoe, line 451: might be worth waiting to see if there's a known configuration where this exception happens, before we protect against it.

bcoe (changes requested):
I'd rather use `try`/`catch` when we're protecting against a specific known error, and rethrow when exceptions occur that aren't the known error, e.g.,:

```
try
...
} catch (err) {
  if (err.status === 404) {
    logger.warn(`404 on ${repo}`);
  } else {
    throw err;
  }
}
```

packages/merge-on-green/src/merge-on-green.ts
tmatsuo: Probably it's better to fetch the baseBranch later in `checkForBranchProtection`.

Here is the reason.

When you rename the default branch on the github UI, github will modify existing PRs' base branch to the new default branch.

tmatsuo (commented):
I think we should fetch the `base` branch in `checkForBranchProtection`.

packages/merge-on-green/src/merge-on-green.ts
sofisl: I also like that it's less lines of code :)

tmatsuo (approved):
Looks ok to merge.

I think we still should fetch base branch and the branch protection later in the entire workflow. Maybe I'll file another issue for that.

tmatsuo (commented):
One comment. I think we should change the PR title to "base" branch to reflect the actual change.

tmatsuo (changes requested):
One more comment.

The error message when `getBranchProtection` failed contains `master` magic string.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -261,7 +267,7 @@ handler.checkForBranchProtection = async function checkForBranchProtection(
       repo,
       issue_number: prNumber,
       body:
-        "Your PR doesn't have any required checks. Please add required checks to your master branch and then re-add the label. Learn more about enabling these checks here: https://help.github.com/en/github/administering-a-repository/enabling-required-status-checks.",
+        "Your PR doesn't have any required checks. Please add required checks to your base branch and then re-add the label. Learn more about enabling these checks here: https://help.github.com/en/github/administering-a-repository/enabling-required-status-checks.",
```
tmatsuo, line 270: Can we use `{branch}` instead of `base`?

tmatsuo (approved):
@sofisl Thanks!

bcoe (approved):

PR #1435, opened 2021-02-18
feat: database functions for PRs triggered by changes to googleapis-gen

SurferJeffAtGoogle:


packages/owl-bot/test/database.ts
```diff
@@ -42,29 +47,30 @@ describe('database', () => {
...
     const store = new FirestoreConfigsStore(db, 'test-');
-    const repo = 'googleapis/' + uuidv4();
+    const repoA = 'googleapis/' + uuidv4();
+    const repoB = 'googleapis/' + uuidv4();
     const dockerImageA = uuidv4();
     const dockerImageB = uuidv4();
 
     // Confirm that the new repo and dockerImage aren't stored yet.
-    const noConfigs = await store.getConfigs(repo);
+    const noConfigs = await store.getConfigs(repoA);
```
bcoe, line 59: overtime it would be nice to pull together a mock config store for a set of smoke tests.

packages/owl-bot/test/handlers.ts
```diff
@@ -188,6 +224,24 @@ class FakeConfigStore implements ConfigsStore {
...
+    googleapisGenCommitHash: string
+  ): Promise<string | undefined> {
+    throw new Error('Method not implemented.');
+  }
+  recordPullRequestForChangedFiles(
+    repo: string,
+    googleapisGenCommitHash: string,
+    pullRequestId: string
+  ): Promise<string> {
+    throw new Error('Method not implemented.');
```
bcoe, line 238: will there be a helper that actually creates the PR with the change set?

bcoe (commented):
This structure looks good to me. One thing that jumps out at me, is we're collecting quite a few tests in the integration `test/database.ts`.

It would be nice to try to get unit tests for some of these methods.

packages/owl-bot/test/database.ts
SurferJeffAtGoogle: Agreed.

SurferJeffAtGoogle:
> This structure looks good to me. One thing that jumps out at me, is we're collecting quite a few tests in the integration `test/database.ts`.
> 
> It would be nice to try to get unit tests for some of these methods.

Tests for the test methods? Maybe I don't follow.

packages/owl-bot/test/handlers.ts
SurferJeffAtGoogle: The code to create PRs should be implemented outside ConfigsStore.

SurferJeffAtGoogle:
Ready for another look.

packages/owl-bot/src/copy-tasks-store.ts
```diff
@@ -0,0 +1,49 @@
...
+   *  findPullRequestForUpdatingLock().
+   * @returns pullRequestId, which may differ from the argument if there
+   *   already was a pull request recorded.
+   *   In that case, the caller should close the pull request they
+   *   created, to avoid annoying maintainers with duplicate pull requests.
+   */
+  recordPubsubMessageIdForCopyTask(
+    repo: string,
+    googleapisGenCommitHash: string,
+    pubsubMessageId: string
```
bcoe, line 47: Do we expect to do anything with the messageId in exceptional flows, or is it just for book keeping.

I ask because I was glancing at the API, and didn't see an option to (as an example) delete a message by an ID.

packages/owl-bot/src/database.ts
```diff
@@ -15,12 +15,18 @@
 import admin from 'firebase-admin';
 import {OwlBotLock} from './config-files';
 import {Configs, ConfigsStore} from './configs-store';
+import {IMinimatch, Minimatch} from 'minimatch';
```
bcoe, line 18: I recollect adding a few extensions to Minimatch, because in some cases it didn't quite match my mental model of globs:

https://github.com/istanbuljs/test-exclude/blob/master/index.js#L134

Might be worth eyeballing some of the helpers in that library.

packages/owl-bot/src/database.ts
```diff
@@ -135,4 +150,89 @@ export class FirestoreConfigsStore implements ConfigsStore {
...
+}
+
+// Exported for testing purposes.
+export function newMinimatchFromSource(pattern: string): IMinimatch {
+  return new Minimatch(makePatternMatchAllSubdirs(pattern), {matchBase: true});
+}
+
+function makePatternMatchAllSubdirs(pattern: string): string {
+  // Make sure pattern always ends with /**
+  if (pattern.endsWith('/**')) {
```
bcoe, line 228: looks like you're already adding some of the edge case support I mentioned.

packages/owl-bot/src/configs-store.ts
```diff
@@ -87,4 +87,14 @@ export interface ConfigsStore {
...
+
+  /**
+   * Finds repositories who list one of the changed files as a source in
+   * copy-files.
+   * @param changedFilePaths file paths in googleapis that changed.
+   *   ex: ["/google/cloud/vision/v1/vision-v1-nodejs/src/v1/image_annotator_client.ts"]
+   * @returns the list of repo names.
+   *   ex: ["googleapis/nodejs-vision", "googleapis/python-vision"]
+   */
+  findReposAffectedByFileChanges(changedFilePaths: string[]): Promise<string[]>;
```
bcoe, line 99: Am I understanding the algorithm correctly that when we get an event triggered by an update to `googleapis-gen`, we need to:

1. get a list of all paths affected by that push.
2. run this method to get repos affected.
3. dispatch a PubSub job for each repo, which contains the `repo`, and `sha` of change?

bcoe (changes requested):
Is it okay if we add the work to be done as a TODO comment again, let me know if my picture of the implementation is correct.

packages/owl-bot/src/copy-tasks-store.ts
SurferJeffAtGoogle: It's just for book keeping and debugging.

packages/owl-bot/src/database.ts
SurferJeffAtGoogle: Good to know.  We should document exactly which variant of glob we support.

packages/owl-bot/src/configs-store.ts
SurferJeffAtGoogle: Exactly.

bcoe (approved):

PR #1446, opened 2021-02-19
fix(snippet-bot): add a sentence indicating the end of violations

### LINKED ISSUE #1445: feat(snippet-bot): Clarify which items are violations and which are not
Example PR: https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1282

> Since the first one has a title ("Here is a summary of possible violations") and the second and third ones don't, it looks like the title applies to all three. It would be clearer if you had a title above the second and third sections that said something like "Here is a summary of safe code snippet deletions" or something.

> When I saw the PR, I did not intuitively understand that you meant that the second and third sections were separate from the first. I am just one user, but if one user can be confused, any user can be confused.

tmatsuo:
fixes #1445


tmatsuo:
An example UI rendering:
https://github.com/tmatsuo/python-docs-samples/pull/20

JustinBeckwith (approved):

PR #1456, opened 2021-02-25
feat: implement CopyTasksStore.filterMissingCopyTasks()

SurferJeffAtGoogle:


packages/owl-bot/src/database.ts
```diff
@@ -207,6 +207,22 @@ export class FirestoreConfigsStore implements ConfigsStore, CopyTasksStore {
     return pubsubMessageId;
   }
 
+  async filterMissingCopyTasks(
+    repos: string[],
+    googleapisGenCommitHash: string
+  ): Promise<string[]> {
+    const snapshot = this.db.collection(this.copyTasks);
+    const result: string[] = [];
+    for (const repo of repos) {
```
JustinBeckwith, line 216: It's probably fine and safe for an initial iteration, but I do want to flag a potential perf optimization.  Depending on how many repos this is running over, this is going to apply your updates serially.  It's possible this could be a lot faster using something like [p-queue](http://npm.im/p-queue) to enable 10/20/30? concurrent tasks to work through the list.  I use this pattern _a lot_.

JustinBeckwith (approved):

packages/owl-bot/src/database.ts
SurferJeffAtGoogle: Acknowledged.

SurferJeffAtGoogle: p-queue looks cool.

PR #1460, opened 2021-02-26
fix(header-checker-lint): allow comma in the copyright holder name

### LINKED ISSUE #1459: header-checker-lint: unable to match a copyright line with Google, Inc
I think we should include `,` (comma) to the regex here:
https://github.com/googleapis/repo-automation-bots/blob/54e274124ee93cdcfc7df19c35a17bbc78c1b147/packages/header-checker-lint/src/header-parser.ts#L23

tmatsuo:
fixes #1459


JustinBeckwith (changes requested):
Can we get a check-in from someone in OSPO here? @gauntface maybe?

packages/header-checker-lint/test/fixtures/google-inc.txt
```diff
@@ -0,0 +1,14 @@
+# Copyright 2021 Google, Inc.
```
gauntface, line 1: OSPO actually expects `Copyright 2021 Google LLC`

See https://opensource.google/docs/releasing/preparing/#Apache-header for an example

tmatsuo: Yeah, I know we should use `Google LLC` for any new license headers.
I updated the year to 2016 in the test fixture for avoiding confusion.

The point of this PR is to correctly parse license headers in the past. We still have many instances of license headers with the year of 2016 and Google, Inc.

Currently, the parser in this bot can not parse those license headers correctly (it detect `Google` as the license holder).

JustinBeckwith: @gauntface for code that hasn't been touched in years, when we used to have headers that said `Copyright 1996 Google, Inc.` - should we be going around, and updating those to use the new recommended header?  I guess I'm asking if we should be pro-actively chasing these down.

tmatsuo: In my understanding, we should keep `Copyright 2016 Google, Inc` for old files, but I may be wrong. 

Anyways, the parser should correctly parse the copyright holder part, regardless of whether we should keep the original holder, or we should proactively change them to `Google LLC`.

google-admin: I agree that this is a correct change for parsing.

We definitely shouldn't be changing the date/year on the header, unsure if changing from `Google, Inc` to `Google LLC` is desirable but I would ask the licensing team if it's something you encounter.

gauntface: I agree that this is a correct change for parsing.

We definitely shouldn't be changing the date/year on the header, unsure if changing from Google, Inc to Google LLC is desirable but I would ask the licensing team if it's something you encounter.

JustinBeckwith (approved):

PR #1461, opened 2021-02-27
fix: authenticate octokit instance properly when receiving payload from cloud scheduler

### LINKED ISSUE #1444: Sync repo settings failing to update settings?
https://github.com/googleapis/google-api-java-client/pull/1717#issuecomment-782274306

sofisl:
Fixes #1444, and hopefully a bunch of our other bots that use a cron job as well.

### DIFF
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -381,11 +381,21 @@ export class GCFBootstrapper {
   }
 
   // TODO: How do we still get access to this installation token?
-  async getAuthenticatedOctokit(installationId: number): Promise<Octokit> {
+  async getAuthenticatedOctokit(
+    installationId: number,
+    wrapOptions?: WrapOptions
+  ): Promise<Octokit> {
     // See: https://github.com/probot/probot/issues/1003
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const app = (this.probot as any).apps[0] as Probot;
-    return ((await app.auth(installationId)) as unknown) as Octokit;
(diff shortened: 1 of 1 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -381,11 +381,21 @@ export class GCFBootstrapper {
   }
 
   // TODO: How do we still get access to this installation token?
-  async getAuthenticatedOctokit(installationId: number): Promise<Octokit> {
+  async getAuthenticatedOctokit(
+    installationId: number,
+    wrapOptions?: WrapOptions
```
bcoe, line 386: do we have access to `wrapOptions` when we call `getAuthenticatedOctokit` wondering if we could pass it in?

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -381,11 +381,21 @@ export class GCFBootstrapper {
...
+  ): Promise<Octokit> {
     // See: https://github.com/probot/probot/issues/1003
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const app = (this.probot as any).apps[0] as Probot;
-    return ((await app.auth(installationId)) as unknown) as Octokit;
+    const cfg = await this.getProbotConfig(wrapOptions?.logging);
+    return new Octokit({
```
bcoe, line 391: I think rather than `new cfg. Octokit` because we add a couple plugins to Octokit when loading the config.

bcoe (changes requested):

bcoe (approved):

PR #1477, opened 2021-03-04
[WIP] feat(owlbot): wiring together the copy task

bcoe:
[work in progress].


packages/owl-bot/src/copy-code.ts
```diff
@@ -84,6 +88,8 @@ export async function copyCodeAndCreatePullRequest(
   const workDir = tmp.dirSync().name;
   logger.info(`Working in ${workDir}`);
 
+  // TODO(bcoe): is this clone necessary if we're already cloned the
```
bcoe, line 91: This was confusion on my part, we're cloning the repo being pushed to not the `googleapis-gen` repo.

packages/owl-bot/src/copy-code.ts
```diff
@@ -362,3 +368,65 @@ export async function copyExists(
...
+  const tasks: Array<CopyTask> = [];
+  let count = 0;
+  // Iterate over the maxWalk most recent commits to sourceRepo,
+  // checking for repos that require an update. The update required
+  // is pushed onto a stack, so that PRs are created in reverse
+  // chronological order:
+  //
+  // TODO(@bcoe): talk to rennie about the problem of commits that
+  // have no repos associated with them.
+  for await (const hash of commitsIterator(sourceRepo, octokit)) {
```
bcoe, line 403: @SurferJeffAtGoogle advantage of using API where we can, is it becomes easier to move back to something like Cloud Functions for some of our operations -- 6 of one half dozen of another if we're running in Cloud Build.

packages/owl-bot/src/core.ts
```diff
@@ -379,7 +379,9 @@ export async function getFilesModifiedBySha(path: string, sha: string) {
     // There will be a blank line between the commit message and the
     // files list, we use this as a stopping point:
     if (file === '') break;
-    files.push(file);
+    // TODO(bcoe): discuss with rennie about what we should do with leading
+    // strict regex that didn't catch leading "/".
+    files.push(file.charAt(0) === '/' ? file : `/${file}`);
```
bcoe, line 384: @SurferJeffAtGoogle noticed this bug too, any reason not to prepend it at the outset?

SurferJeffAtGoogle: Where is the outset?

packages/owl-bot/src/copy-code.ts
SurferJeffAtGoogle: SG.

packages/owl-bot/src/core.ts
bcoe: @SurferJeffAtGoogle was thinking in the helper itself we could add the `/` prefix if it's missing?

packages/owl-bot/src/copy-code.ts
SurferJeffAtGoogle: Wait, the commitsIterator needs an octokit to call the github API.  But the repo has already been cloned locally to disk, and the commit history can be observed there.  Why go back and call the github API?

PR #1483, opened 2021-03-06
feat(auto-label): check for enabled flag for config

sofisl:


packages/auto-label/src/auto-label.ts
```diff
@@ -23,6 +23,11 @@ import {Endpoints} from '@octokit/types';
 type IssueResponse = Endpoints['GET /repos/{owner}/{repo}/issues']['response'];
 type ConfigResponse = Endpoints['GET /repos/{owner}/{repo}/contents/{path}']['response'];
 
+const ignoreList = [
```
JustinBeckwith, line 26: Generally, I'd prefer we use the pattern where we have a config flag that lets users disable the bot on a per-repository basis.  That way, folks can choose on a repo by repo basis if they want this to be enabled (assuming it's on for the org).  Looks like there's already a config set up for this bot - could we add an `enabled` flag?

tbpg: +1. Another alternative is, why isn't the bot working for these repos? Can we improve the default behavior of the bot instead? What issues/PRs aren't working well?

sofisl: @JustinBeckwith  So, I think there is a config right now, but it's currently optional. If we add an `enabled` flag, this bot will stop working for a lot of repos that don't have a config and don't have this flag set up. Should we change the default behavior of enabling for the exception of repos? 

@tbpg, Reading through the conversation, it seems like we could also try a fix that ignores `build` before the colon. I just figured that, since these are microgenerator repos and not attached to a product, they would never really want to be associated with a product, and so an ignoreList seems a bit more appropriate in this case than fixing the logic. But, there is room for improvement in the logic as well, so maybe we could make both changes?

Thinking through all of these suggestions and potentially combining - what if we have repos that don't want to be included self-select, and add a config file in their repository that just has `enabled: false`?

JustinBeckwith: Oh!  I was thinking it's enabled by default, but you could add a config to *disable* it.

bcoe:
Sounds like we have an alternate implementation planned, where we add a config file for the repos we wish to turn it off for?

tbpg (approved):

tmatsuo:
Can you update the PR title too?

sofisl:
@alexander-fenster, you should be able to self-select which repos you want to ignore now by adding a config that just has `enabled: false`on it

PR #1486, opened 2021-03-08
fix(flakybot): use os.WalkDir to hopefully avoid lstat error

### LINKED ISSUE #1475: flakybot binary dies when one of the files was deleted
An example build log:
https://source.cloud.google.com/results/invocations/4ea58817-a9c1-4b41-b553-3ed0dddd0bd1/log

```
[FlakyBot] Sending logs to Flaky Bot...
[FlakyBot] See https://github.com/googleapis/repo-automation-bots/tree/master/packages/flakybot.
[FlakyBot] Error publishing logs: lstat app_engine_app/noxfile.py: no such file or directory
```

This is happening in `iap` directory. This directory has `app_engine_app` sub directory. This sub directory has its own test. The test driver is running these tests in parallel. The test driver copy noxfile.py into the target directory before running the test, and delete it after the test.

Can flakybot binary ignore this failure?

tbpg:
I'm hoping the optimization in Go 1.16 to call `lstat` less often will avoid the error in #1475. But, I'm not 100% as that was caused by a filesystem race condition.

I also added a test by faking out PubSub publishing.

Fixes #1475.

### DIFF
--- a/packages/flakybot/flakybot_test.go
+++ b/packages/flakybot/flakybot_test.go
@@ -16,13 +16,18 @@ package main
 
 import (
 	"bytes"
+	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 
+	"cloud.google.com/go/pubsub"
 	"github.com/google/go-cmp/cmp"
 )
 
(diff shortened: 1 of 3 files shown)


tmatsuo (approved):

codyoss (approved):
LGTM

packages/flakybot/flakybot_test.go
```diff
@@ -215,3 +220,57 @@ func TestJSONMarshalMessage(t *testing.T) {
...
+
+type fakePublisher struct {
+	called []string
+}
+
+func (p *fakePublisher) publish(_ context.Context, msg *pubsub.Message) (serverID string, err error) {
+	p.called = append(p.called, string(msg.Data))
+	return "", nil
+}
+
```
bcoe, line 232: I tend to toss a comment above tests that describes the test case, if I'm not using a framework that has a `describe` block.

JustinBeckwith (approved):
Can confirm this is in fact Go

bcoe (approved):
👍 with nit: about tests.

packages/flakybot/flakybot_test.go
tbpg: The test name is used for that in Go. The test tests the thing after `Test` in the function name. If you have more than one test for a given function, you can add more detail then.

PR #1494, opened 2021-03-10
fix: Windows shell interprets quote characters differently

SurferJeffAtGoogle:
Test failed on windows.  With this change, they succeed.


JustinBeckwith (approved):
bug fixes with no tests (👁️_👁️ )

PR #1496, opened 2021-03-10
fix(owlbot): prevent infinite processor loop

bcoe:
If we see more than two commits from OwlBot in a row (_initial commit + post processor_) throw an error.

I've opted for an error rather than warning, so that we can detect cycles being introduced into our containers.


packages/owl-bot/package.json
```diff
@@ -8,7 +8,7 @@
     "build/src"
   ],
   "engines": {
-    "node": ">=10.0.0"
+    "node": ">=12.10.0"
```
bcoe, line 11: `recursive: true` not introduced until `>=12.10.0`.

packages/owl-bot/src/copy-code.ts
```diff
@@ -272,7 +272,7 @@ export function copyDirs(
     deadPath = path.join(destDir, deadPath);
     if (stat(deadPath)) {
       logger.info(`rm -r ${deadPath}`);
-      fs.rmSync(deadPath, {recursive: true});
+      fs.rmdirSync(deadPath, {recursive: true});
```
bcoe, line 275: `rmSync` lands in `14.x` I believe, this was breaking for me locally. `rmdirSync` with recursive is equivalent (and not deprecated until `16.x`).

packages/owl-bot/src/core.ts
```diff
@@ -415,6 +415,38 @@ export async function* commitsIterator(
...
 
+/*
+ * Detect whether there's an update loop created by OwlBot post-processor.
+ *
+ * @param owner owner of repo.
+ * @param repo short repo name.
+ * @param prNumber PR to check for loop.
+ * @param octokit authenticated instance of octokit.
+ */
+const OWLBOT_USER = 'gcf-owl-bot[bot]';
```
bcoe, line 426: I tested this on the loop we'd introduced, this is the username reported.

packages/owl-bot/src/core.ts
```diff
@@ -415,6 +415,38 @@ export async function* commitsIterator(
...
+async function hasOwlBotLoop(
+  owner: string,
+  repo: string,
+  prNumber: number,
+  octokit: Octokit
+): Promise<boolean> {
+  const circuitBreaker = 3;
```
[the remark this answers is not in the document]
bcoe: done.

SurferJeffAtGoogle (approved):

PR #1514, opened 2021-03-15
feat: add label to PRs created by owl-bot copies

SurferJeffAtGoogle:
So we can easily observe them an collect statistics.


packages/owl-bot/src/copy-code.ts
```diff
@@ -161,6 +161,12 @@ ${err}`,
     base: githubRepo.data.default_branch,
   });
   logger.info(`Created pull request ${pull.data.html_url}`);
+  await octokit.issues.update({
```
bcoe, line 164: one thought, we can look at the user that opened the pull request. Does the label give us benefits above and beyond this? (_what I don't like about labels is that they can be ephemeral_).

SurferJeffAtGoogle: We can look at the user that opened the pull request, but
Does that user open pull requests of other types?
Will we update the system at some point so that user also creates pull requests of other types?

bcoe (approved):

bcoe:
This is good to go, but we need to update the permissions in owl-bot.

PR #1522, opened 2021-03-17
feat: auto-approve bot initial stub

sofisl:
Initial stub for auto-approve bot. See internal PRD on description.


packages/auto-approve/src/check-config.ts
```diff
@@ -0,0 +1,151 @@
...
+  }
+}
+
+/**
+ * Takes in the auto-approve.yml file and checks to see that the schema matches the valid-pr-schema.json rules
+ *
+ * @param configYaml the string of auto-approve.yml
+ * @returns true if it matches valid-pr-schema.json, the error messages if it doesn't match
+ */
+export async function validateSchema(
```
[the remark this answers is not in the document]
bcoe: I agree 👍  I'd be tempted to throw an Error instead, which could have the additional contextual information (which you can catch at the calling function).

Then I'd just make this resolve `Boolean` 😄

packages/auto-approve/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "probot-app"
+  ],
+  "files": [
+    "build/src"
+  ],
+  "scripts": {
+    "compile": "tsc",
+    "start": "probot run ./build/src/auto-approve",
+    "pretest": "npm run compile",
+    "test": "cross-env LOG_LEVEL=fatal c8 mocha --exit build/test",
```
bcoe, line 24: Are you setting the LOG_LEVEL to fatal in tests so that the console output is quieter?

My only concern would be that it might sometimes make it difficult to debug. If an error is catchable, I would be tempted to use `console.warn`, rather than `console.error`, and on actual errors throw (_which you could catch in tests._).

With this approach you could avoid setting `LOG_LEVEL`.

packages/auto-approve/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "fix": "gts fix",
+    "lint": "gts check"
+  },
+  "dependencies": {
+    "@octokit/openapi-types": "^5.3.2",
+    "ajv": "^7.1.1",
+    "gcf-utils": "^7.1.3",
+    "jsonschema": "^1.4.0"
```
bcoe, line 32: I don't think you need this dependency, since you're using `ajv` for schema validation?

packages/auto-approve/src/auto-approve.ts
```diff
@@ -0,0 +1,226 @@
...
+ * and checks both files to ensure they match schema, format, and appropriate CODEOWNERS
+ *
+ * @param owner owner of the repo of the incoming PR
+ * @param repo string, the name of the repo of the incoming PR
+ * @param codeOwnersFile string, the CODEOWNERS file in .github/CODEOWNERS for the repo of the incoming PR
+ * @param octokit the octokit instance
+ * @param headSha the sha upon which to check whether the config and the CODEOWNERS file are configured correctly
+ * @returns true if the status check passed, false otherwise
+ */
+export async function logicForConfigCheck(
```
bcoe, line 40: I'd probably use a more specific verb here, like `validateAutoApproveFile`, I'd probably also break it into two methods:

```
validateAutoApproveFile()
validateCodeOwnersFile()
```

packages/auto-approve/src/auto-approve.ts
```diff
@@ -0,0 +1,226 @@
...
+        typeof isCodeOwnersCorrect === 'string' ? isCodeOwnersCorrect : ''
+      }\n` +
+      `${typeof isYamlValid === 'string' ? isYamlValid : ''}\n` +
+      `${
+        typeof isSchemaValid === 'object'
+          ? JSON.stringify(isSchemaValid, null, 2)
+          : ''
+      }`;
+
+    await octokit.checks.create({
```
bcoe, line 97: Along with splitting out into specific validation methods, I'd probably make the actual adding a `check` happen outside of the validation, that way you can perform validation in isolation without a side effect.

packages/auto-approve/src/check-pr.ts
```diff
@@ -0,0 +1,119 @@
...
+  pr: PullRequestEvent,
+  octokit: InstanceType<typeof ProbotOctokit>
+): Promise<Boolean> {
+  const repoOwner = pr.pull_request.head.repo.owner.login;
+  const prAuthor = pr.pull_request.user.login;
+  const repo = pr.pull_request.head.repo.name;
+  const prNumber = pr.number;
+  const title = pr.pull_request.title;
+
+  const validTypeOfPR = config.rules.find(x => x.author === prAuthor);
```
bcoe, line 51: Is this grabbing the first rule that is associated with the author of the PR?

Might be worth adding a comment that this assumes there will only one rule per PR author?

packages/auto-approve/src/check-pr.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  if (validTypeOfPR) {
```
bcoe, line 53: perhaps call this `rulesToValidateAgainst`? took me a second to realize that this represented rules for merging a PR.

packages/auto-approve/src/check-pr.ts
```diff
@@ -0,0 +1,119 @@
...
+  octokit: InstanceType<typeof ProbotOctokit>
+): Promise<Boolean> {
+  const repoOwner = pr.pull_request.head.repo.owner.login;
+  const prAuthor = pr.pull_request.user.login;
+  const repo = pr.pull_request.head.repo.name;
+  const prNumber = pr.number;
+  const title = pr.pull_request.title;
+
+  const validTypeOfPR = config.rules.find(x => x.author === prAuthor);
+
```
bcoe, line 52: I like that we don't even try to validate the PR if the author doesn't match 👍

packages/auto-approve/src/auto-approve.ts
```diff
@@ -0,0 +1,226 @@
...
+
+        // Check to see whether the incoming PR matches the incoming PR
+        const isPRValid = await checkPRAgainstConfig(
+          config,
+          context.payload,
+          context.octokit
+        );
+
+        // If both PR and config are valid, pull in approving-mechanism to tag and approve PR
+        if (isPRValid && isConfigValid) {
```
bcoe, line 202: I would be explicit here;

```
ifPrValid === true && isConfigValid === true
```

Otherwise if we ever forget an `await`, we might start merging all pull requests, since promises will evaluate as true:

```js
> if (Promise.resolve(false)) console.info('hello')
hello
```

vs.,

```
> if (Promise.resolve(false) === true) console.info('hello')
undefined
```

bcoe (commented):
Haven't gotten to reviewing the tests yet, but left some initial comments. This is looking great to me 👍

packages/auto-approve/src/auto-approve.ts
sofisl: Not sure if you've taken a look at src/check-config.ts, which actually does the heavy lifting and is broken up into three separate methods (`validateYaml`, `validateSchema`, and `checkCodeOwners`). This is just a wrapper function that makes sure that all three are true before submitting a passing check, else it submits a failing one. Beyond that, do you still feel that we should break up this function?

packages/auto-approve/src/check-config.ts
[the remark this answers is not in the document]
sofisl: Sounds good! Thanks for the thoughts :)

packages/auto-approve/src/auto-approve.ts
sofisl: Same comment as above... let me know if you still think it's worth further dividing!

packages/auto-approve/src/check-pr.ts
sofisl: Good catch. I've been thinking though, this configuration would allow for multiple rules per author. For example, in the future, we could have something like 
```
rules:
- author: "googleapis-publisher"
  title: 
        -"^chore: autogenerated discovery document update"
        -"^XYZ$"
```

And then just add further logic to check which author and title combination we're checking for. 

What's cool about the JSON schema validation process is that we can add further conditional rules to make sure that, for example, only certain files are changed for combination one vs. combination 2, etc.

packages/auto-approve/src/auto-approve.ts
sofisl: Nice catch! Thank you!!

packages/auto-approve/src/check-config.ts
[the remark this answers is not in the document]
sofisl: So, I considered returning an error but that seemed to be a lot of lines of code... so instead I'm returning a standardized validation object, that looks like this: 

```
export interface validationResult {
  checkType: 'Yaml' | 'Schema' | 'Codeowners';
  isValid: boolean;
  errorMessages?: ErrorMessage[];
  message?: string | undefined;
}
```
Feel free to advocate for an error though, if I'm missing something!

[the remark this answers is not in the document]
SurferJeffAtGoogle: `validationResult` still makes it complicated to call this function.  To understand the the result, I have to find the definition of this interface, and then write code that inspects the 4 members of the result.

I propose the following function interface:

```js
/**
 * Takes in the auto-approve.yml file and checks to see that it is formatted correctly
 *
 * @param configYaml the string of auto-approve.yml
 * @returns empty if the yaml is valid, otherwise an error message.
 */
export function validateYaml(configYaml: string): string;
```

With this declaration, it's very easy to understand the return value and hard to get the logic wrong.

packages/auto-approve/test/check-config.test.ts
```diff
@@ -0,0 +1,300 @@
...
+        {
+          wrongProperty: {allowedValue: '^chore: regenerate README$'},
+          message: 'should be equal to constant',
+        },
+      ]);
+    });
+
+    it('should fail if title does not match second author', async () => {
+      //title does not correspond to author
+      assert.deepStrictEqual((await invalidateSchema(3)).errorMessages, [
```
SurferJeffAtGoogle, line 103: Asserting that the error text exactly match given error text makes for a very fragile test.  If I need to fix a typo in the error message later, I then need to fix tests too.

There's always a trade off between writing tests that confirm every detail and writing tests that are easy to maintain.

A less fragile test confirms the function returns *any* error message for the given test.  That's still going to confirm bad input is reported, but be smaller burden to maintain.

packages/auto-approve/test/check-config.test.ts
```diff
@@ -0,0 +1,300 @@
...
+    yaml.load(
+      fs.readFileSync(
+        `./test/fixtures/config/invalid-schemas/invalid-schema${configNum}.yml`,
+        'utf8'
+      )
+    )
+  );
+}
+
+describe('check for config', () => {
```
SurferJeffAtGoogle, line 54: Kudos for being thorough.  With all these tests, I'm very confident this code will behave as expected.

packages/auto-approve/src/get-PR-info.ts
```diff
@@ -0,0 +1,80 @@
...
+    await octokit.pulls.listFiles({
+      owner,
+      repo,
+      pull_number: prNumber,
+    })
+  ).data;
+}
+
+/**
+ * Gets the blob of a file from an array of changed files from a PR
```
SurferJeffAtGoogle, line 49: Thank you for writing per-function comments like this.  They help the reader (me) so much.

packages/auto-approve/src/check-config.ts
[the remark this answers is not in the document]
sofisl: Awesome, thanks for the comments! I added this in.

packages/auto-approve/test/check-config.test.ts
sofisl: I thought a bit about this suggestion, and I implemented what I think might be halfway point: https://www.npmjs.com/package/npm-snapshot

I'm using this library to confirm that the error message is the same as it was last time. It's less work to change in the future (you just have to delete `snapshots` and then rerun tests), but it can provide some more specific checking.

packages/auto-approve/src/get-PR-info.ts
sofisl: Thanks for appreciating :)

packages/auto-approve/src/auto-approve.ts
bcoe: Not a blocker; I think what jumped out at me is that `logic` is a little vague, it could mean *anything*.

I think the term `handle` might resonate with me more, because I'm used to the term handler being used to refer to something that performs the logic for a route -- I realize now that I type this that handler is pretty much as vague as "logic for" 😛

packages/auto-approve/src/check-config.ts
```diff
@@ -0,0 +1,153 @@
...
+  configYaml: string | undefined | null | number | object
+): Promise<ErrorMessage[] | undefined> {
+  const validateSchema = await ajv.compile(schema);
+  await validateSchema(configYaml);
+  const errorText = (await validateSchema).errors?.map(x => {
+    return {wrongProperty: x.params, message: x.message};
+  });
+  console.log(errorText);
```
bcoe, line 81: probably a left over log line.

packages/auto-approve/test/auto-approve.test.ts
```diff
@@ -0,0 +1,268 @@
...
+function submitReview(status: number) {
+  return nock('https://api.github.com')
+    .post('/repos/testOwner/testRepo/pulls/1/reviews/1/events', body => {
+      snapshot(body);
+      return true;
+    })
+    .reply(status);
+}
+
+function addLabels(status: number) {
```
bcoe, line 59: I like how you've encapsulated each mock, this is great 👍

bcoe (approved):
This is looking really solid to me, great work 👍

packages/auto-approve/src/check-config.ts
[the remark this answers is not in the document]
SurferJeffAtGoogle: I still see validateYaml returning `string | undefined` on line 53.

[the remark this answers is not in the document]
sofisl: I just made the changes we talked about last week :)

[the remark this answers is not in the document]
SurferJeffAtGoogle: I see.

SurferJeffAtGoogle (approved):

PR #1534, opened 2021-03-19
fix(owlbot): address rate limit issues

bcoe:
We bumped into a few rate limit issues when we bootstrapped OwlBot across dozens of repos the first time:

1. GitHub's abuse detection algorithm was triggered when opening PRs for updating all docker containers in sequence (I've added a delay to hopefully make this less likely).
2. We were hitting the Concurrent Get requests on Cloud Build (I've increased our polling interval to hopefully reduce this issue).
3. we were retrying almost immediately upon receiving an exception in Cloud Tasks, this lead to odd behavior where multiple jobs attempted to concurrently create the same PR, let's wait 5s making it less likely that this thundering herd happens.

---

Unsolved Problem:

* we're very close to the limit on Cloud Build for creating a maximum number of concurrent builds (we're allowed to queue 60 in a minute).
* perhaps less of an issue, we're hitting the 10 job concurrent job limit -- as I understand it the fact that jobs queue will protect us from this.

### DIFF
--- a/packages/owl-bot/src/core.ts
+++ b/packages/owl-bot/src/core.ts
@@ -156,7 +156,7 @@ async function waitForBuild(
   id: string,
   client: CloudBuildClient
 ) {
-  for (let i = 0; i < 120; i++) {
+  for (let i = 0; i < 60; i++) {
     const [build] = await client.getBuild({projectId, id});
     if (build.status !== 'WORKING' && build.status !== 'QUEUED') {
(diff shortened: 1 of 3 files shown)


JustinBeckwith:
On the Cloud Build bit - we can totally file an issue to ask for a quota bump (and you should)

JustinBeckwith (approved):

PR #1541, opened 2021-03-22
feat(owlbot): close empty PRs

bcoe:
If running the OwlBot Post Processor results in a pull request with no changes, automatically close the pull request.

### DIFF
--- a/packages/owl-bot/src/owl-bot.ts
+++ b/packages/owl-bot/src/owl-bot.ts
@@ -104,7 +104,7 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
       // Fetch the .Owlbot.lock.yaml from the head ref:
       const lock = await core.getOwlBotLock(
         head.repo.full_name,
-        context.payload.number,
+        context.payload.pull_request.number,
         context.octokit
       );
       if (!lock) {
@@ -121,7 +121,7 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
           appId,
           installation,
           repo: head.repo.full_name,
-          pr: context.payload.number,
+          pr: context.payload.pull_request.number,
           trigger,
         },
         context.octokit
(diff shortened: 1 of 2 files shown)


packages/owl-bot/src/owl-bot.ts
```diff
@@ -104,7 +104,7 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
       // Fetch the .Owlbot.lock.yaml from the head ref:
       const lock = await core.getOwlBotLock(
         head.repo.full_name,
-        context.payload.number,
+        context.payload.pull_request.number,
```
bcoe, line 107: `context.payload.number` was an alias for `context.payload.pull_request.number`, I've moved us to consistently using `context.payload.pull_request.number` (since using both was confusing in tests).

packages/owl-bot/src/owl-bot.ts
```diff
@@ -141,6 +141,23 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
...
       );
+      // If running post-processor has created a noop change, close the
+      // pull request:
+      const files = (
+        await context.octokit.pulls.listFiles({
+          owner,
+          repo,
+          pull_number: context.payload.pull_request.number,
+        })
+      ).data;
```
alexander-fenster, line 152: One concern is the metadata synth metadata change that often is the only change in the PR. Will that still trigger a PR?

alexander-fenster (approved):

packages/owl-bot/src/owl-bot.ts
bcoe: @SurferJeffAtGoogle mind speaking to this? We don't believe we actually maintain a synth metadata any longer. Instead  we track what commits haven't yet been applied from `googleapis-gen`.

chingor13 (approved):

packages/owl-bot/src/owl-bot.ts
alexander-fenster: If it goes away, that's great and resolves the concern :)

Issue #1540, opened 2021-03-22
Owlbot looks too far into past when creating pull requests.


SurferJeffAtGoogle:
On Fri, Mar 19, 2021 at 12:04 PM Sofia Leon <sofialeon@google.com> wrote:

> Hi Jeff & Ben,
> During triage today I ran across some owl bot PRs that had specific titles but the files changed did not reflect the title. Moreover, the change came up in the individual commits, but then wasn't reflected in files changed. See:https://github.com/googleapis/nodejs-analytics-data/pull/78/filesLooked at this one.
> 

I see this change upstream in googleapis-gen: 

Why wasn't this change reflected in https://github.com/googleapis/nodejs-analytics-data/pull/78/files?
Because beta_analytics_data_client.ts already had that change:
 

Hmmm.  I thought adding this commit to the "migrate to owlbot" PR would have prevented Owlbot from looking so far back in time:https://github.com/googleapis/nodejs-analytics-data/pull/74/commits/9364cafdc0009d34f8f3947c1b480173c357d628

However, I now see why that's not the case.

PR #1548, opened 2021-03-22
fix: limit how far owl bot looks into the past with begin-after-commi…

SurferJeffAtGoogle:
…t-hash

Fixes https://github.com/googleapis/repo-automation-bots/issues/1540

### DIFF
--- a/packages/owl-bot/src/scan-googleapis-gen-and-create-pull-requests.ts
+++ b/packages/owl-bot/src/scan-googleapis-gen-and-create-pull-requests.ts
@@ -46,14 +46,17 @@ export async function scanGoogleapisGenAndCreatePullRequests(
   const cmd = newCmd(logger);
   const stdout = cmd(`git log -${cloneDepth} --format=%H`, {cwd: sourceDir});
   const text = stdout.toString('utf8');
-  const commitHashes = text.split(/\r?\n/).filter(x => x);
+  const commitHashes = text
+    .split(/\r?\n/)
+    .map(s => s.trim())
+    .filter(x => x);
 
   const todoStack: Todo[] = [];
   let octokit: null | OctokitType = null;
 
   // Search the commit history for commits that still need to be copied
   // to destination repos.
-  for (const commitHash of commitHashes) {
+  for (const [commitIndex, commitHash] of commitHashes.entries()) {
     const commitText = cmd(`git log -1 --pretty=oneline ${commitHash}`, {
       cwd: sourceDir,
     }).toString('utf8');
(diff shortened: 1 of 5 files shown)


sofisl (approved):

packages/owl-bot/src/scan-googleapis-gen-and-create-pull-requests.ts
```diff
@@ -71,7 +74,20 @@ export async function scanGoogleapisGenAndCreatePullRequests(
     const stackSize = todoStack.length;
     for (const repo of repos) {
       octokit = octokit ?? (await octokitFactory.getShortLivedOctokit());
-      if (!(await copyExists(octokit, repo, commitHash, logger))) {
+      // Compare the config's begin-after-commit-hash and this commit hash,
+      // and don't create a pull request if this commit hash is older.
+      const repoString = repo.toString();
```
bcoe, line 79: I'd be tempted to encapsulate this bit in a private helper method.

packages/owl-bot/test/scan-googleapis-gen-and-create-pull-requests.ts
```diff
@@ -25,191 +25,195 @@ import path from 'path';
...
+  beforeEach(() => {
+    cmd('git checkout main', {cwd: abcRepo});
+  });
+
+  it('does nothing with zero repos affected', async () => {
     assert.strictEqual(
       await scanGoogleapisGenAndCreatePullRequests(
         abcRepo,
-        factory(octokit),
+        {} as OctokitFactory,
```
bcoe, line 205: I think we could potentially use `sinon.stub(OctokitFactory)`, if we wanted to assert that octokit is actually executed? I'm surprised the empty object being passed in does cause anything to throw?

bcoe (changes requested):
One nit about encapsulating the new short-circuit logic in a helper method.

packages/owl-bot/test/scan-googleapis-gen-and-create-pull-requests.ts
SurferJeffAtGoogle: Octokit is never executed.  If it were executed, this test would fail; that's by design.

`scanGoogleapisGenAndCreatePullRequests()` calls `configsStore.findReposAffectedByFileChanges()` which returns 0 repos.  If no repos are affected, then it has nothing to do, and no reason to call Octokit.

packages/owl-bot/src/scan-googleapis-gen-and-create-pull-requests.ts
SurferJeffAtGoogle: Done.

bcoe (approved):

Issue #1531, opened 2021-03-18
Setup npm releases for release-brancher


chingor13:
Set up this repo for multi-module releases

- [x] rename previous releases to be gcf-utils <version>
- [ ] set up release builds for release-brancher

chingor13:
Manually pushed a tag to gcf-utils-v7.1.3 at v7.1.3 which should fix bootstrapping release-please for multi-module handling.

PR #1549, opened 2021-03-22
feat(release-please): allow configuring manifest releases

chingor13:
Towards #1531 

Upgrades release-please to 11.3.0


packages/release-please/src/release-please.ts
```diff
@@ -52,6 +53,7 @@ interface BranchOptions {
   bumpMinorPreMajor?: boolean;
   path?: string;
   changelogPath?: string;
+  manifest?: boolean;
```
chingor13, line 56: I'm not particularly a fan of this configuration. I would prefer that the manifest releaser be a `ReleaseType`, but that's not available in release-please -- you have to use a specific factory method. If we know that `manifest` will be a release type, then I'd prefer handling this by adding a `manifest` enum option to `releaseType` and special case handling it to use the new factory until we clean up how to build a manifest PR.

chingor13:
The alternate configuration I'd propose is to have `manifest` as a release type.

Example `.github/release-please.yaml`:
```yaml
releaseType: manifest
handleGHRelease: true
```

packages/release-please/test/release-please.ts
```diff
@@ -263,6 +264,51 @@ describe('ReleasePleaseBot', () => {
...
+
+        await probot.receive({name: 'push', payload, id: 'abc123'});
+        requests.done();
+        assert(executed, 'should have executed the runner');
+      });
+
+      it('should handle GitHub releases, if configured', async () => {
+        let runnerExecuted = false;
+        let releaserExecuted = false;
+        Runner.manifest = async (manifest: Manifest) => {
```
bcoe, line 291: You could use sinon and replace this with something like:

```
const stub = sinon.stub(Runner, 'manifest');
sinon.assert.calledOnce(stub)
```

That way we get rid of the complained about `Manifest` being unused.

bcoe (approved):
Non-blocking feedback, that I think it would be a bit nicer to use `sinon.stub`.

packages/release-please/test/release-please.ts
chingor13: Will do in a follow-up PR.

PR #1557, opened 2021-03-24
fix(release-brancher): fix build configuration and set private to false

bcoe:
This token was causing a build failure, and does not appear to be used.

### DIFF
--- a/.kokoro/common.cfg
+++ b/.kokoro/common.cfg
@@ -11,7 +11,7 @@ action {
 gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/trampoline"
 
 # Use the trampoline script to run in docker.
-build_file: "nodejs-secret-manager/.kokoro/trampoline_v2.sh"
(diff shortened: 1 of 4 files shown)


.kokoro/common.cfg
```diff
@@ -11,7 +11,7 @@ action {
 gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/trampoline"
 
 # Use the trampoline script to run in docker.
-build_file: "repo-automation-bots/.kokoro/trampoline.sh"
+build_file: "repo-automation-bots/.kokoro/trampoline_v2.sh"
```
chingor13, line 14: Right now, `trampoline.sh` is running the `populate-secrets.sh` then calling `trampoline_v2.sh`

chingor13 (changes requested):

.kokoro/common.cfg
chingor13: Oh, I see, `.trampolinerc` is calling `populate-secrets.sh`

chingor13 (approved):

PR #1566, opened 2021-03-25
feat: start tracing request with opentelemetry

### LINKED ISSUE #1031: FR(gcf-utils): enable cloud trace for bots
It would be nice to add additional observability for our GCF bots.

chingor13:
Fixes #1031


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -307,25 +336,35 @@ export class GCFBootstrapper {
...
+          const span = tracer.startSpan('enqueue task');
+          try {
+            await this.enqueueTask({
+              id,
+              name,
+              signature,
+              body: JSON.stringify(request.body),
+            });
+          } finally {
+            span.end();
```
bcoe, line 366: is `span.end` synchronous? wondering if there's a way to await all spans flushing?

chingor13: This just marks the span as ended (writes the end timestamp). Completed spans are published when the publisher is force flushed.

chingor13: I'm not a fan of these try/finally blocks - most other OpenCensus/OpenTelemetry implementations use callable closures to ensure the spans are closed - but I see how it's an issue with the `async` functions.

bcoe:
If you want to experiment with this, I'm supportive.

PR #1572, opened 2021-03-25
feat: gaurantee files are copied on order listed in .OwlBot.yaml

SurferJeffAtGoogle:
This will allow us to avoid using staging directories and just specify the default version last!


bcoe (commented):
I only see tests?

SurferJeffAtGoogle:
> I only see tests?

Indeed.  The code happened to behave this way already, but it was not defined or guaranteed.  Adding a test that fails if the behavior changes guarantees it.

bcoe (approved):

PR #1576, opened 2021-03-26
fix(policy): use uuid for branch names

JustinBeckwith:
Otherwise we keep stomping on the same branch, and ideally we will use this code for other PRs as well.  Also, it was causing failures in the case that the branch name already existed, but was behind the default branch

### DIFF
--- a/packages/policy/src/changer.ts
+++ b/packages/policy/src/changer.ts
@@ -15,6 +15,7 @@
 import {Octokit} from '@octokit/rest';
 import {createPullRequest, Changes} from 'code-suggester';
 import {request} from 'gaxios';
+import {v4 as uuid} from 'uuid';
 import {PolicyResult} from './policy';
 
 export const cocUrl =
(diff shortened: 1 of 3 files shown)


packages/policy/src/changer.ts
```diff
@@ -77,6 +78,7 @@ export async function addCodeOfConduct(
     upstreamOwner: owner,
     upstreamRepo: repo,
     fork: false,
+    branch: `policy-bot-${uuid()}`,
```
bcoe, line 81: One thing that comes to mind, is we'll perhaps want a bot pretty fast that cleans up dead branches if we're using this as a pattern.

bcoe (approved):

packages/policy/src/changer.ts
JustinBeckwith: Good news here is that merging the PR works the same as any other contribution.  The "delete branch after merge" flag is one we should broadly enable by default for all our repos I think.

Issue #1575, opened 2021-03-25
Running owlbot-cli copy-code should not create a commit on the repo.


SurferJeffAtGoogle:
Instead, it should just leave the modified files in the working directory.

PR #1582, opened 2021-03-29
fix: don't commit changes when user runs owl-bot copy-code

SurferJeffAtGoogle:
Fixes https://github.com/googleapis/repo-automation-bots/issues/1575

### DIFF
--- a/packages/owl-bot/src/copy-code.ts
+++ b/packages/owl-bot/src/copy-code.ts
@@ -118,7 +118,7 @@ ${err}`,
     logger.error(`Created issue ${issue.data.html_url}`);
     return; // Success because we don't want to retry.
   }
-  sourceRepoCommitHash = await copyCode(
+  const {sourceCommitHash, commitMsgPath} = await copyCode(
     sourceRepo,
     sourceRepoCommitHash,
     destDir,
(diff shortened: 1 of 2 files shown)


packages/owl-bot/src/copy-code.ts
```diff
@@ -118,20 +118,22 @@ ${err}`,
...
+  const {sourceCommitHash, commitMsgPath} = await copyCode(
     sourceRepo,
     sourceRepoCommitHash,
     destDir,
     workDir,
     yaml,
     logger
   );
+  cmd('git add -A', {cwd: destDir});
+  cmd(`git commit -F "${commitMsgPath}" --allow-empty`, {cwd: destDir});
```
chingor13, line 130: I'm not following how this does what the title of this PR says (not committing changes).

chingor13 (changes requested):

packages/owl-bot/src/copy-code.ts
SurferJeffAtGoogle: These git commands were moved from the `copyCode()` function into the `copyCodeAndCreatePullRequest()` function.

The [copy-code command](https://github.com/googleapis/repo-automation-bots/blob/master/packages/owl-bot/src/bin/commands/copy-code.ts) invokes `copyCode()` directly, so it will no longer execute the git commands.  The hourly cron job invokes `copyCodeAndCreatePullRequest()` so it will continue committing changes and creating pull requests.

chingor13 (approved):

PR #1588, opened 2021-03-30
feat(policy): send pull requests to add SECURITY.md

broady:


packages/policy/src/changer.ts
```diff
@@ -38,6 +41,21 @@ export async function getCoC() {
...
+ * Fetch the security policy from googleapis/.github.  Cache it.
+ * @returns Promise with the text of our template security policy
+ */
+export async function getSecurity() {
+  if (!securityContents) {
+    const res = await request<string>({
+      url: securityUrl,
+      responseType: 'text',
+    });
+    cocContents = res.data;
```
JustinBeckwith, line 54: Looks like this is touching `cocContents`, not `securityContents`

packages/policy/src/changer.ts
```diff
@@ -48,6 +66,50 @@ export async function addCodeOfConduct(
...
+  // first, make sure there's no open PR for this
+  const title = 'chore: add SECURITY.md';
+  const prs = await octokit.search.issuesAndPullRequests({
+    q: `repo:${owner}/${repo} "${title}" in:title is:open`,
+  });
+  if (prs.data.total_count > 0) {
+    return;
+  }
+
+  // fetch the CoC from `googleapis/.github`
```
JustinBeckwith, line 79: Another `CoC`

JustinBeckwith (commented):
Thanks!  Could I trouble you to add a test to `test/test.changer.ts`?  Should look mostly the same as the one that's already there.

JustinBeckwith:
Tagging this as do-not-merge until we work out the problems with code-suggester

broady:
Should I make the URL fetching stuff (and maybe some of the addX function more generic?

JustinBeckwith:
Yeah, if you're up for it, that would be super nice.  We could keep a `Map` as a cache, and just make all requests go through one function.

packages/policy/src/changer.ts
```diff
@@ -76,8 +87,48 @@ export async function addCodeOfConduct(repo: GitHubRepo, octokit: Octokit) {
...
+  const [owner, name] = repo.full_name.split('/');
+  await createPullRequest(octokit, changes, {
+    title,
+    message: title,
+    description: 'add a security policy',
+    upstreamOwner: owner,
+    upstreamRepo: name,
+    fork: false,
     primary: repo.default_branch,
-    retry: 0,
```
JustinBeckwith, line 80: It's super important to make sure both calls to `createPullRequest` use `retry: 0` in the options

broady: oops. thought it was an old thing, but i needed to `npm i` when i rebased. fixed

JustinBeckwith (approved):

JustinBeckwith:
Looks great!  Let's maybe wait for monday to land this, since it's going to create a ton of pull requests.

PR #1613, opened 2021-04-05
feat: create a second pull request to the new branch duplicating ci config

chingor13:


bcoe (approved):
Unrelated to this PR, what does the workflow look like for actually back-porting to the branch once created? Will we want a CLI tool for this cherry picking process too.

chingor13:
> Unrelated to this PR, what does the workflow look like for actually back-porting to the branch once created? Will we want a CLI tool for this cherry picking process too.

We expect this to not happen very often and when it does, it's likely not feasible to script it. `git cherry-pick` for most fixes or possibly direct PRs from other automation tools (e.g. dependency updates).

PR #1621, opened 2021-04-08
feat(owlbot): support owlbot:run label

bcoe:
Allow an `owlbot:run` label to be added, similar to `kokoro:run`, which will allow us to run the OwlBot Post-Processor against a fork, or to re-run without pushing (in the case of a failure).

When a label event happens:

1. it will run post-processor if label is `owlbot:run`, and removes it after.
2. it will not run post-processor if it is **not** a fork.
3. it will not run post-processor on fork for labels other than `owlbot:run`, e.g., `cla:yes`.


packages/owl-bot/src/owl-bot.ts
```diff
@@ -76,15 +78,29 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
...
         logger.info(
           `head ${head.repo.full_name} does not match base ${base.repo.full_name} skipping`
         );
         return;
       }
+      // If the pull request is *not* from a fork, ignore label being added
+      // (we will be running OwlBot post processor for the push itself);
+      if (
+        head.repo.full_name === base.repo.full_name &&
+        context.payload.action === 'labeled'
```
chingor13, line 97: Suggestion: if there's separate validation based on the event type, use separate probot `app.on()` handlers and move the common behavior to a helper function.

packages/owl-bot/src/owl-bot.ts
```diff
@@ -76,15 +78,29 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
       if (!installation) {
         throw Error(`no installation token found for ${head.repo.full_name}`);
       }
+      const hasRunLabel = !!context.payload.pull_request.labels.filter(
+        l => l.name === OWLBOT_RUN_LABEL
+      ).length;
```
chingor13, line 83: If it's a label event and this label is added, do we want to remove it afterwards (like release-please, kokoro, etc.)

chingor13 (changes requested):

packages/owl-bot/src/owl-bot.ts
bcoe: that's a good call, didn't think of that.

packages/owl-bot/src/owl-bot.ts
```diff
@@ -76,15 +78,29 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
...
       // added by a maintainer to trigger the post processor.
-      // TODO(bcoe): add support for "owlbot:run" label.
-      if (head.repo.full_name !== base.repo.full_name) {
+      if (head.repo.full_name !== base.repo.full_name && !hasRunLabel) {
         logger.info(
           `head ${head.repo.full_name} does not match base ${base.repo.full_name} skipping`
         );
         return;
       }
+      // If the pull request is *not* from a fork, ignore label being added
```
SurferJeffAtGoogle, line 93: Users may want to re-run Owl Bot if:
1.  There was a transient network or cloud build error.
2.  They don't believe the results of the first run.

packages/owl-bot/src/owl-bot.ts
```diff
@@ -76,15 +78,29 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
       if (!installation) {
         throw Error(`no installation token found for ${head.repo.full_name}`);
       }
+      const hasRunLabel = !!context.payload.pull_request.labels.filter(
+        l => l.name === OWLBOT_RUN_LABEL
+      ).length;
+
       // If the pull request is from a fork, the label "owlbot:run" must be
       // added by a maintainer to trigger the post processor.
```
SurferJeffAtGoogle, line 86: Which line of code confirms that the label was added by a maintainer?

SurferJeffAtGoogle (changes requested):

packages/owl-bot/src/owl-bot.ts
chingor13: adding a label requires >= "Triage" permissions

packages/owl-bot/src/owl-bot.ts
bcoe: I've split into two `on`s, I think it simplifies logic.

packages/owl-bot/src/owl-bot.ts
bcoe: Labeling now works when *not* from fork.

packages/owl-bot/src/owl-bot.ts
```diff
@@ -63,105 +63,167 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
...
+
+    if (!installation) {
+      throw Error(`no installation token found for ${head}`);
+    }
+
+    const hasRunLabel = !!context.payload.pull_request.labels.filter(
+      l => l.name === OWLBOT_RUN_LABEL
+    ).length;
+
+    // If the pull request is *not* from a fork, ignore label being added
```
SurferJeffAtGoogle, line 82: Is this comment still accurate?

chingor13 (approved):

packages/owl-bot/src/owl-bot.ts
bcoe: Removed stale comment.

SurferJeffAtGoogle (approved):

Issue #1338, opened 2021-02-02
bazel query does not find dotnet build targets


SurferJeffAtGoogle:
Bazel bot uses [this bazel query and regular expression](https://github.com/googleapis/repo-automation-bots/blob/a4ae53d45996ded87315e40ee496a0c7adcee9d8/packages/bazel-bot/docker-image/generate-googleapis-gen.sh#L72) to find the build targets:

```
bazel query $BAZEL_FLAGS \
'filter("-(go|csharp|java|php|ruby|nodejs|py)\.tar\.gz$", kind("generated file", //...:*))' \
| grep -v -E ":(proto|grpc|gapic)-.*-java\.tar\.gz$")
```

I see two issues:

1.  It fails to find csharp generated files like `google-cloud-domains-v1beta1-csharp.tar.gz`.
2. It smells bad.  

I seek a better bazel query.

vam-google:
Try the following query:
```sh
bazel query $BAZEL_FLAGS \
  'filter("-(go|csharp|java|php|ruby|nodejs|py)$", kind("rule", //...:*))' \
  | grep -v -E ":(proto|grpc|gapic)-.*-java$"
```

It is very similar to the original one, but filters by target names instead of the generated files. The C# rules use different style for declaring generated output, which still results into generating the `.tar.gz` files but those do not match the `generated files` bazel query langauge filter per se.

PR #1642, opened 2021-04-15
feat: update bazel query to include rules instead of tar.gz output files

viacheslav-rostovtsev:
- updates bazel query that creates the list of the build targets to include rules instead of tar.gz output files
- updates script docs
- updates test with new targets

see https://github.com/googleapis/repo-automation-bots/issues/1338

### DIFF
--- a/packages/bazel-bot/test-generate-googleapis-gen.sh
+++ b/packages/bazel-bot/test-generate-googleapis-gen.sh
@@ -66,8 +66,8 @@ git -C googleapis-gen checkout -b other
 # Test!
 export GOOGLEAPIS_GEN=`realpath googleapis-gen-clone`
 export INSTALL_CREDENTIALS="echo 'Pretending to install credentials...''"
-export BUILD_TARGETS="//google/cloud/vision/v1:google-cloud-vision-v1-csharp //google/bogus:api"
-export FETCH_TARGETS="//google/cloud/vision/v1:google-cloud-vision-v1-csharp"
+export BUILD_TARGETS="//google/cloud/vision/v1:vision-v1-nodejs //google/bogus:api"
+export FETCH_TARGETS="//google/cloud/vision/v1:vision-v1-nodejs"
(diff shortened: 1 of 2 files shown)


packages/bazel-bot/test-generate-googleapis-gen.sh
```diff
@@ -66,8 +66,8 @@ git -C googleapis-gen checkout -b other
 # Test!
 export GOOGLEAPIS_GEN=`realpath googleapis-gen-clone`
 export INSTALL_CREDENTIALS="echo 'Pretending to install credentials...''"
-export BUILD_TARGETS="//google/cloud/vision/v1:vision-v1-nodejs.tar.gz //google/bogus:api.tar.gz"
-export FETCH_TARGETS="//google/cloud/vision/v1:vision-v1-nodejs.tar.gz"
+export BUILD_TARGETS="//google/cloud/vision/v1:google-cloud-vision-v1-csharp //google/bogus:api"
```
vam-google, line 69: Looks like it replaces node with C#, which proves that C# works, but stops checking node, so there is a potential for regression. Can you please instead of replaceing node with C#, just add C#?

viacheslav-rostovtsev: adding a second working target does nothing to the test since it only tests that `a commit` is produced

viacheslav-rostovtsev: also the tests does not check for all languages, so I would argue that the potential of regression is unchanged by replacing one with another

bcoe (changes requested):
@viacheslav-rostovtsev would you be comfortable leaving the Node.js, but switching to the new build target format which doesn't include `.tar.gz`?

That way we can be confident in the new approach, but aren't concerned there was a regression for Node.

viacheslav-rostovtsev:
sure

bcoe (approved):

PR #1645, opened 2021-04-17
feat(generated-files-bot): support glob matching

### LINKED ISSUE #1624: Allow generated files bot to accept glob or regex
For Node.js, many of our generated files could be identified by blanket flagging `.kokoro/**`.

It would be convenient to be able to provide a glob (`/**`, or regex `/.*`) in configuration.

danielbankhead:
BREAKING CHANGE: changes function signatures for `parseManifest`, `readExternalManifest`, and `getFileList`

Fixes #1624

### DIFF
--- a/packages/generated-files-bot/test/fixtures/config/valid-config.yml
+++ b/packages/generated-files-bot/test/fixtures/config/valid-config.yml
@@ -1,5 +1,8 @@
 generatedFiles:
 - 'file1.txt'
+- path: '**.html'
+- path: 'docs/**.md'
+- path: 'src/.data/**.json'
 externalManifests:
 - type: json
   file: 'manifest.json'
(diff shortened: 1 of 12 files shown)


chingor13 (commented):
Would it hurt to just treat the `generatedFiles` list entries as globs rather than introducing an extra nesting of the configuration?

chingor13:
> Would it hurt to just treat the `generatedFiles` list entries as globs rather than introducing an extra nesting of the configuration?

Oh, I see the followup in #1646

packages/generated-files-bot/test/fixtures/config/valid-config.yml
```diff
@@ -1,5 +1,8 @@
 generatedFiles:
 - 'file1.txt'
+- path: '**.html'
+- path: 'docs/**.md'
+- path: 'src/.data/**.json'
```
chingor13, line 5: nit: I might separate this config/test scenario into an explicit test case for globbing rather than overloading the basic commenting test.

chingor13 (approved):

PR #1646, opened 2021-04-17
feat(generated-files-bot): support custom messages

### LINKED ISSUE #1625: Allow message created by generated-files-bot to be customized
OwlBot is going to be a major offender in triggering generated-files-bot, I would like to be able to provide a customized message for Node.js repositories specific to our workflow, something like:

1. It looks like you have edited a file that is automatically generated, this will lead to the file being automatically reverted.
2. If you would not like this file to be automatically generated, you can exclude the file by following the instructions here.
3. If you would like to make this change across all of the Node.js repositories, you can edit the templates here.

danielbankhead:
Fixes #1625


bcoe (approved):
This is great 👍

PR #1651, opened 2021-04-20
feat(auto-label): support dashes and more prefixes

### LINKED ISSUE #1649: auto label does not work well with nested paths
#### Steps to reproduce

  1. File an issue with a test failure in nested path (like [this one](https://github.com/GoogleCloudPlatform/python-docs-samples/issues/5657) for example) or simply file an issue with the nested path name
  2. auto-label bot cannot properly extract the correct product label from the issue's tile

This happens as the title for the example above prefixes with the directory (`python-docs-samples/data-science-onramp/ai-platform`) rather than with the product. Would it be possible to have auto-label bot try and extract possible APIs from the entire title with unique product keys rather than just the prefix?

tbpg:
This removes another known prefix from titles (see https://github.com/GoogleCloudPlatform/python-docs-samples/issues/5657). Alternatively, we could build fancy auto-detect logic, but that would be much more complex. So, I kept it simple for now.

Fixes #1649.

### DIFF
--- a/packages/auto-label/test/auto-label.ts
+++ b/packages/auto-label/test/auto-label.ts
@@ -841,6 +841,7 @@ describe('auto-label', () => {
         {title: 'com.google.spanner.helper: ignored', want: 'api: spanner'},
         {title: 'fix(snippets.spanner.helper): ignored', want: 'api: spanner'},
         {title: 'snippets.video: ignored', want: 'api: videointelligence'},
+        {title: 'data-science-onramp.video-intelligence.ignored: ignored', want: 'api: videointelligence'},
         {title: 'unknown: ignored', want: undefined},
         {title: 'spanner with no separator', want: undefined},
         {title: 'fix(unknown): ignored', want: undefined},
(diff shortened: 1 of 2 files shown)


packages/auto-label/test/auto-label.ts
```diff
@@ -841,6 +841,10 @@ describe('auto-label', () => {
         {title: 'com.google.spanner.helper: ignored', want: 'api: spanner'},
         {title: 'fix(snippets.spanner.helper): ignored', want: 'api: spanner'},
         {title: 'snippets.video: ignored', want: 'api: videointelligence'},
+        {
+          title: 'data-science-onramp.video-intelligence.ignored: ignored',
+          want: 'api: videointelligence',
```
dandhlee, line 846: I think this should be `ai-platform`?

dandhlee (commented):
Thanks for putting this in! I've only seen this happen once so I'm happy moving forward with this solution at the moment. Just one comment below.

packages/auto-label/test/auto-label.ts
tbpg: I used Video Intelligence (see `title`) because AI Platform isn't included in the test resources.

See https://github.com/googleapis/repo-automation-bots/blob/master/packages/auto-label/test/fixtures/events/downloadedfile.json.

tbpg (commented):
Thanks!

packages/auto-label/test/auto-label.ts
dandhlee: Ah it's for a test, didn't see that. My bad 😅

dandhlee (approved):

sofisl (approved):

PR #1655, opened 2021-04-20
feat(gcf-utils): extend logger to populate count/event/type

bcoe:
Introduces additional fields to `logger.metric()` for the benefit of a BigQuery log sink:

* `count`: how many times did the event occur?
* `type`: always set to `metric` (this allows us to configure a filter when writing logs into BigQuery).
* `event`: what type of event are we emitting a structured log for?


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -274,8 +274,6 @@ export class GCFBootstrapper {
        */
       logger.resetBindings();
       logger.addBindings(buildTriggerInfo(triggerType, id, name, request.body));
-      logger.metric(`Execution started by ${triggerType}`);
```
bcoe, line 277: Looking through our applications, this was currently the only place we explicitly call `logger.metric`, I wasn't convinced we need to keep this log line.

packages/gcf-utils/README.md
```diff
@@ -146,12 +146,13 @@ The following properties are logged:
 
 > Note: To have your Bot's custom metrics collected from Cloud Logging and processed with other metrics, please refer to the [data-processor documentation](../monitoring-system/data-processor#add-support-for-a-new-metric-from-an-existing-data-source)
 
-To log a metric with field `foo` and value `bar`:
+To log a metric for the event `foo`, with additional structured information
```
JustinBeckwith, line 149: I try to avoid the `foo` and `bar` terminology, because they make it wildly unclear what type of information you'd actually use.  Would `your-event-name` and `example-value`, or something similar work here?

JustinBeckwith (approved):

packages/gcf-utils/README.md
bcoe: Stopped using `foo`.

chingor13 (approved):

packages/gcf-utils/src/logging/gcf-logger.ts
```diff
@@ -104,14 +105,29 @@ export class GCFLogger {
...
+    if (typeof objOrMsg === 'string') {
+      payload.event = objOrMsg;
+    } else {
+      payload = {...payload, ...objOrMsg};
+    }
+    if (typeof addMsg === 'object') {
+      payload = {...payload, ...addMsg};
+      addMsg = undefined;
+    }
+    this.log('metric', payload, addMsg, ...args);
```
danielbankhead, line 130: Is it ok if `objOrMsg` or `addMsg` overwrites `{type: metric}`? If not, perhaps `{...payload, type: 'metric'}` could be used here

danielbankhead (approved):

packages/gcf-utils/src/logging/gcf-logger.ts
bcoe: I was on the fence about whether we should allow an upstream application to override `type`.

I think you're correct that we probably shouldn't allow this, since someone might accidentally set this field, and then not have the metrics data in our BigQuery sink. I've updated the approach and added a test.

PR #1660, opened 2021-04-21
fix(gcf-utils): address octokit type regressions

bcoe:
~@octokit/rest` methods are now name-spaced under `github.rest`.~, I'm not sure what was forcing me to use namespaces, but this seems to have been corrected once I got all libraries updated appropriately.


JustinBeckwith:
was the intent to delete the lock file?

bcoe:
> was the intent to delete the lock file?

@JustinBeckwith no, I think it was removed because the postinstall script fails.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -387,16 +387,23 @@ export class GCFBootstrapper {
...
-        installationId,
-      },
-    });
+    const opts = {
+      appId: cfg.appId,
+      privateKey: cfg.privateKey,
+      installationId,
+    };
+    if (wrapOptions?.logging) {
+      const LoggingOctokit = Octokit.plugin(LoggingOctokitPlugin)
```
bcoe, line 397: The Probot Options types seemed to conflict with our approach of instantiating Octokit ourselves.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -109,7 +109,7 @@ export const addOrUpdateIssueComment = async (
   for (const comment of listCommentsResponse.data) {
     if (comment.body?.includes(commentMark)) {
       // We found the existing comment, so updating it
-      await github.issues.updateComment({
+      await github.rest.issues.updateComment({
```
bcoe, line 112: In the  newest version of `probot`, `github` is moved under the `rest` namespace.

packages/gcf-utils/src/gcf-utils.ts
bcoe: we will want to test our cron based applications.

chingor13 (approved):

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: I tested it and failed🤣🤣

PR #1677, opened 2021-04-24
fix: update title and author for auto-approve option 1

parthea:
Update auto-approve option 1 for approving PRs like [this one](https://github.com/googleapis/google-api-python-client/pull/1308).


sofisl:
@parthea, I think we want to add this use case instead of deleting the current one, since that one is in use by another repository.

PR #1681, opened 2021-04-26
refactor(sync-repo-settings): split config out from the bot

JustinBeckwith:
This is part of a broader effort to enable organization level configuration for sync-repo-settings.  It starts by:
- Fixing a gaffe I tripped on in the README
- Splitting config validation and fetching out into it's own file
- Introducing config specific tests instead of relying on the bot tests
- Moving away from probot's built in config fetcher, and instead just directly fetching the files.

On the last bit - I did this because eventually I would like to be able to _merge_ the configurations between a local repo, and the upstream `.github` repository where a centralized config will be stored.   The current probot implementation doesn't allow for that, so this should be behaviorally compatible.

### DIFF
--- a/packages/sync-repo-settings/README.md
+++ b/packages/sync-repo-settings/README.md
@@ -19,6 +19,9 @@ squashMergeAllowed: true
 # Defaults to `false`
 mergeCommitAllowed: false
 
+# Automatically delete head branches after merging PRs.  Defaults to `true`.
+deleteBranchOnMerge: true
+
 # Rules for branch protection (add multiple entries to configure multiple branches)
 branchProtectionRules:
 # Identifies the protection rule pattern. Name of the branch to be protected.
@@ -35,8 +38,6 @@ branchProtectionRules:
   requiresCodeOwnerReviews: true
   # Require up to date branches
   requiresStrictStatusChecks: true
-  # Automatically delete head branches after merging PRs.  Defaults to `true`.
-  deleteBranchOnMerge: true
   # List of required status check contexts that must pass for commits to be accepted to matching branches.
   requiredStatusCheckContexts:
     - check1
--- a/packages/sync-repo-settings/src/config.ts
+++ b/packages/sync-repo-settings/src/config.ts
@@ -1,0 +1,122 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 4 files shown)


packages/sync-repo-settings/README.md
```diff
@@ -19,6 +19,9 @@ squashMergeAllowed: true
 # Defaults to `false`
 mergeCommitAllowed: false
 
+# Automatically delete head branches after merging PRs.  Defaults to `true`.
+deleteBranchOnMerge: true
```
bcoe, line 23: I'm glad we're turning this own.

packages/sync-repo-settings/src/config.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+  return null;
+}
+
+/**
+ * For a given repository, fetch the config yaml.
+ * @param options GetConfigOptions
+ * @returns an individual Repo config
+ */
+async function getConfigFile(
```
bcoe, line 108: I wonder if it would be worth pulling config handling logic up to `gcf-utils`, so that we can use it for other repos?

bcoe: Not blocker for this PR.

bcoe (approved):
I like how you've extracted the config loading logic, I wonder if we could pull this logic somewhere such that we could use it for other bots.

packages/sync-repo-settings/src/config.ts
```diff
@@ -0,0 +1,122 @@
...
+  options: GetConfigOptions
+): Promise<RepoConfig | null> {
+  const orgYamlRes = await options.octokit.repos.getContent({
+    owner: options.owner,
+    repo: options.repo,
+    path: `.github/${configFileName}`,
```
bcoe, line 114: I think the `octokit` config helper looks for both `.yml` and `.yaml`?

packages/sync-repo-settings/src/config.ts
JustinBeckwith: For other bots, for now, I'm content with the default behavior where the `.github` repository level config flat overrides the per-repo config.  This bot is "special" in the way I am trying to treat it for sure.  If the pattern plays out and we want to use it elsewhere, I'd happily up-level it to the util library 👍

packages/sync-repo-settings/src/config.ts
JustinBeckwith: The `context.config` method requires a file path, which includes the extension.  So this is consistent with the previous behavior:  https://probot.github.io/api/latest/classes/context.html#config

PR #1697, opened 2021-04-27
fix(policy): do not use search API due to rate limiting

JustinBeckwith:
This required more refactoring to be efficient than expected (sorry)


packages/policy/src/changer.ts
```diff
@@ -55,108 +51,95 @@ async function cachedGet(url: string) {
...
+  private async hasMatchingPR(title: string) {
+    if (!this.prsCache) {
+      this.prsCache = await this.octokit.paginate(this.octokit.pulls.list, {
+        owner: this.owner,
+        repo: this.name,
+        state: 'open',
+        per_page: 100,
+      });
+    }
+    const matches = this.prsCache.filter(x => x.title === title);
```
danielbankhead, line 83: Would `prsCache` have the potential to cause the process to run out of memory for repos that have a lot of open PRs? This may not be a problem atm - I'll share this snippet in case it becomes one:

```ts
const params = {
  owner: this.owner,
  repo: this.name,
  state: 'open',
  per_page: 100,
};

for await (const response of octokit.paginate.iterator(this.octokit.pulls.list, params)) {
  if (response.title === title) {
    return true;
  }
}

return false;
```

danielbankhead (approved):

packages/policy/src/changer.ts
JustinBeckwith: This is handy, and usually I'd totally agree with you.  In this particular case though, I'm caching the total set of open PRs on the repository because I want to go back through them multiple times to check for other open PRs.  The caching happens since we're already submitting 2 different types of PR in here, each where I have to check for the title.  If a repository has more than a few dozen open pull requests, they're probably having a bad day anyhow :)

packages/policy/src/changer.ts
```diff
@@ -55,108 +51,95 @@ async function cachedGet(url: string) {
...
-    [
-      'SECURITY.md',
-      {
-        content,
-        mode: '100644',
-      },
-    ],
-  ]);
+  /**
+   * Submit a pull request to the target repository to add a SECURITY.md.
```
bcoe, line 88: This could be a SECURITY.md or a CODE_OF_CONDUCT.md? perhaps just rephrase:

`Submit a pull request to the target repository to add files required to meet policy, e.g., SECURITY.md`

packages/policy/src/changer.ts
```diff
@@ -55,108 +51,95 @@ async function cachedGet(url: string) {
...
- */
-export async function addCodeOfConduct(repo: GitHubRepo, octokit: Octokit) {
-  // first, make sure there's no open PR for this
-  const title = 'chore: add a Code of Conduct';
-  const prs = await octokit.search.issuesAndPullRequests({
-    q: `repo:${repo.full_name} "${title}" in:title is:open`,
-  });
-  if (prs.data.total_count > 0) {
-    return;
+export class Changer {
```
bcoe, line 54: nit: I feel like `Changer` is a bit generic, perhaps `MissingFilesSubmitter`, or something along those lines (_a a great name didn't immediately jump out at me_).

bcoe (commented):
Just added a couple nits for future work.

PR #1704, opened 2021-04-29
feat(snippet-bot): new status checks

tmatsuo:
started to create status checks for 2 types of violations.
- Region tag product prefix
- Disruptive region tag removal

also merged REMOVE_CONFLICTING_TAG into REMOVE_USED_TAG violations.

The bot will start submitting new status checks, so I'll announce this
change before merging.


packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -263,6 +264,11 @@ async function scanPullRequest(
   let tagsFound = false;
   const failureMessages: string[] = [];
 
+  // Whether to ignore prefix requirement.
+  const noPrefixReq = pull_request.labels.some((label: Label) => {
+    return label.name === NO_PREFIX_REQ_LABEL;
```
bcoe, line 269: will it be the case that we ignore prefixes for an entire repository, or will it tend to be on a single PR?

in what cases do we ignore prefixes?

packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -535,6 +577,31 @@ ${REFRESH_UI}
...
 
+  // Status checks for missing region tag prefix
+  if (
+    configuration.alwaysCreateStatusCheck() ||
+    productPrefixViolations.length > 0
+  ) {
+    try {
+      await context.octokit.checks.create(prefixCheckParams);
+    } catch (e) {
+      e.message = `Error creating prefix status check: ${e.message}`;
```
bcoe, line 588: If we were to allow this to throw, and the job was retried, would it result in a comment being placed on a PR twice?

Since we don't expect exceptions to happen, my temptation is usually to allow the process to crash, so that we catch bugs faster.

packages/snippet-bot/test/snippet-bot.ts
```diff
@@ -580,6 +580,61 @@ describe('snippet-bot', () => {
...
+        name: 'pull_request',
+        payload,
+        id: 'abc123',
+      });
+
+      sinon.assert.calledOnce(getSnippetsStub);
+      requests.done();
+      diffRequests.done();
+    });
+
```
bcoe, line 637: It would be good to have a test that demonstrates a status check being set to `failure`, and one setting it to success -- I'm a bit surprised we didn't need to update existing tests?

bcoe (changes requested):
A couple questions, and a request for two tests.

Looking good to me otherwise.

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: For some languages, snippet-gen (snippet auto generator) is using a different prefix (apishortname instead of registered product prefix). In this case, you can add this label for ignoring this check.

This feature is requested by @amanda-tarafa

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: > If we were to allow this to throw, and the job was retried, would it result in a comment being placed on a PR twice?
No, we always re-use the same comment.

> Since we don't expect exceptions to happen, my temptation is usually to allow the process to crash, so that we catch bugs faster.

I can remote `try`.

packages/snippet-bot/test/snippet-bot.ts
tmatsuo: Ok, I can add some tests with `alwaysCreateStatusCheck` config.

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: It turned out the `try` swallowed the nock errors.

tmatsuo:
@bcoe Thanks! PTAL

packages/snippet-bot/test/snippet-bot.ts
tmatsuo: Because we used `try` - `catch`, it swallowed the nock errors.

Now I added a new test case for 4 success checks.

tmatsuo (commented):
Added a new test case

bcoe (approved):
Thanks for the tests.

PR #1707, opened 2021-04-30
feat: additional validation for release PRs

sofisl:
This PR adds additional validation checks if the incoming PR is a release PR for Node. It checks that (i) it is a release PR, (ii) the major version is not upgraded, and (iii), the minor version is upgraded, not downgraded, etc.


sofisl:
@bcoe, I wrote out the initial logic for checking whether Node release PRs followed these checks, but I attempted to write it in a way where we could maybe add more rules/cases for other types of PRs and languages.

packages/auto-approve/src/utils-for-PR-checking.ts
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * @param versions an object containing the previous and newer versions of the package being updated
+ * @returns whether the major version changed.
+ */
+export function isMajorVersionChanging(versions: Versions): boolean {
+  return Number(versions.newMajorVersion) - Number(versions.oldMajorVersion) ===
+    0
+    ? false
+    : true;
+}
```
danielbankhead, line 136: Minor, but this may be simplified to:

```suggestion
export function isMajorVersionChanging(versions: Versions): boolean {
  return versions.newMajorVersion === versions.oldMajorVersion;
}
```

packages/auto-approve/src/utils-for-PR-checking.ts
```diff
@@ -0,0 +1,188 @@
...
+ * This function determines whether the minor version of a package was updated.
+ *
+ * @param versions an object containing the previous and newer versions of the package being updated
+ * @returns whether the minor version was upgraded.
+ */
+export function isMinorVersionUpgraded(versions: Versions): boolean {
+  return Number(versions.newMinorVersion) - Number(versions.oldMinorVersion) > 0
+    ? true
+    : false;
+}
```
danielbankhead, line 148: Optional: this can also be:
```suggestion
export function isMinorVersionUpgraded(versions: Versions): boolean {
  return Number(versions.newMinorVersion) > Number(versions.oldMinorVersion);
}
```
or

```suggestion
export function isMinorVersionUpgraded(versions: Versions): boolean {
  return BigInt(versions.newMinorVersion) > BigInt(versions.oldMinorVersion);
}
```

danielbankhead (approved):
Left some optional notes/suggestions

sofisl:
@danielbankhead, thanks so much for the thoughtful and in-depth review :) Super helpful comments!

packages/auto-approve/src/check-pr.ts
```diff
@@ -86,38 +110,51 @@ export async function checkPRAgainstConfig(
         pr.pull_request.changed_files <= rulesToValidateAgainst.maxFiles;
     }
     logger.info(
-      `Info for ${repoOwner}/${repo}/${prNumber}\nAuthor: ${rulesToValidateAgainst.author}\nTitles Match? ${titlesMatch}\nFile Paths Match? ${filePathsMatch}\nFile Count Matches? ${fileCountMatch}`
+      `Info for ${repoOwner}/${repo}/${prNumber}\nAuthor: ${rulesToValidateAgainst.author}\nFile Paths Match? ${filePathsMatch}\nFile Count Matches? ${fileCountMatch}\nVersions are correct? ${versionChecks}`
```
chingor13, line 113: Future: consider emitting each of these separately - it'll be easier to parse (human and metrics).

packages/auto-approve/src/check-pr.ts
```diff
@@ -86,38 +110,51 @@ export async function checkPRAgainstConfig(
...
+ * @param file The incoming target file that has a matching ruleset in language-versioning-rules
+ * @param pr The matching ruleset of the file above from language-versioning-rules
+ * @returns true if the package was upgraded appropriately, and had only one thing changed
  */
-export function checkFilePathsMatch(
-  prFiles: string[],
-  validTypeOfPR: ValidPr
-): boolean {
-  if (!validTypeOfPR.changedFiles) {
+function runVersioningValidation(addtlValidationFile?: {
```
chingor13, line 132: nit: no need to abbreviate words

chingor13 (commented):
In the future, consider splitting a refactor (moving logic around) from the feature that adds new logic. It's hard to review both a refactor and new feature because it's hard to see what's moved and what's newly added.

packages/auto-approve/src/auto-approve.ts
```diff
@@ -223,6 +228,10 @@ export function handler(app: Probot) {
               issue_number: prNumber,
               labels: ['automerge: exact'],
             });
+            logger.metric('auto_approve.approved_tagged', {
```
bcoe, line 231: thanks for metrics \o/

packages/auto-approve/src/check-pr.ts
```diff
@@ -15,8 +15,18 @@
 // eslint-disable-next-line node/no-extraneous-import
 import {ProbotOctokit} from 'probot';
 import {PullRequestEvent} from '@octokit/webhooks-definitions/schema';
-import {getChangedFiles} from './get-PR-info';
+import {getChangedFiles, File} from './get-PR-info';
```
bcoe, line 18: nit: I would use all lower case for file names:

* convention for JavaScript.
* also works better for case-insensitive file systems.

packages/auto-approve/src/check-pr.ts
```diff
@@ -15,8 +15,18 @@
...
 import {logger} from 'gcf-utils';
+import {
+  FileSpecificRule,
+  getTargetFile,
+  getVersions,
+  isMajorVersionChanging,
+  isMinorVersionUpgraded,
```
bcoe, line 25: I like the functional naming of helpers 😄 it's clear what each of these does.

packages/auto-approve/src/check-pr.ts
```diff
@@ -55,25 +64,40 @@ export async function checkPRAgainstConfig(
...
+      changedFiles,
+      rulesToValidateAgainst.author,
+      languageVersioningRules
+    );
+
+    // If we've found the file in this ruleset, let's run the additional checks
+    if (fileAndFileRule) {
+      // TODO: make the checks conditional based on the different kinds of conditions
+      // to be checked, i.e., runDependencyUpgradeValidation, etc.
+      versionChecks = runVersioningValidation(fileAndFileRule);
```
bcoe, line 96: structural nit, I would probably just return `false` early on failure, to avoid needing to reassign `versionChecks`:

```js
 if (fileAndFileRule) {
      // TODO: make the checks conditional based on the different kinds of conditions
      // to be checked, i.e., runDependencyUpgradeValidation, etc.
     if (!runVersioningValidation(fileAndFileRule)) return false;
}
```

This is a personal preference, and non blocking.

packages/auto-approve/test/check-pr.test.ts
```diff
@@ -181,16 +199,73 @@ describe('check pr against config', async () => {
...
+        validPR,
+        pr,
+        octokit
+      );
+      scopes.done();
+      assert.ok(prMatchesConfig);
+    });
+  });
+
+  describe('additional validation checks', async () => {
```
bcoe, line 225: rather than using pros, I would just indicate that this block of tests is testing the `changedFiles` configuration:

```
describe("changedFiles", () => {});
```

Since we'll be adding _additional_ additional validation checks over time, then I think the assertions within the describe block can describe the algorithm in detail:

```
it("should indicate PR is mergeable if it only modifies the files indicated in changedFiles list");
it('should indicate PR is not-mergeable if it modifies files other than those in the changedFiles list");
```

packages/auto-approve/test/check-pr.test.ts
```diff
@@ -181,16 +199,73 @@ describe('check pr against config', async () => {
...
+    it('should correctly identify a PR that needs special attention', async () => {
```
bcoe, line 226: See my above comment, I think we could delve deeper into the algorithm in this set of assertions 👍

bcoe (changes requested):
Left a few nits, one thing that jumped out at me which I think would be good to describe in tests, is whether it's necessary that *only* the files in `changedFiles` are modified.

Thanks for doing this work, can't wait to see it in action.

packages/auto-approve/src/check-pr.ts
sofisl: File here is an interface I'm exporting (perhaps the name is too generic). I've noticed the convention for interfaces are uppercase; should I just change the name to something more specific?

packages/auto-approve/test/check-pr.test.ts
sofisl: > rather than using pros, I would just indicate that this block of tests is testing the `changedFiles` configuration:
> 
> ```
> describe("changedFiles", () => {});
> ```
> 

I'm a bit confused, because this block tests whether a PR requires additional checks _besides_ the changedFiles and titles checks. (i.e., whether or not we need to check versions, or something else). 

Would it be alright to still call this block "additional validation checks", and further detail each tests' algorithm in the description of the test?

packages/auto-approve/src/check-pr.ts
danielbankhead: To assist, I think Ben meant the `'./get-PR-info'` part

bcoe (approved):

branbennett2 (approved):

PR #1715, opened 2021-05-03
fix: a test in database.ts

SurferJeffAtGoogle:
Test was failing because left-overs from earlier tests were left in the database.  Fixed so left-overs can't affect new tests.


danielbankhead (approved):
lgtm

chingor13 (approved):

packages/owl-bot/test/database.ts
```diff
@@ -44,7 +44,7 @@ describe('database', () => {
 
   it('stores and retrieves configs', async () => {
     const db = admin.firestore();
-    const store = new FirestoreConfigsStore(db, 'test-');
+    const store = new FirestoreConfigsStore(db, 'test-' + uuidv4() + '-');
```
bcoe, line 47: @SurferJeffAtGoogle is there a risk that we gradually collect more stale data.

An approach we've taken in a few places, is to have a step that deletes any entries with a common prefix, e.g., `test-`, that are older than a certain age -- a delta longer than we expect any tests to take.

Issue #1614, opened 2021-04-06
owlbot should not open lock PRs if there are no other changes


chingor13:
If the image rebuilds, it currently always sends a PR regardless of whether or not the code will be changed. We should run the post-processor on the codebase to see if there will be changes before opening the PR.

SurferJeffAtGoogle:
Let's discuss in the meeting.

PR #1724, opened 2021-05-04
fix: only create pull requests when new docker image yields changes

SurferJeffAtGoogle:
fixes https://github.com/googleapis/repo-automation-bots/issues/1614


SurferJeffAtGoogle:
PR generated with this code:
https://github.com/googleapis/nodejs-dlp/pull/605

Although from my GCP project surferjeff-firestore instead of the prod project.

packages/owl-bot/src/update-lock.ts
```diff
@@ -0,0 +1,62 @@
...
+export async function maybeCreatePullRequestForLockUpdate(
+  octokitFactory: OctokitFactory,
+  githubRepo?: GithubRepo,
+  localRepoDir?: string,
+  createPullRequestFromLastCommit = createPr.createPullRequestFromLastCommit,
+  logger = console
+): Promise<void> {
+  const cmd = newCmd(logger);
+  const cwd = localRepoDir ?? '.';
+  const status = cmd('git status --porcelain', {cwd}).toString('utf8').trim();
```
bcoe, line 29: does this rely on the fact that an empty string will be returned if no changes occur, I think it might be worth adding a comment that points this out 👍

packages/owl-bot/test/github-repo.ts
```diff
@@ -0,0 +1,45 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {githubRepoFromUri} from '../src/github-repo';
+import * as assert from 'assert';
+
+describe('githubRepoFromUri()', () => {
```
bcoe, line 18: I like this helper 👍

bcoe (changes requested):
A couple nits.

One minor concern, we'll now end up running the OwlBot post-processor twice for every lock update correct? Could we skip running for the first commit of a lock update, knowing that we will have already run?

packages/owl-bot/src/update-lock.ts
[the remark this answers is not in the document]
SurferJeffAtGoogle: The [docs](https://git-scm.com/docs/git-status) say it will default to v1, which is fine.

packages/owl-bot/cloud-build/update-lock-branch.yaml
```diff
@@ -0,0 +1,94 @@
...
+  # TODO: Dump the change history from the docker image so we can include it
+  #       in the pull request description.
+  - name: 'gcr.io/cloud-builders/git'
+    args:
+      - commit
+      - -m
+      - "chore: new owl bot post processor docker image\n\n${_CONTAINER}"
+    dir: '${_REPOSITORY}'
+
+  # Run container listed in OwlBot.yaml.
```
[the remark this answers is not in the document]
SurferJeffAtGoogle: Yes.

packages/owl-bot/src/update-lock.ts
SurferJeffAtGoogle: Added a comment.

SurferJeffAtGoogle:
> A couple nits.
> 
> One minor concern, we'll now end up running the OwlBot post-processor twice for every lock update correct? Could we skip running for the first commit of a lock update, knowing that we will have already run?

Yes, that's a potential optimization for the future after we observe the cost of running it twice.

SurferJeffAtGoogle:
Ready for another look.

bcoe (approved):

Issue #1709, opened 2021-05-02
Owl Bot : Only close PRs that were opened by Owl Bot


parthea:
Owl bot has a feature where it closes PRs that have no file changed. The feature should be adjusted to support `release-please`. @SurferJeffAtGoogle suggested the following to solve the issue:

- Owl bot should only close PRs that it opens
- A new label could be added to signal to owl bot that the PR should be ignored.

**More details**
Unrelated to owl bot, I opened a PR for `release-please` to change the version number using these [steps](https://github.com/googleapis/release-please#how-do-i-change-the-version-number).  As part of the process, a PR is opened with an empty commit. Unexpectedly, owl bot closed the PR. This issue repeated several times. See https://github.com/googleapis/python-security-private-ca/pull/42 and https://github.com/googleapis/python-security-private-ca/pull/43.

bcoe:
This feels like a `p2` to me?

SurferJeffAtGoogle:
@JustinBeckwith is suffering from this too:
https://github.com/googleapis/nodejs-service-control/pull/29

PR #1725, opened 2021-05-05
fix: issue 1709

SurferJeffAtGoogle:
Stop owl bot from closing pull requests that have nothing to do with owl bot.

Fixes https://github.com/googleapis/repo-automation-bots/issues/1709

### DIFF
--- a/packages/owl-bot/src/core.ts
+++ b/packages/owl-bot/src/core.ts
@@ -72,7 +72,9 @@ interface Token {
   repository_selection: string;
 }
 
-const UPDATE_LOCK_PULL_REQUEST_LABEL = 'owl-bot-update-lock';
+export const OWL_BOT_LOCK_UPDATE = 'owl-bot-update-lock';
+export const OWL_BOT_COPY = 'owl-bot-copy';
+export const OWL_BOT_IGNORE = 'owl-bot-ignore';
 
 export async function triggerPostProcessBuild(
   args: BuildArgs,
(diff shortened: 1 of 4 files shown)


packages/owl-bot/src/core.ts
```diff
@@ -497,14 +519,9 @@ async function updatePullRequestAfterPostProcessor(
     });
   } else {
     // If the pull request is a DRAFT lock file update, close it or promote it.
-    const {data: pull} = await octokit.pulls.get({
-      owner,
-      repo,
-      pull_number: prNumber,
-    });
     if (
       pull.draft &&
```
bcoe, line 523: It seems like we want to revert this check for `draft` since we opted no to take this route when performing lock updates?

bcoe (changes requested):
It seemed like we currently only close a pull request if it's a draft, but I think that we're moving away from using drafts for lock updates?

packages/owl-bot/src/core.ts
SurferJeffAtGoogle: I'd like that to fall out of scope for this pull request.  I'll remove that logic in a separate pull request.

SurferJeffAtGoogle:
> It seemed like we currently only close a pull request if it's a draft, but I think that we're moving away from using drafts for lock updates?

Yes, but that move is not part of this pull request.  This pull request is designed to fix the bug, nothing more.

SurferJeffAtGoogle:
> > It seemed like we currently only close a pull request if it's a draft, but I think that we're moving away from using drafts for lock updates?
> 
> Yes, but that move is not part of this pull request. This pull request is designed to fix the bug, nothing more.

Actually no.  Owl Bot closes pull requests that are not drafts if they change 0 files.  With this pull request, it will only close such pull requests if they have a 'owl-bot-copy' or 'owl-bot-update-lock' label and do not have an 'owl-bot-ignore' label.

bcoe (approved):

PR #1732, opened 2021-05-06
fix(blunderbuss): implement per PR lock backed by datastore

### LINKED ISSUE #1171: blunderbuss can assign multiple times (race condition)
Since blunderbuss listens to a couple of events, it's possible for a single pull request or issue to be assigned multiple times.

A per issue lock would likely solve this.

Example: https://github.com/googleapis/java-vision/pull/332 assigned the pull request twice (likely from the pull_request.opened event and then the pull_request.labeled event).

tmatsuo:
fixes #1171

Currently I'm using https://www.npmjs.com/package/google-datastore-emulator for the test.
It uses `google/cloud-sdk` docker images under the cover. I hope docker is available on github actions environment.

Initially I tried mocking the `DatastoreLock` object, but it failed and it was really a painful experiences.


packages/blunderbuss/src/blunderbuss.ts
```diff
@@ -129,6 +133,13 @@ export function blunderbuss(app: Probot) {
       const byConfig = issue ? config.assign_issues_by : config.assign_prs_by;
       const issuePayload = issue || pullRequest;
 
+      // Acquire the lock.
+      const lock = new DatastoreLock('blunderbuss', lockTarget);
+      const lockResult = await lock.acquire();
+      if (!lockResult) {
+        throw new Error('Failed to acquire a lock');
```
kurtisvg, line 140: Can you include the issue's info in the error, so it's easier to pinpoint in the logs?

packages/blunderbuss/src/blunderbuss.ts
```diff
@@ -129,6 +133,13 @@ export function blunderbuss(app: Probot) {
...
+      }
+
```
kurtisvg, line 142: Once the lock is successfully acquired, we need to refresh the issue's data (to ensure it hasn't changed outside of the critical section). We may also want to introduce a brief sleep (500ms to 1s) to give auto label a chance to do its thing.

packages/blunderbuss/src/datastore-lock.ts
```diff
@@ -0,0 +1,149 @@
...
+      await transaction.run();
+      // First get the entity
+      const [entity] = await transaction.get(this.key);
+      logger.debug(entity);
+
+      if (entity === undefined || Date.now() - entity.t > this.lockExpiry) {
+        const entity = {
+          key: this.key,
+          data: {
+            t: Date.now(),
```
kurtisvg, line 105: Since you are storing date the lock is acquired rather than the date the lock should expire, different versions of the app could consider this app unlocked at different times. 

I think it would be better to store the expiration time, and `release()` can change the expiration to `Date.now()` if the lock is released early.

packages/blunderbuss/src/datastore-lock.ts
```diff
@@ -0,0 +1,149 @@
...
+
+const DEFAULT_LOCK_EXPIRY = 20 * 1000; // 20 seconds
+const LOCK_ACQUIRE_TIMEOUT = 120 * 1000; // 120 seconds
+const BACKOFF_INITIAL_DELAY = 2 * 1000; // 1 seconds
+const BACKOFF_MAX_DELAY = 10 * 1000; // 10 seconds
+
+enum AcquireResult {
+  Success = 1,
+  Failure,
+  LockActive,
```
kurtisvg, line 27: I'm not sure I understand the differentiation between Failure / LockActive. It seems like the possible outcomes are:

1. Acquired the lock successfully (Success)
2. The lock is acquired by someone else (Failure)
3. Something else unexpected happens

It seems like #3 should just throw an error?

packages/blunderbuss/src/datastore-lock.ts
```diff
@@ -0,0 +1,149 @@
...
+   */
+  public async release(): Promise<boolean> {
+    const transaction = this.datastore.transaction();
+    try {
+      await transaction.run();
+      const [entity] = await transaction.get(this.key);
+      if (entity === undefined || entity.uuid !== this.uniqueId) {
+        await transaction.rollback();
+        return true;
+      }
```
kurtisvg, line 137: This section seems wrong to me - if we attempt to release a lock, it means we think we successfully acquired it earlier. If the lock is acquired by another uuid, doesn't that mean something's gone wrong? Should this throw an error instead?

kurtisvg (changes requested):

packages/blunderbuss/src/blunderbuss.ts
tmatsuo: Done (with the next commit)

packages/blunderbuss/src/blunderbuss.ts
tmatsuo: You're right in a long run, but I think it is better to introduce the lock first, then do another round of refactoring.

For giving auto label a chance, we already have 10 seconds sleep, and now it does the trick.

packages/blunderbuss/src/datastore-lock.ts
tmatsuo: Good point, I changed it to `expiry`.

packages/blunderbuss/src/datastore-lock.ts
tmatsuo: Datastore can throw an error when there are too many writes on one single entity. In that case, we want to retry with some delay.

packages/blunderbuss/src/datastore-lock.ts
tmatsuo: If the lock is acquired by another uuid, this means the lock is expired and another process has obtained the lock. In this case, it should not delete the lock.

packages/blunderbuss/src/blunderbuss.ts
tmatsuo: It's done here:
https://github.com/googleapis/repo-automation-bots/blob/819e646055cfd277f673a3caa80cf1c2d49536fb/packages/blunderbuss/src/blunderbuss.ts#L195

tmatsuo (commented):
@kurtisvg 
Thanks for the reivew! Addressed the review comment.

For re-fetching the issue/pr, I still think you're right in a long run, but I prefer doing another refactoring in another PR for that.

PTAL

packages/blunderbuss/src/blunderbuss.ts
kurtisvg: I don't think it will fix #1171, because the lock itself isn't enough. While it prevents blunderbuss from operating on any individual issue concurrently, it's still operating on an outdated copy of that issue's information that was sent in the webhook. Imagine the following scenario (where T(X) is an action taking place at some arbitrary X unit of time):

T(0): Someone opens a PR and a webhook event is triggered (Issue = no assignees)
T(1): Blunderbuss receives the event from T(0) and sleeps for 10
T(2): The issue is applied a label and another webhook event is triggered
T(1): Blunderbuss receives the event from T(2) and sleeps for 10
T(11): Blunderbuss process the event at T(0), acquires the lock, assigns it to someone
T(12): Blunderbuss process the event at T(1), acquires the lock, assigns it to someone else

If the issue status is refreshed after the lock is acquired, Blunderbuss will know it has assignees and will not reassign (unless it's the `blunderbuss: assign` label, which allows for assigning to multiple times).

kurtisvg: (in fact with the introduction of the lock and the updating of issues, we can probably significantly reduce the sleep time without experiencing the race condition - 10s is a long time to wait)

packages/blunderbuss/src/datastore-lock.ts
kurtisvg: Isn't this an error condition? It essentially means that blunderbuss didn't acquire the lock for long enough and had exited the critical section prematurely. I think it would be better to throw an error in those situations, so that the error is logged and we can possible address it by increasing the lock timeout

packages/blunderbuss/src/blunderbuss.ts
kurtisvg: (If you still want to wait to a separate PR to do this, let's just make sure we don't close #1171 by merging this one)

packages/blunderbuss/src/datastore-lock.ts
kurtisvg: The logic for both sections is [essentially the same](https://github.com/googleapis/repo-automation-bots/pull/1732/files#diff-3215fb0ab8076b6916f2a23222a8360fe81a7cafdbc9e2813c73e7e6012792d4R79-R88)

Would it be better to just throw an error when the lock is already taken and then just retry on any error? 

At some point you might want to introduce functionality for providing a timeout for acquiring the lock, in which case returning the error from the most recent attempt that doesn't exceed the user provided timeout will likely be useful.

packages/blunderbuss/src/blunderbuss.ts
tmatsuo: Ya, separate PR. I changed the comment to `part of #1171`

packages/blunderbuss/src/datastore-lock.ts
tmatsuo: Changed to throw an error.

tmatsuo (commented):
@kurtisvg Thanks, PTAL

packages/blunderbuss/src/datastore-lock.ts
tmatsuo: Sorry I don't understand. If the lock is active, wait and retry is the most natural thing to do...

tmatsuo (commented):
@kurtisvg
Sorry the previous commit swallowed the error.

Now it correctly throws an error when the lock is acquired by another process.
I also added a test for it.

PTAL

packages/blunderbuss/src/datastore-lock.ts
kurtisvg: It's more of a nit at this point since they are handled the same, but what I'm trying to say is:

1. I agree with you that retrying on an error makes sense
2. I agree with you that retrying on lock being taken makes sense. However, since we are already retrying on an error because of the above statement, there is no reason to treat already locked locks special. Treat it as an error that is retried with the same logic. 
3.  Finally, I think that the above is more desirable behavior because (probably) eventually, user's of your lock may want to timeout how long they wait for it (something like `await lock.acquire(30s)`). In these situations, it would be better to not return true/false, but instead have no return value and assume that if it completed if it locked, and if it's not it'll throw an error (which will then contain the root cause - it was already locked, database errors, etc).

kurtisvg (approved):
LGTM from the logic side - may also be a good idea to have someone that knows NodeJS better than myself to look too.

packages/blunderbuss/src/datastore-lock.ts
tmatsuo: > 2
I understand. It's more of a sake of development. It is better to emit different debug message during development.

> 3
Currently only failure is by timeout.

packages/blunderbuss/src/blunderbuss.ts
```diff
@@ -143,6 +154,7 @@ export function blunderbuss(app: Probot) {
             issueOrPRNumber,
             context.payload.label?.name
           );
+          lock.release();
```
chingor13, line 157: Might be nicer to wrap the main logic into a helper, the acquire and release the lock around calling the helper. I could this being easy to add a logic branch and forget to release the lock.

tmatsuo: Good point. Will do.

tmatsuo: Done

tmatsuo (commented):
@chingor13 Thanks, I made a helper function and wrap it with the lock. PTAL

chingor13 (approved):

PR #1733, opened 2021-05-06
feat(publish-script): conditionally add timeout when publishing bots

sofisl:

### DIFF
--- a/scripts/publish-function.sh
+++ b/scripts/publish-function.sh
@@ -26,6 +26,7 @@ bucket=$3
 keyLocation=$4
 keyRing=$5
 functionRegion=$6
+timeout=$7
 
 # To use a different runtime for a bot, give 7th parameter in bot's
 # cloudbuild.yaml.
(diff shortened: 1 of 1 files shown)


scripts/publish-function.sh
```diff
@@ -26,6 +26,7 @@ bucket=$3
 keyLocation=$4
 keyRing=$5
 functionRegion=$6
+timeout=$7
 
 # To use a different runtime for a bot, give 7th parameter in bot's
 # cloudbuild.yaml.
```
tmatsuo, line 32: Is this comment still accurate?

sofisl: Good catch! I've updated my code to incorporate this comment.

tmatsuo: I see a small bug in the logic.
If you also specify the 8th parameter, the 7th parameter is ignored.

tmatsuo (changes requested):
I see a small bug in the logic.
If you also specify the 8th parameter, the 7th parameter is ignored.

scripts/publish-function.sh
chingor13: +1  https://github.com/googleapis/repo-automation-bots/pull/1733/files#diff-9f7786d7ebdf02a8d4c090d3ec026cbc30755ea4db5604c7c4010f092ff7b042R35 is checking for exactly 7 args

chingor13 (changes requested):

chingor13 (approved):

PR #1750, opened 2021-05-11
feat(release-please): allow overriding label added when creating GitHub release

chingor13:
Exposes the configuration that allows for customizing the label applied to the release PR when creating the GitHub release. For some projects, tagging the release should mark the release pull request as "published".

### DIFF
--- a/packages/owl-bot/cloud-build/update-pr.yaml
+++ b/packages/owl-bot/cloud-build/update-pr.yaml
@@ -46,6 +46,8 @@ steps:
   # Run container listed in OwlBot.yaml:
   - name: ${_CONTAINER}
     dir: '${_REPOSITORY}'
+    env:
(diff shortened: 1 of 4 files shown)


sofisl:
Is there a test that asserts the default behavior without creating a label(to ensure we don't break current functionality?) Otherwise perhaps we can add it?

chingor13:
> Is there a test that asserts the default behavior without creating a label(to ensure we don't break current functionality?) Otherwise perhaps we can add it?

That's a bit harder to test - the default value can be defined by each release class. I could possibly try to replace the factory method with a sinon stub to show that we're not passing a value to the constructor.

chingor13:
@sofisl Added a `sinon` spy to check the value passed to the factory.

sofisl (approved):

PR #1755, opened 2021-05-11
feat(auto-approve): authenticate with yoshi-approver account using secret stored in secret manager

sofisl:
Following the discussion in our GH working group today. FYI, I have not yet created the secret in secret manager.


chingor13 (approved):

packages/auto-approve/test/auto-approve.test.ts
```diff
@@ -115,6 +122,7 @@ describe('auto-approve', () => {
         checkPRAgainstConfigStub.returns(true);
         validateSchemaStub.returns(undefined);
         checkCodeOwnersStub.returns('');
+        getSecretStub.returns(new Octokit({auth: '123'}));
```
tmatsuo, line 125: Can you add a spy on the Octokit so that we can test that the bot is using this Octokit?

packages/auto-approve/test/auto-approve.test.ts
```diff
@@ -290,4 +298,37 @@ describe('auto-approve', () => {
...
+
+      await probot.receive({
+        name: 'pull_request.opened',
+        payload,
+        id: 'abc123',
+      });
+
+      scopes.forEach(scope => scope.done());
+      assert.ok(getSecretStub.calledOnce);
+      assert.strictEqual(secretOctokit instanceof TestingOctokit, false);
```
tmatsuo, line 331: I don't understand this assert. As I said in the other comment, can we add a spy on the Octokit returned from `getSecretStub` and make sure the methods are called on that Octokit?

tmatsuo (changes requested):

packages/auto-approve/test/auto-approve.test.ts
```diff
@@ -290,4 +298,37 @@ describe('auto-approve', () => {
...
+      getSecretStub.returns(secretOctokit);
+
+      const payload = require(resolve(
+        fixturesPath,
+        'events',
+        'pull_request_opened'
+      ));
+
+      const scopes = [
+        getConfigFile('fake-config', 200),
```
tmatsuo, line 317: BTW, these helpers look great, thanks. Let's make it a norm.

tmatsuo (commented):
This pattern looks great, thanks.

packages/auto-approve/test/auto-approve.test.ts
sofisl: @tmatsuo, I don't believe that we need to spy on the Octokit instance, since the test wouldn't be able to make the calls to nock (for submitting the approval) if the `getSecretStub` didn't return with a valid Octokit instance. I definitely agree that this isn't immediately obvious though, and perhaps it points to a need to refactor all the tests to moving to one dependency or another (i.e., sinon > nock). But that might take more work!

This assertion is basically just saying, make sure this Octokit instance is a different type than the one we use to authenticate the bot.

tmatsuo: This assert is just comparing the local variable `secretOctokit` and the global variable `TestingOctokit`, they are different type from the start and it doesn't ensure anything in the app's code.

To make sure the bot is using the `secretOctokit` correctly, we need to use spy or mock.
If you want, I'll try to come up with the code change and push it to your branch, let me know.

tmatsuo (changes requested):
Thanks for the response. I still think we need a spy or mock. Maybe I can create a change and push it to your branch?

tmatsuo:
@sofisl I added 2 spies to make sure the correct octokit is used in the app. PTAL.

tmatsuo (approved):

Issue #1691, opened 2021-04-27
owl bot: copies from googleapis-gen to mono repos will merge/accumulate


SurferJeffAtGoogle:
As discussed with @jskeet 

With a single .OwlBot.yaml file like this:
```yaml
deep-copy-regex:
  - source: /a/b/c
    dest: ...
  - source: /x/y/z
    dest: ...
```

Imagine two commits in googleapis-gen:
1. modifies /a/b/c
2. modifies /x/y/z

If Owl Bot generates the code for 2 before the maintainer has merged 1, then the generated PR for 2 will also contain the changes for 1.  What's worse, if commit 1 contains a breaking change that should not be merged, it's going to be a nasty chore to remove it from all subsequent pull requests generated by Owl Bot.  That's not what Jon Skeet wants, nor can I imagine anyone wanting this behavior.

CC @dazuma

bcoe:
@SurferJeffAtGoogle I might argue this is a "feature", but one that's necessary before mono-repo adoption?

SurferJeffAtGoogle:
> @SurferJeffAtGoogle I might argue this is a "feature", but one that's necessary before mono-repo adoption?

Yes.  I wouldn't want to maintain a repo with owl bot before this is fixed.

SurferJeffAtGoogle:
Acknowledged.

SurferJeffAtGoogle:
Fixed by https://github.com/googleapis/repo-automation-bots/pull/2433

PR #1757, opened 2021-05-12
feat: support multiple .OwlBot.yamls in one repo

SurferJeffAtGoogle:
Also deleted CopyTasksStore, because it was never used.  To avoid duplicate copy operations, the old code searched for source commit hashes, and the new code searchs for copy tags.

Replaced the yamls table (firestore collection) with a repo-configs table.
The schema changed enough so that the new code would be incompatible with the old schema.  We must run scan-configs once before releasing the new docker image, to populate the new table.

Fixes https://github.com/googleapis/repo-automation-bots/issues/1691


packages/owl-bot/package.json
```diff
@@ -60,6 +61,7 @@
   "dependencies": {
     "@google-cloud/cloudbuild": "^2.0.6",
     "@octokit/rest": "^18.5.3",
+    "adm-zip": "^0.5.5",
```
bcoe, line 64: Wondering if we could get away with using Node's built in Zlib implementation, vs., pulling in dep?

https://nodejs.org/api/zlib.html#zlib_class_zlib_unzip

packages/owl-bot/src/handlers.ts
```diff
@@ -253,42 +250,28 @@ export async function refreshConfigs(
...
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        yaml.load(lockContent) as Record<string, any>
-      );
-    } catch (e) {
-      logger.error(
-        `${repoFull} has an invalid ${owlBotLockPath} file: ${e.message}`
-      );
-    }
-  }
+  const response = await octokit.repos.downloadZipballArchive({
```
bcoe, line 253: I'm a little concerned this might cause some memory issues in the context of the Cloud Function, my preference would be using the `getTree` API, if possible, rather than materializing the repo on disk ([example here](https://github.com/googleapis/release-please/blob/master/src/github.ts#L1551)).

bcoe (changes requested):
On a first pass, I'm a little concerned about us materializing the repos on disk, if we could get away with the tree query. The major reason being, I think mono-repos in particular will tend to be huge, and I feel a bit weird about populating their code to disk in a Cloud Function (_which would be necessary for the scan configs method).

Another question, regarding:

> The schema changed enough so that the new code would be incompatible with the old schema. We must run scan-configs once before releasing the new docker image, to populate the new table.

Does this mean that we will need to change the config in all of the repositories that have been migrated so far?

packages/owl-bot/src/configs-store.ts
```diff
@@ -97,5 +122,39 @@ export interface ConfigsStore {
...
+): [OwlBotLock | undefined, OwlBotYamlAndPath[]] {
+  let lock: OwlBotLock | undefined;
+  const yamls: OwlBotYamlAndPath[] = [];
+
+  // .OwlBot.lock.yaml is always in a known location.
+  const lockPath = path.join(dir, owlBotLockPath);
+  if (fs.existsSync(lockPath)) {
+    const lockText = fs.readFileSync(lockPath, 'utf8');
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const lockYaml = load(lockText) as Record<string, any>;
```
sofisl, line 143: nit: is it necessary to use any here? Would it be possible to create an interface instead?

sofisl (approved):

bcoe (changes requested):
The only suggestion that I think is potentially blocking for me, is that it would be nice to encapsulate the logic that finds all configs in a helper, so we could swap it out with a tree API.

With regards to `adm-zip` I'm just always hesitant to pull in new deps, if there's a chance we could use Node's built in.

---

I think my other question was answered with a re-read, the schema changes in the firestore database, not the config files?

What does our rollout look like, might need to pair with you on this.

packages/owl-bot/src/handlers.ts
SurferJeffAtGoogle: I just tried downloading zip files for google-cloud-ruby, google-cloud-dotnet, and google-cloud-php.  google-cloud-ruby was the heaviest an 20.1 MB.  That's small enough that I'm comfortable with downloading zip balls in the cloud function.  Cloud functions are limited to 4 GB of memory.

SurferJeffAtGoogle: Also, as we discussed verbally, the getTree() call has limits.  If the limits are exceeded, then a truncated tree is returned:
https://github.community/t/github-get-tree-api-limits-and-recursivity/1300

packages/owl-bot/src/copy-code.ts
```diff
@@ -333,45 +355,56 @@ export function copyDirs(
...
+  const copyTag = copyTagFrom(destRepo.yamlPath, sourceCommitHash);
+
+  // A generic function that finds matches in either pull request or issue
+  // bodies.
+  const findInBodies = (
+    kind: 'Pull request' | 'Issue',
+    response: {data: {number: number; body?: string | null}[]}
+  ): boolean => {
+    for (const issue of response.data) {
+      const copyTagPos = indexOfCopyTagFooter(issue.body ?? '');
```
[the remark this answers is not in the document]
SurferJeffAtGoogle: I added this comment at the first mention of copy-tag:

```ts
/**
 * Owl bot must avoid creating duplicate pull requests that copy code from
 * googleapis-gen.  Duplicate pull requests would annoy and confuse library
 * maintainers.
 *
 * To avoid creating duplicate pull requests, owl bot creates a unique id,
 * called a copy tag, for each copy operation.
 *
 * Before multiple .OwlBot.yaml files were permitted in a single repo, the
 * commit hash from googleapis-gen effectively functioned as a unique id
 * for the copy operation.  However, with multiple .OwlBot.yamls in a single
 * library repo, the commit hash from googleapis-gen no longer uniquely
 * identifies the copy operation.  There is potentially one copy operation for
 * each .OwlBot.yaml.  Therefore, Owl bot composes a unique copy tag comprising
 * the commit hash from googleapis-gen and the path to owlBotYaml in the library
 * repo.
 *
 * Before creating a copy-code pull request, Owl Bot first checks if a pull
 * request or issue already exists with the same copy tag.  If one exists, then
 * Owl Bot does not create a second, duplicate pull request.
 */
export function copyTagFrom(owlBotYamlPath: string, sourceCommitHash: string) 
```

packages/owl-bot/src/scan-googleapis-gen-and-create-pull-requests.ts
```diff
@@ -41,15 +39,35 @@ interface Todo {
...
+  }
   const beginIndex = beginAfterCommitHash
     ? commitHashes.indexOf(beginAfterCommitHash)
     : -1;
-  return beginIndex >= 0 && beginIndex <= commitIndex;
+  if (beginIndex >= 0 && beginIndex <= commitIndex) {
+    return true;
+  }
+  // Compare to environment variable.
+  if (process.env.OWL_BOT_BEGIN_AFTER_COMMIT_HASH) {
```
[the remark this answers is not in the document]
SurferJeffAtGoogle: It doesn't.  It was a left-over for an idea that turned out to be unnecessary.  I removed this code.

packages/owl-bot/src/copy-code.ts
```diff
@@ -46,8 +46,43 @@ function sourceLinkFrom(sourceCommitHash: string): string {
...
+ * commit hash from googleapis-gen effectively functioned as a unique id
+ * for the copy operation.  However, with multiple .OwlBot.yamls in a single
+ * library repo, the commit hash from googleapis-gen no longer uniquely
+ * identifies the copy operation.  There is potentially one copy operation for
+ * each .OwlBot.yaml.  Therefore, Owl bot composes a unique copy tag comprising
+ * the commit hash from googleapis-gen and the path to owlBotYaml in the library
+ * repo.
+ *
+ * Before creating a copy-code pull request, Owl Bot first checks if a pull
+ * request or issue already exists with the same copy tag.  If one exists, then
```
bcoe, line 67: Thanks for this comment 👍 does a great job of clarifying for me.

packages/owl-bot/package.json
SurferJeffAtGoogle: I read zlib's documentation, and I don't see a way to unpack a zip archive using zlib.  Do you see a way?  All zlib's functions appear to work on single file streams, not archives.

packages/owl-bot/src/configs-store.ts
SurferJeffAtGoogle: I haven't found a better alternative to using any here.  :-/

SurferJeffAtGoogle:
> Does this mean that we will need to change the config in all of the repositories that have been migrated so far?

Nope.  No need to update repos.

packages/owl-bot/src/copy-code.ts
```diff
@@ -46,6 +46,12 @@ function sourceLinkFrom(sourceCommitHash: string): string {
   return `https://github.com/googleapis/googleapis-gen/commit/${sourceCommitHash}`;
 }
 
+const copyTagFooter = 'Copy-Tag: ';
+
+export function indexOfCopyTagFooter(body: string): number {
```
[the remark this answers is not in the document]
SurferJeffAtGoogle: Fixed.

packages/owl-bot/src/handlers.ts
```diff
@@ -253,42 +250,28 @@ export async function refreshConfigs(
...
-    octokit
-  );
-  if (yamlContent) {
-    try {
-      newConfigs.yaml = owlBotYamlFromText(yamlContent);
-    } catch (e) {
-      logger.error(`${repoFull} has an invalid ${owlBotYamlPath} file: ${e}`);
-    }
+  const tmpDir = tmp.dirSync().name;
+  const zip = new AdmZip(Buffer.from(response.data as ArrayBuffer));
```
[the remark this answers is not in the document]
SurferJeffAtGoogle: Done.  Moved into `fetchConfigs()`.

SurferJeffAtGoogle:
Ready for another look.

PR #1758, opened 2021-05-12
feat(flakybot): add the config file and allow users to change the issue priority

### LINKED ISSUE #591: flakybot: Allow priority of label to be configured
I am looking at enabling our flaky test detection bot more widely for Node.js and, in case this ends up turning into a mountain of work, I would love to be able to configure issues as P1s vs., P2s.

tmatsuo:
fixes #591 

The `config` module is written in a way that we will be able to factor it out to gcf-utils.


packages/flakybot/src/config.ts
```diff
@@ -0,0 +1,183 @@
...
+    });
+    const pullRequestCommitSha = context.payload.pull_request.head.sha;
+    try {
+      let errorText = '';
+      const files = await context.octokit.paginate(
+        context.octokit.pulls.listFiles,
+        listFilesParams
+      );
+      for (let i = 0; files[i] !== undefined; i++) {
+        const file = files[i];
```
chingor13, line 91: nit: I think we can do `for (file in files) {}` or maybe it's `file of files`.

packages/flakybot/src/config.ts
```diff
@@ -0,0 +1,183 @@
...
+      if (!isValid) {
+        errorText = JSON.stringify(validateSchema.errors, null, 4);
+      }
+    } else {
+      errorText = `.github/${this.configPath} is not valid YAML 😱`;
+    }
+    return {isValid: isValid, errorText: errorText};
+  }
+
+  public async validateConfigChanges(context: Context): Promise<void> {
```
chingor13, line 78: If we make the parameter an octokit instance + pr info, we can drop the probot dependency when this is extracted to a common library.

packages/flakybot/src/config.ts
```diff
@@ -0,0 +1,183 @@
...
+      }
+    } catch (err) {
+      logger.error(err);
+      return;
+    }
+  }
+}
+
+export async function getConfig<T>(
+  context: OctokitContext,
```
chingor13, line 139: This one could just provide the octokit instance, rather than a probot Context.

packages/flakybot/src/config.ts
```diff
@@ -0,0 +1,183 @@
...
+  owner: string,
+  repo: string,
+  fileName: string,
+  defaultConfig?: T
+): Promise<T | null> {
```
chingor13, line 144: You can provide multiple function head type definitions. We could declare one that always returns `Promise<T>` if `defaultConfig` argument is provided.

packages/flakybot/src/flakybot.ts
```diff
@@ -120,6 +146,14 @@ export function flakybot(app: Probot) {
     const commit = context.payload.commit || '[TODO: set commit]';
     const buildURL = context.payload.buildURL || '[TODO: set buildURL]';
 
+    const config = (await getConfig<Config>(
+      context,
+      owner,
+      repo,
+      CONFIG_FILENAME,
+      DEFAULT_CONFIG
+    )) as Config; // This is safe because we pass the default config.
```
chingor13, line 155: If you do the multiple type heads, you could make this call without the explicit cast.

chingor13 (approved):
All my comments are non-blocking suggestions.

packages/flakybot/src/config.ts
tmatsuo: Done

packages/flakybot/src/config.ts
tmatsuo: Good idea.

packages/flakybot/src/config.ts
tmatsuo: Will do

packages/flakybot/src/config.ts
tmatsuo: Good idea, will do

packages/flakybot/src/flakybot.ts
tmatsuo: right, will do

packages/flakybot/src/config.ts
tmatsuo: Ended up providing different functions. The compiler complains about duplicate function.

tmatsuo (commented):
@chingor13 Thanks,
I ended up providing 2 similar functions with different names.
PTAL

chingor13 (approved):

packages/flakybot/src/config-schema.json
```diff
@@ -0,0 +1,13 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "type": "object",
+    "title": "flakybot config",
+    "description": "Schema for flakybot configuration",
+    "additionalProperties": false,
+    "properties": {
+	"defaultIssuePriority": {
```
tbpg, line 8: Should this just be `issuePriority`? This is the only config, so this isn't really the default, right?

tbpg (approved):
Thank you for doing this! One comment about the config field name.

packages/flakybot/src/config-schema.json
tmatsuo: Thanks, I hesitant to use `issuePriority` because I'm also planning to introduce dynamic priority, but default sounds incorrect.
I'll change it to `issuePriority`.

tmatsuo (commented):
- Changed the config key to issuePriority
- Use context.repo for populate the owner and the repo.

tmatsuo:
@chingor13 @tbpg Thanks for the review, merging.

PR #1759, opened 2021-05-12
fix(generated-files-bot): Use `sender` in addition to PR user

### LINKED ISSUE #1751: `generated-files-bot` should not warn if `owl-bot` commits changes to a PR
Currently `generated-files-bot` makes noise when `owl-bot` does its thing. Let's fix that.

danielbankhead:
Part of the fix for #1751

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/repo-automation-bots/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #1751 🦕
Related: https://github.com/googleapis/synthtool/pull/1080


packages/generated-files-bot/src/generated-files-bot.ts
```diff
@@ -206,10 +206,10 @@ export function handler(app: Probot) {
     const owner = context.payload.repository.owner.login;
     const repo = context.payload.repository.name;
     const pullNumber = context.payload.pull_request.number;
-    const prAuthor = context.payload.pull_request.user.login;
+    const sender = context.payload.sender.login;
 
     // ignore PRs from a configurable list of authors
-    if (config.ignoreAuthors?.includes(prAuthor)) {
+    if (config.ignoreAuthors?.includes(sender)) {
```
chingor13, line 212: We may want to check both - consider the case where owlbot opens a PR, then I update the branch from main. I _think_ I would be the event `sender` (of `pull_request.synchronize`) but the PR may be legitimately updating a generated file.

danielbankhead: Good point - what are your thoughts on handling the use case where owl-bot opens a PR, then someone edits something that shouldn't be edited? Perhaps I should refactor to capture the files that changed between commits and create messages based on those files rather than all files in the PR itself

chingor13: That might work, but it's worth investigating what GitHub reports for files changed on a merge from main. The case I'm thinking of is a maintainer updates the branch from main and the files it pulls in might have updated a templated file (not sure what the data looks like). In that case, it might not be possible to distinguish whether or not to comment unless we special-case ignore the updating of a branch from the pull request base branch.

danielbankhead: Diving into this, there doesn't appear to be a clean way to get a clean diff between merges/commits. I can come back to work on this functionality if desired, while checking both the `prAuthor` and `sender` for now

chingor13 (approved):

PR #1764, opened 2021-05-13
feat(datastore-lock)!: factoring out datastore-lock

tmatsuo:
This small library is initially implemented in `blunderbuss`, but it should be useful for some other bots, so I'm trying to make it a npm package.

### DIFF
--- a/packages/datastore-lock/src/datastore-lock.ts
+++ b/packages/datastore-lock/src/datastore-lock.ts
@@ -1,0 +1,171 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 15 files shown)


chingor13 (approved):
Will want @bcoe's feedback on naming. We can set up the release config in a followup PR.

packages/datastore-lock/src/datastore-lock.ts
```diff
@@ -0,0 +1,171 @@
...
+      waitTime = Math.min(
+        maxDelay,
+        waitTime + Math.floor(Math.random() * 2000)
+      );
+    }
+    // Timeout
+    return false;
+  }
+
+  private async _acquire(): Promise<AcquireResult> {
```
bcoe, line 105: Fun fact, I learned recently that `private` fields aren't actually private outside of the context of TypeScript, so when I actually wanted private methods, I ended up using symbols as method names:

https://github.com/yargs/yargs/blob/master/lib/yargs-factory.ts#L88

I'm not suggesting we change anything, just wanted to share this surprising TypeScript behavior.

tmatsuo: Ah it makes sense.

tmatsuo (commented):
Changed the package name to `@github-automations/datastore-lock`.
@bcoe PTAL

tmatsuo:
I'll add release config first :)

tmatsuo:
I'm trying to rewrite the first commit message to be a breaking change, but my local git client drops the exclamation mark.

I'm going to squash locally and re-push the change with a new commit message.

tmatsuo (commented):
@bcoe I guess I'm ready for the release, PTAL

PR #1765, opened 2021-05-13
chore: add release please config for datastore-lock

tmatsuo:


bcoe (approved):
I believe you also want to add an entry to this file: 

https://github.com/googleapis/repo-automation-bots/blob/master/.release-please-manifest.json

It would just start at the current version you have in your `package.json`, for the initial release.

chingor13 (approved):

Issue #1439, opened 2021-02-18
cleanup(merge-on-green): Consider to use `mergeable` and `mergeable_state` to decide whether to merge.


tmatsuo:
Currently MOG is checking the result of the github checks against a list of required check.
However, I think it's better to use `mergeable` and `mergeable_state` to decide whether we can merge the PR.

Probably these fields:
```
{
  "draft": false
  "merged": false,
  "mergeable": true,
  "mergeable_state": "clean"
}
```
plus not having `do-not-merge` label.

sofisl:
MOG kind of does that, see [here](https://github.com/googleapis/repo-automation-bots/blob/958f244f60197536a8b6f16050df8c5a61be6f58/packages/merge-on-green/src/merge-logic.ts#L612).

Basically, first we check against the list of required checks, and if it has at least one approval. If we attempt to merge and it fails, then we look into whether its mergeable and mergeable_state.

tmatsuo:
What I'm suggesting is the following:

I suspect that we actually don't need to check against the list of required checks. I think `mergeable` is true only if all the required checks are passing.

sofisl:
```
SELECT jsonPayload.repo,
  jsonPayload.number,
  jsonPayload.mergeable,
  jsonPayload.mergeable_state,
  jsonPayload.checkreviews AND jsonPayload.checkstatus as MOG_MERGEABLE,
  timestamp
FROM
  `repo-automation-bots.demo.cloudfunctions_googleapis_com_cloud_functions` 
WHERE
  resource.labels.function_name = "merge_on_green"
  AND ((jsonPayload.checkreviews AND jsonPayload.checkstatus) = false) 
  AND (jsonPayload.checkreviews AND jsonPayload.checkstatus) is not NULL 
  AND jsonPayload.mergeable_state = "clean"
  AND jsonPayload.mergeable = true
ORDER BY timestamp DESC 
```

So far, this list is not 0, however I suspect there are two major culprits:

1) These repos are not in our allowlist, 

2) We require at least one approval, and no reviewer to have left comments/requested changes previously without afterwards giving approval.

If we want to discuss opening up these requirements, then we can potentially use the `mergeable` and `mergeable_state` tags.

tmatsuo:
wontfix

PR #1772, opened 2021-05-14
chore: add metrics to confirm whether GHs mergeability is correct

sofisl:
Add logging to see if GH's mergeability definition matches our own

See  #1439

### DIFF
--- a/packages/merge-on-green/src/merge-logic.ts
+++ b/packages/merge-on-green/src/merge-logic.ts
@@ -519,6 +519,31 @@ async function commentOnPR(
   }
 }
 
+// TODO(sofisl): Remove once metrics have been collected (06/15/21)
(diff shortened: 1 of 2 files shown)


bcoe (approved):

tmatsuo:
I'm not opposed with this change, but I'm just wondering if we can just use the GH's mergeability (maybe + do not merge label).

bcoe:
> I'm not opposed with this change, but I'm just wondering if we can just use the GH's mergeability (maybe + do not merge label).

@tmatsuo that's the intention, but GitHub's documentation on the `mergeability` and `mergeable_status` is somewhat spotty, we want to track metrics for a week and make sure that it behaves exactly as we expect, before swapping over to this approach.

PR #1773, opened 2021-05-14
fix(m-o-g): grab all the last commits from a PR, not just the last 100

sofisl:
Fixes PRs like this one: https://github.com/googleapis/google-api-python-client/pull/1353

MOG isn't merging this PR because it has too many commits. This should fix it.

CC @parthea


parthea:
Thanks for finding and fixing this, @sofisl!

PR #1774, opened 2021-05-14
fix(m-o-g): grab all the last commits from a PR, not just the last 100

sofisl:
Weirdness with git in #1773, reopening here cc @parthea

Fixes PRs like: googleapis/google-api-python-client#1353


packages/merge-on-green/test/Fixtures/events/longListOfCommits.json
```diff
@@ -0,0 +1,105 @@
+[{"sha": "6dcb09b5b57875f334f61aebed695e2e4193db5e"},
```
tmatsuo, line 1: It certainly works, but can it be dynamically created by a for loop so that it's more expressive?

JustinBeckwith (approved):

PR #1776, opened 2021-05-15
fix(flakybot): use lock for updating an issue

### LINKED ISSUE #483: Flakybot: comments multiple times on the same issue at the same time
If two builds finish at the same time, separate invocations of the Build Cop Bot can lead to two (or more) duplicate comments on an issue. Similarly, two invocations can try to close the issue at the same time.

There is some logic in place to only comment with one failure per commit. However, it doesn't handle this race condition.

We should consider some sort of distributed lock to make sure only one instance of the bot comments on it at a time.

For example, see https://github.com/googleapis/google-cloud-go/issues/1840#issuecomment-601197212 and https://github.com/googleapis/google-cloud-go/issues/1915 (different types of comments, same race condition).

P2 because while this is annoying, nothing is broken and requiring immediate attention.

tmatsuo:
fixes #483

I think this will fix #483.
This PR acquires a lock per issue. Another approach is to acquire a lock per repo, but I think it'll be problematic because one build may contain many test failures and taking a long time to process.


tmatsuo:
It should use the new package `@google-automations/datastore-lock`

sofisl (approved):
Non-blocking, but perhaps best to separate out the refactor from the fix, as it's a bit difficult to parse out this PR.

tmatsuo:
@sofisl I agree in general. In this case, I needed to use try...finally and it makes the diff a bit hard to parse.

PR #1780, opened 2021-05-17
cleanup(snippet-bot): use bot-config-utils

tmatsuo:
It slightly changed the behavior where the user set `alwaysCreateStatusCheck` to true.
When `alwaysCreateStatusCheck` is set to true, the bot used to always create a successful status check for the config schema check, but now the library doesn't create a status check when the validation succeeds.

However there are virtually no users of that config except for python-docs-samples, so it should not be a problem.


tmatsuo:
I think the bot should check the config validation first thing in the pull_request hander.

tmatsuo:
We need the fix in #1785

sofisl (commented):
More of a meta-comment: I'm sure we are testing that bot-config-utils rejects incorrect files, but could we also add some tests here where we are testing the result of calling `validateConfigChanges` (it seems like we are only confirming that it is `calledOnce` or `calledWith`? Especially since this library is new, it would be nice to confirm that it's returning the results we expect. (Totally possible I missed these tests in the PR though, it's a bit lengthy :))

tmatsuo:
@sofisl Good call. I can add a test for failing status check for the config schema.

tmatsuo:
@sofisl I added some tests for broken config and correct config case. PTAL

chingor13 (approved):

Issue #1038, opened 2020-10-16
.github/header-checker-lint.yml ignoreFiles not working


aabmass:
I am using the header-checker-lint and want to ignore some config files (`.prettierrc.js`). The header checker lint doesn't seem to respect my config in `.github/header-checker-lint.yml` `ignoreFiles` array. I originally tried `.bots/header-checker-lint.json` as well but I see the docs are out of date (https://github.com/googleapis/repo-automation-bots/issues/897#issuecomment-679426370). Is there something I'm missing here?

#### Steps to reproduce

Here is a repro PR in a brand new repo: https://github.com/aabmass/gts-header-checker-lint-repro/pull/1. The header-check fails even though the failing file should be ignored.

chingor13:
Thanks for filing this!

I think what's happening is probot's `context.config()` method is reading config from the default branch (not from the PR branch). So in the example PR, the config is not in your `main` branch. This is a design decision by probot (likely to ensure the configuration is trusted by default).

It's possible in this case, that we try to read the config from the current branch.

tmatsuo:
So the workaround in this case now is to first merge `.github/header-checker-lint.yml` first then sync your PR to include that file.
I just released https://github.com/googleapis/repo-automation-bots/tree/master/packages/bot-config-utils
(which is broken right now).

Maybe I can add a capability of returning the new config from the config checker so that the bot can opt in using that config instead of the config in the main branch.

tmatsuo:
I'll add one more PR to fix this.

PR #1797, opened 2021-05-19
feat(header-checker-lint): use bot-config-utils

tmatsuo:
part of #1038

With this PR, the bot will start validating the config schema on PRs
and submit failing status checks when the config is invalid.

### DIFF
--- a/packages/header-checker-lint/test/header-checker-lint.ts
+++ b/packages/header-checker-lint/test/header-checker-lint.ts
@@ -55,8 +55,6 @@ describe('HeaderCheckerLint', () => {
     beforeEach(() => {
       payload = require(resolve(fixturesPath, './pull_request_opened'));
       getConfigStub = sandbox.stub(botConfigUtilsModule, 'getConfig');
-      // By default, getConfigStub resolves to null.
-      getConfigStub.resolves(null);
       validateConfigStub = sandbox.stub(
         ConfigChecker.prototype,
         'validateConfigChanges'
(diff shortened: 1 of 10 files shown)


tmatsuo:
I also manually tested python-docs-samples config file (it's actually a json file) locally just in case, and it worked.

packages/header-checker-lint/test/header-checker-lint.ts
```diff
@@ -39,15 +43,30 @@ describe('HeaderCheckerLint', () => {
...
     let payload: EventPayloads.WebhookPayloadPullRequest;
+    const sandbox = sinon.createSandbox();
+    let getConfigStub: sinon.SinonStub;
+    let validateConfigStub: sinon.SinonStub;
 
     beforeEach(() => {
       payload = require(resolve(fixturesPath, './pull_request_opened'));
+      getConfigStub = sandbox.stub(botConfigUtilsModule, 'getConfig');
+      // By default, getConfigStub resolves to null.
+      getConfigStub.resolves(null);
```
chingor13, line 59: nit: I might leave this unset, so we don't need to call `reset()` in all the tests that do use config. Then we would explicitly resolve `null` in the test cases that don't have config.

chingor13 (approved):
Minor (non-blocking) nit

packages/header-checker-lint/test/header-checker-lint.ts
tmatsuo: sounds good

sofisl (approved):

Issue #1673, opened 2021-04-23
Proposal: Bots should create their labels


chingor13:
For bots that listen to label events (e.g. release-please's `release-please:force-run`), the bot should create the label on the repository it's installed on at some point (installation, or perhaps on a cron). It will help discoverability of those features (the label could even link to the bot in the description).

tmatsuo:
Maybe provide a common library to do that and let the bot use the library?
`installation` hook is theoretically the best place, but probably not practical because we have many bots already installed.

The bot can call the library helper function in any of the popular handler.

It will add 1 datastore.read most of the time and when updating, datastore transactional read/write + few github API calls. It might be easier to use datastore lock.

tmatsuo:
One more thoughts on this. Probably it's safer to share the lock (and probably logic) with label-sync bot.

tmatsuo:
I came up with a really nice idea.

1. Build a common library for bot specific label
2. It provides a lightweight function for checking the label existence, and if not, file an issue on the repo informing about the bot's label.
3. Optionally that issue has a checkbox saying [ ] Create these labels.
4. If you want to use this feature, bot should listen to `issues.edited` events.
5. When user check the box, library will create these labels, then close the issue, report errors upon failure.

The most lightweight way of using the library is to use the function in 2. in the installation.
For already installed bots, bot can opt in calling the function 2 in its main event handler.

In any cases, a bot can opt in to use the checkbox and the automatic label creation.

tmatsuo:
The easier option:

1. Build a common library for bot specific label
2. It provides an async function for checking and syncing labels. The bot can call it in the main handler and await at the end.

I'd prototype this easier version.

tmatsuo:
Alright, snippet-bot opted in using the `@google-automations/label-utils` and now all the repos has the same labels.
Probably it is better to create a dedicated scheduler job because scanning GCP org taks like 3 minutes.

tmatsuo:
I did a quick scan. I think here is the list of bots that have its own labels.

- [ ] blunderbuss
- [ ] flakybot
- [ ] merge-on-green
- [ ] owl-bot
- [ ] release-please
- [x] snippet-bot

sofisl:
Adding myself in to implement for merge-on-green.

tmatsuo:
I think we're syncing all of the bot labels now.

PR #1813, opened 2021-05-20
chore: release config for label-utils

tmatsuo:
part of #1673


chingor13 (approved):
I don't think the order of the release config vs the library code actually matters.

PR #1826, opened 2021-05-21
fix(gcf-utils)!: fix getAuthenticatedOctokit

### LINKED ISSUE #1825: snippet-bot Cloud scheduler crashes with auth related error
```
GET https://api.github.com/installation/repositories" failed most likely due to lack of authentication.
```
auto_label is also getting failure.

### LINKED ISSUE #1827: test: add end 2 end authentication test for getAuthenticatedOctokit
Background is #1825 

My guess is the recent gcf-utils are broken for scheduler authentication.
It's better if we have end 2 end test for getAuthenticatedOctokit to prevent similar things to happen.

tmatsuo:
fixes #1825
fixes #1827 

This change fixed the snippet-bot scheduler run at least locally.

According to:
https://github.com/octokit/auth-app.js/

We need to pass `{auth: opts}` instead of just `opts`.

This is a significant change, so I'm marking this as a breaking change.

### DIFF
--- a/cloudbuild-test/Dockerfile
+++ b/cloudbuild-test/Dockerfile
@@ -1,0 +1,26 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


tmatsuo:
The system test succeeds on cloudbuild!
```
  GCFBootstrapper Integration
    getAuthenticatedOctokit
{"severity":"INFO","level":30,"message":"custom logging instance not enabled"}
      ✓ creates authenticated Octokit (455ms)
```

tmatsuo:
FYI, the trigger is controlled by a comment.
If non-contributor creates a PR, the trigger won't run.
When owner or contributor adds /gcbrun in the comment, it will be triggered.

bcoe (approved):
Thanks for going the extra mile and adding the system test.

PR #1850, opened 2021-05-22
feat(auto-label): migrate to bot-config-utils

tmatsuo:
The bot will start to check config file schema on PRs.

I added some config compatibility tests with real config file.


tmatsuo:
This also needs read/write access on Checks.

sofisl (approved):

PR #1868, opened 2021-05-24
feat(cron): add new schedule.global and schedule.installation cron handlers

### LINKED ISSUE #1616: Automatically detect installation ids for crons
We may have to re-wire gcf-utils to support this.

Idea: scheduled request hits a cron endpoint which fans out by requesting the [list of app installations](https://docs.github.com/en/rest/reference/apps#list-installations-for-the-authenticated-app) and dispatching new requests for each installation. These new requests would be routed to the probot handler function.

### LINKED ISSUE #1857: Crons can only be scheduled per-repository
We have some crons that we may want to run per-installation or globally. We currently can only handle per-repository (we fan-out based on the installation id provided to the cron endpoint).

chingor13:
Fixes #1616
Fixes #1857

Wait for go/github-bot-cron-scheduling


chingor13:
The general design is approved. Will need to test locally to ensure signatures are functioning as intended.

packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -342,24 +384,202 @@ describe('GCFBootstrapper', () => {
...
+      it('ensures that task is enqueued when called by scheduler for many repos', async () => {
+        await mockBootstrapper();
+        req.body = {
+          installation: {id: 1},
+        };
+        req.headers = {};
+        req.headers['x-github-event'] = 'schedule.repository';
+        req.headers['x-github-delivery'] = '123';
+        req.headers['x-cloudtasks-taskname'] = '';
+        nockListInstallationRepos();
```
tmatsuo, line 404: Normally we call `.done()` for nock scopes. Do you want to do the same?

tmatsuo (approved):

packages/gcf-utils/test/gcf-bootstrapper.ts
chingor13: Good call, it turns out some of the nock scopes were unnecessary.

packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -342,24 +384,206 @@ describe('GCFBootstrapper', () => {
...
+    });
+
+    describe('global cron', () => {
+      it('enqueues a single task for global scheduled task', async () => {
+        await mockBootstrapper();
+        req.body = {
+          cron_type: 'global',
+        };
+        req.headers = {};
+        req.headers['x-github-event'] = 'schedule.repository';
```
tmatsuo, line 558: `schedule.global` instead of `schedule.repository`?

packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -342,24 +384,206 @@ describe('GCFBootstrapper', () => {
...
+      it('ensures that task is enqueued when called by scheduler with an installation id', async () => {
+        await mockBootstrapper();
+        req.body = {
+          cron_type: 'installation',
+          installation: {
+            id: 1,
+          },
+        };
+        req.headers = {};
+        req.headers['x-github-event'] = 'schedule.repository';
```
tmatsuo, line 516: `schedule.installation`?

packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -342,24 +384,206 @@ describe('GCFBootstrapper', () => {
...
+        listInstallationRequests.done();
+      });
+
+      it('ensures that task is enqueued when called by scheduler for many installations', async () => {
+        await mockBootstrapper();
+        req.body = {
+          cron_type: 'installation',
+        };
+        req.headers = {};
+        req.headers['x-github-event'] = 'schedule.repository';
```
tmatsuo, line 490: schedule.installation?

packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -342,24 +384,206 @@ describe('GCFBootstrapper', () => {
...
+    });
+
+    describe('per-installation cron', () => {
+      it('ensures that task is enqueued when called by scheduler', async () => {
+        await mockBootstrapper();
+        req.body = {
+          cron_type: 'installation',
+        };
+        req.headers = {};
+        req.headers['x-github-event'] = 'schedule.repository';
```
tmatsuo, line 467: schedule.installation?

tmatsuo (commented):
Does the value of `x-github-event` matter?

packages/gcf-utils/test/gcf-bootstrapper.ts
chingor13: This is actually hard-coded by serverless-scheduler-proxy: https://github.com/googleapis/repo-automation-bots/blob/e74ce012113a13631676cca20e1c4f148d19627e/serverless-scheduler-proxy/main.go#L76

When we queue the task, we change the name to the appropriate `schedule.type`.

packages/gcf-utils/test/gcf-bootstrapper.ts
chingor13: same as above

packages/gcf-utils/test/gcf-bootstrapper.ts
chingor13: same as above

packages/gcf-utils/test/gcf-bootstrapper.ts
chingor13: same as above

chingor13:
> Does the value of `x-github-event` matter?

The name passed to `probot.receive()` is what routes to the app handlers: 
https://github.com/googleapis/repo-automation-bots/blob/e74ce012113a13631676cca20e1c4f148d19627e/packages/gcf-utils/src/gcf-utils.ts#L333 

We parse this from the `x-github-event` header at the start of the gcf wrapper. https://github.com/googleapis/repo-automation-bots/blob/e74ce012113a13631676cca20e1c4f148d19627e/packages/gcf-utils/src/gcf-utils.ts#L279-L280

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -360,58 +373,266 @@ export class GCFBootstrapper {
...
+    }
+  }
+
+  /**
+   * Async iterator over each repository for an app installation.
+   *
+   * See https://docs.github.com/en/rest/reference/apps#list-repositories-accessible-to-the-app-installation
+   * @param wrapOptions {WrapOptions}
+   */
+  private async *eachInstalledRepository(
```
bcoe, line 448: Love the use of async iterators, vs., our old approach of sometimes just grabbing the first page of results with octokit.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -360,58 +373,266 @@ export class GCFBootstrapper {
...
+  ) {
+    const octokit = await this.getAuthenticatedOctokit(
+      installationId,
+      wrapOptions
+    );
+    const installationRepositoriesPaginated = octokit.paginate.iterator(
+      octokit.apps.listReposAccessibleToInstallation,
+      {
+        mediaType: {
+          previews: ['machine-man'],
```
bcoe, line 460: Might be worth a comment with a link to what `machine-man` gives us access to.

bcoe (approved):
A few non blocking recommendations.

tmatsuo (approved):

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -360,58 +373,266 @@ export class GCFBootstrapper {
...
+    } else {
+      const generator = this.eachInstallation(wrapOptions);
+      for await (const installation of generator) {
+        const extraParams: Scheduled = {
+          installation: {
+            id: installation.id,
+          },
+        };
+        if (
+          installation.target_type === 'Organization' &&
```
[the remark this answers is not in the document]
chingor13: The `schedule.installation` hook would still be run via a Cloud Task, but would not have the additional `cron_org` parameter that we've been using by convention.

packages/gcf-utils/src/gcf-utils.ts
chingor13: I don't believe this is necessary any more, but it was moved from the previous request. I think when we first implemented the fanout, the list repos for installation API was under preview.

Issue #131, opened 2019-10-23
trusted-contribution: Don't run for repositories without kokoro


crwilcox:
This bot is currently running on repositories without kokoro which is a bit awkard.

We can look if there is a .kokoro folder present and if so add the label instead of doing it just on contributor.

tbpg:
Not all repos have the `.kokoro` folder at the root. For example, https://github.com/GoogleCloudPlatform/golang-samples/tree/master/testing/kokoro.

bcoe:
@chingor13 have you found that the `search` approach you used for finding `pom.xml` works? maybe we could use this for kokoro.

bcoe:
👋 I switched this to a feature request, in practice I haven't heard too many complaints about trusted contributor bot not doing what it needs to do, is this still something we need to fix?

JustinBeckwith:
Ok bad idea time.  What if checked each repository, and specifically looked for ones with a webhook configured to hit `http://35.184.182.122:8080`?  I think you need that webhook for kokoro to work, so maybe that's a good global thing to check?  

I would of course try to detect this once, and then cache the result inside Datastore with a TTL of 24 hours or something.

tmatsuo:
Looking at:
https://github.com/googleapis/repo-automation-bots/tree/master/packages/trusted-contribution

I'm a bit confused why this is a problem.
Is this because we enabled this bot at the org level?

tmatsuo:
I mean, if this causes problems because we enabled this bot at the org level, can we just do a one time sweep and stop org level configuration and add repo level configuration files?

tmatsuo:
I think that it is still possible to implement automatic detection for whether the repo is using Kokoro or not, but I don't think this gives enough value justifying the cost.

If we think long term, there will be a high chance that other CI systems start introducing manual trigger (the label for Kokoro builds) for builds for contributions.

I'd like to expand this bot's capability to handle such manual triggers for other CIs.

tmatsuo:
There's no org level configuration, but it seems like the bot has the default trusted contributors list and it uses the default list.
Presumably the bot is installed at the org level.

I think we need some designs for solving this situation.

tmatsuo:
FYI, for a workaround, I added `disabled:` field in the config file, so if you add `.github/trusted-contribution.yml` with the following:

```
disabled: true
```

The bot will ignore that repo.

tmatsuo:
[...]

PR #1873, opened 2021-05-25
feat(trusted-contribution): migrate to bot-config-utils

tmatsuo:
The bot will start validating config schema on PRs.
It will create a failing status check if the config doesn't follow the schema.

I also added `disabled` field for opting out from this bot.
Related: #131

The bot needs Read/Write permission on Checks.
I added the permission to the bot. The org admins need to approve the requests.


tmatsuo:
To opt out from the trusted-contributor bot, I added a new field `disabled`.
Another idea is to allow empty list for annotations.

```yaml
annotations: []
```

I think `disabled` is more expressive.

chingor13:
Maybe `enabled: false` is better language?

tmatsuo:
@chingor13 Possible, but we'll have to make sure the default value is set to true, and more tests needed.

chingor13 (approved):

PR #1887, opened 2021-05-25
feat: pull the lock-update commit message from the docker image

SurferJeffAtGoogle:
Look for the file `post-processor-changes.txt` in the docker image.  If it exists, use it as the commit message when updating lock files.


tmatsuo (approved):

bcoe (approved):
This looks like a good approach to me 👍

PR #1891, opened 2021-05-26
fix: ensure that release PRs are not merged on a Friday

sofisl:
@bcoe, following up on the JS Client libraries team meeting. This PR ensures that release PRs are not merged on Friday or the weekend. There's still a TODO, which is to create a scheduled job that 'picks up' these hanging PRs, but that can be left as an issue: #1892


tmatsuo:
This is a nice feature in general, but what about the timezone? Probably it's fine because it's opt in...

One idea is to make this feature flexible and generic is to accept time window with timezone from the user, and maybe provide a shortcut for pacific timezone workday.

tmatsuo:
One more caveat. Many web application platform like App Engine (or likely Cloud Functions too) uses UTC by default.

parthea:
cc @busunkim96

parthea:
Great feature! It would be helpful to also include have the ability to prevent auto-approve from running on holidays and freeze periods (through a configuration file).

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -181,3 +181,18 @@ export function checkFilePathsMatch(
...
+
+/**
+ * Checks whether the attempted merge is happening during the week.
+ *
+ * @returns true if merge is happening on the weekday, false if it's on Friday or the weekend.
+ */
+export function mergesOnWeekday(): boolean {
+  const date = new Date(Date.now());
+  const dayOfWeek = date.getDay();
+  console.log(dayOfWeek);
```
bcoe, line 193: I'd drop the log and add a comment 👍

packages/auto-approve/src/check-pr.ts
```diff
@@ -163,7 +165,8 @@ function runVersioningValidation(addtlValidationFile?: {
     majorBump = isMajorVersionChanging(versions);
     minorBump = isMinorVersionUpgraded(versions);
     oneDependencyChanged = isOneDependencyChanged(addtlValidationFile.file!);
+    weekdayMerge = mergesOnWeekday();
```
bcoe, line 168: nit: I think this would read better if we called this method, `isReleaseDay()`, and `releaseDay` (for the variable respectively). I think this explains what's going on well.

bcoe (changes requested):

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -181,3 +181,18 @@ export function checkFilePathsMatch(
...
   return filesMatch;
 }
+
+/**
+ * Checks whether the attempted merge is happening during the week.
+ *
+ * @returns true if merge is happening on the weekday, false if it's on Friday or the weekend.
+ */
+export function mergesOnWeekday(): boolean {
+  const date = new Date(Date.now());
```
tmatsuo, line 191: My guess is that the date is in UTC timezone, but I may be wrong. Can you confirm?
If it in UTC, you should convert to pacific time zone, right?

tmatsuo (changes requested):

packages/auto-approve/src/check-pr.ts
SurferJeffAtGoogle: I actually prefer the current language, weekday.  If I'm reading the code, "release day" tells me nothing.  "Week day" gives me the right idea, even if it isn't perfectly exact.

SurferJeffAtGoogle (approved):

packages/auto-approve/src/utils-for-pr-checking.ts
bcoe: I suggest using dayjs with the timezone extension:

https://day.js.org/docs/en/timezone/timezone

Works something like this:

`dayjs(new Date()).tz('Australia/ACT').$D`.

sofisl: Thanks for that tip!

sofisl:
@bcoe @tmatsuo, sorry for taking so long, addressed your comments!

tmatsuo (approved):

tmatsuo:
Code looks good. Just to confirm, `Date()` returns date in UTC on Cloud Function, right?

bcoe (approved):

sofisl:
> Code looks good. Just to confirm, `Date()` returns date in UTC on Cloud Function, right?

Are you referring to `const dayOfWeek = date.day();`? This is returning it in PST, since we set the timezone right before that. I thought that's what we preferred; would you prefer UTC instead?

tmatsuo:
@sofisl I want Pacific time.

I was talking about the original code without dayjs.

Please go ahead and merge.

Issue #1877, opened 2021-05-25
bug: Owl-bot post processor check missing in PR from `python-dialogflow`


parthea:
In [this PR](https://github.com/googleapis/python-dialogflow/pull/304) from `python-dialogflow` the owl bot post processor didn't run.  I've seen the owl bot check run successfully in [another PR](https://github.com/googleapis/python-dialogflow/pull/298) that I reviewed. How do I go about troubleshooting why the owl bot check was missing? I triggered the post processing check to run by adding the label `owlbot:run` so we don't block the API feature from being merged.

sofisl:
@SurferJeffAtGoogle, I've labeled this as a docs issue, but feel free to relabel as appropriate.

SurferJeffAtGoogle:
I expected owl bot post processor to run, but it never did.

bcoe:
I dug into the specific issue having trouble, and the reason `owlbot:run` did not run on subsequent runs was that the branch was deleted.

Please let us know in chat if this happens again, and we can look at the build history and see why the trigger failed.

PR #1901, opened 2021-05-27
fix(owlbot): start to address issues with owlbot:run

bcoe:
Trying to debug #1877 bumped into issues with adding the label `owlbot:run`, this will hopefully make them a bit easier to dig into.

### DIFF
--- a/packages/owl-bot/src/owl-bot.ts
+++ b/packages/owl-bot/src/owl-bot.ts
@@ -117,7 +117,9 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
         repo,
       });
     } catch (err) {
-      logger.error(`${err.message} head = ${head} pr = ${prNumber}`);
+      logger.error(
+        `removeLabel: ${err.message} ${err.code} head = ${head} pr = ${prNumber}`
+      );
     }
     logger.metric('owlbot.run_post_processor');
   });
--- a/packages/owl-bot/src/owl-bot.ts
+++ b/packages/owl-bot/src/owl-bot.ts
@@ -67,7 +67,7 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
   app.on(['pull_request.labeled'], async context => {
     const head = context.payload.pull_request.head.repo.full_name;
     const base = context.payload.pull_request.base.repo.full_name;
-    const [owner, repo] = head.split('/');
+    const [owner, repo] = base.split('/');
     const installation = context.payload.installation?.id;
     const prNumber = context.payload.pull_request.number;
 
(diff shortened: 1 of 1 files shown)


packages/owl-bot/src/owl-bot.ts
```diff
@@ -88,7 +88,15 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
       );
       return;
     }
-
+    logger.metric('owlbot.run_label', {
+      head,
+      base,
```
bcoe, line 93: Added debug line so that I can see why the heck we're getting `Not Found` for `.OwlBot.lock` so frequently in logs.

packages/owl-bot/src/owl-bot.ts
```diff
@@ -101,16 +109,18 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
...
-    });
+    try {
+      await context.octokit.issues.removeLabel({
+        name: OWLBOT_RUN_LABEL,
+        issue_number: prNumber,
+        owner,
+        repo,
+      });
+    } catch (err) {
+      logger.error(`${err.message} head = ${head} pr = ${prNumber}`);
```
bcoe: Once we fail once, we've already removed the `OWLBOT_RUN_LABEL` and so get errors on subsequent runs, I'll loosen this error as soon as I confirm the error code that's thrown.

packages/owl-bot/src/owl-bot.ts
```diff
@@ -67,7 +67,7 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
   app.on(['pull_request.labeled'], async context => {
     const head = context.payload.pull_request.head.repo.full_name;
     const base = context.payload.pull_request.base.repo.full_name;
-    const [owner, repo] = head.split('/');
+    const [owner, repo] = base.split('/');
```
tmatsuo, line 70: This is a significant change, and it should be noted in the commit message.

packages/owl-bot/src/owl-bot.ts
```diff
@@ -101,16 +109,20 @@ export = (privateKey: string | undefined, app: Probot, db?: Db) => {
...
-    });
+    try {
+      await context.octokit.issues.removeLabel({
+        name: OWLBOT_RUN_LABEL,
+        issue_number: prNumber,
+        owner,
+        repo,
+      });
+    } catch (err) {
+      logger.error(
```
tmatsuo, line 120: I'm more comfortable if we only ignore 404 (expected) and re-throw anything else?

bcoe: I just wanted to make sure that the 404 code is actually populated, then I'll tighten. Not sure what the shape of the error object is. I might test locally first and send a different PR.

tmatsuo: https://github.com/googleapis/repo-automation-bots/blob/83ae51929068927043a037b46b22ce7e5364dda8/packages/snippet-bot/src/snippet-bot.ts#L698

You can add a test emulating 404 response with nock.

PR #1903, opened 2021-05-27
fix(owlbot): do not crash on missing label

bcoe:
If we retry running as a result of `owlbot:run`, we should not crash if label has already been removed.


tmatsuo (changes requested):
Can you add tests?

It's very likely works as expected.

But 'it will very likely work' and 'it's tested' are different thing.

tmatsuo (approved):

PR #1905, opened 2021-05-27
fix: copy .dot files, but ignore .git/ directory

SurferJeffAtGoogle:
Owl Bot was not copying files or directories from googleapis-gen that begin with a dot.
The synth.py for nodejs-firestore requires those files to be copied.  At the same time, we never want to copy an entire .git/ directory.


SurferJeffAtGoogle:
DRAFT pull request because I don't want to merge it before the long weekend.  We can merge it 6/1.

JustinBeckwith (approved):

Issue #1908, opened 2021-05-28
Warning: `Octokit` in `@octokit/rest@18.5.4` has incompatible type


tmatsuo:
If one of our bots pulls `@octokit/rest@18.5.4`, it will have incompatible `Octokit` type.

The easy fix is to pin `@octokit/rest@18.5.3`, but it's not a long term solution.

I filed a bug on upstream:
https://github.com/octokit/rest.js/issues/54

We need to establish a long term solution.

tmatsuo:
I haven't seen for a while with other bots. I suspect it only happens when you install the library from the filesystem.

tmatsuo:
Downgrading to P2 until we see this regularly.

tmatsuo:
I just deployed snippet-bot with gcf-utils@10.0.0, which has `@octokit/rest@18.5.6` or something.

There's no problem.

I think it's ok if we use `@octokit/xxx` comes with gcf-utils.

Closing.

Issue #1884, opened 2021-05-25
feat(sync-repo-settings): add org level configs


tmatsuo:
Background:

Now sync-repo-settings has the global config in the source code.
It is great if we can put org level config in `ORG/.github/` repo.

The caveat is that, org level config might have a different schema from the repo level settings (probably object keyed by language or sth), so a simple falling back to `.github` repo strategy doesn't work.

Maybe we can have a schema that allows a list and the config at the same time, or we can have a different config file name.

tmatsuo:
My proposal:

1. Have a different schema and filename for org level config.
2. Modify the function for fetching config
    2-1. First try to fetch the repo config
    2-2. Upon 404, it will fetch the org level config from the `.github` repo and try to get the config from the inferred language
3. We still need to fallback to the global config
4. Once we have org level configs both for googleapis and GCP, we can remove the global config in the source code.

@JustinBeckwith How does it sound?

JustinBeckwith:
In general, that sounds great.  A big open question in my mind is how flexible we want to be in terms of how users select which repositories they want a given config to apply to.  Today, we just provide per-language selection.  I was experimenting with a more open format that let users filter on any property in the repo (is it private, is it archived, does it have a repository topic, regex matches on names, etc).  The problem is, that gets *complicated* very quickly.  I don't know how far we want to go with config selection in the org level config, but your approach is generally where I was trying to go.

tmatsuo:
I see.
The suggested selector has `org:googleapis is:public archived:false`.

For organization, it is determined by the `.github` repo's organization.
For public/private, do we ever want to sync with private repos?
For archived repo, do we ever want to sync with archived repos?

Actually now the scheduler job is skipping archived repository.

So maybe we can allow any key, starting with `language:` selectors?

```json
{
  "language:java": { //Java config here },
}
```

Then later, we can add other keys when necessary, something like?

```json
{
  "archived:true": { //Config for archived repo here },
}
```

tmatsuo:
Or we can still preserve the similar format suggested in https://github.com/googleapis/.github/pull/10/files#diff-3441059af11c9bb893d42a2eae9f0782ffeb3fb4dcc1da0a98d93a1c1f2bc250

```yaml
- name: "java"
  selector: "language:java"
  config:
    branchProtectionRules: #...
- name: "python"
  selector: "language:python"
  config:
    branchProtectionRules: #...
- name: "nodejs"
  selector: "language:javascript language:typescript"
[...]

PR #1910, opened 2021-05-28
feat(sync-repo-settings): use bot-config-utils

### LINKED ISSUE #1886: bug(sync-repo-settings): cli doesn't handle 404
It should be easily fixed by using the helper.

tmatsuo:
fixes #1886

start validating the config schema when fetching it.

Also, because the org level configuration will be likely different
format, the bot stopped to fallback to org .github repo.

I added some compatibility tests with the current real world configs.

Note: I needed to pin `@octokit/rest` until we solve #1908.

Also towards #1884


tmatsuo:
It should cherrypick #1929

JustinBeckwith (approved):

PR #1912, opened 2021-05-29
feat(object-selector)!: initial release of object-selector

tmatsuo:
This is a new utility package for selecting object. Primary use case is sync-repo-setting's org level config for selecting repositories.
It will allow schema safe way of defining various conditions in a config file (see samples in recipes directory).

However, this library should be usable for selecting other type of objects.

Part of #1884


packages/object-selector/README.md
```diff
@@ -0,0 +1,152 @@
+# object-selector
+
+This is a small library for selecting objects.
```
SurferJeffAtGoogle, line 3: This would be a great place to provide the context I lacked and asked for during our meeting?  Who uses this code?  When?  Why?  How?  What are the inputs and outputs?

SurferJeffAtGoogle (changes requested):

packages/object-selector/README.md
tmatsuo: Thanks, I updated the README. PTAL

tmatsuo (commented):
@SurferJeffAtGoogle Thanks, I updated the README, PTAL

SurferJeffAtGoogle (approved):

PR #1925, opened 2021-06-02
chore: add the "owlbot:run" label to labels list

busunkim96:


tmatsuo (approved):
Thank you for the PR, for now it will add a certain value.

In a long run, I'd like to use label-utils to sync bot labels for several reasons.

1. Single source of truth
2. Each bot knows which repos are relevant better than label-sync bots

PR #1945, opened 2021-06-04
feat(blunderbuss): start syncing blunderbuss label

tmatsuo:
part of #1673

### DIFF
--- a/packages/blunderbuss/test/blunderbuss.ts
+++ b/packages/blunderbuss/test/blunderbuss.ts
@@ -113,12 +113,18 @@ describe('Blunderbuss', () => {
         },
         id: 'abc123',
       });
+      // I think calledOnceWith is broken, demonstrating it's behavior.
+      // Compare them with the strictEqual calls below.
+      // It might only checking type, then it is WAI.
       syncLabelsStub.calledOnceWith(
         sinon.match.instanceOf(Octokit),
-        'testOwner',
-        'testRepo',
+        'it does not matter',
+        'what values I put here',
         utilsModule.BLUNDERBUSS_LABELS
       );
+      assert.strictEqual(syncLabelsStub.getCalls()[0].args[1], 'googleapis');
+      assert.strictEqual(syncLabelsStub.getCalls()[0].args[2], 'testRepo');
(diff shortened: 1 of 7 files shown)


tmatsuo:
It might be safer to cherrypick #1944.

packages/blunderbuss/test/blunderbuss.ts
```diff
@@ -93,6 +96,32 @@ describe('Blunderbuss', () => {
...
+          },
+          organization: {
+            login: 'googleapis',
+          },
+        },
+        id: 'abc123',
+      });
+      syncLabelsStub.calledOnceWith(
+        sinon.match.instanceOf(Octokit),
+        'testOwner',
```
tmatsuo, line 118: It seems like calledOnceWith is broken?

I think I need to call `getCalls()` or something.

kurtisvg (approved):

packages/blunderbuss/test/blunderbuss.ts
```diff
@@ -710,14 +747,19 @@ describe('Blunderbuss getConfigWithDefault', () => {
...
       'testOwner',
       'testRepo',
       'c5b0c82f5d58dd4a87e4e3e5f73cd752e552931a',
       6
     );
-    assignStub.calledOnceWith(sinon.match.instanceOf(Context), sinon.match.any);
+    sinon.assert.calledOnceWithExactly(
+      assignStub,
+      sinon.match.instanceOf(Context),
+      sinon.match.any
```
sofisl, line 761: Seems strange that we're confirming whether or not something was called with 'any', I feel like it might defeat the purpose of the method. But not a blocker!

tmatsuo:
We should cherrypick #1956

packages/blunderbuss/test/blunderbuss.ts
tmatsuo: I changed it to sinon.match.object, a slightly narrower version. PTAL

tmatsuo:
@sofisl I changed to use sinon.match.object and cherrypicked the new label-utils. PTAL

sofisl (approved):

Issue #1946, opened 2021-06-04
bug(tests): incorrect usage of sinonStub.calledOnceWith


tmatsuo:
Doh, I misunderstood how to use `calledOnceWith`.
I think I should use `sinon.assert.calledOnceWithExactly` in most cases.

Here is the list of bots with wrong usage.
- [x] auto-label
- [x] blunderbuss
- [x] flakybot
- [x] generated-files-bot
- [x] header-checker-lint
- [x] snippet-bot
- [x] trusted-contribution

tmatsuo:
I think I've fixed all of them in the current PRs. I'll check it once we merged all of them.

PR #1948, opened 2021-06-04
fix(header-checker-lint, trusted-contribution): use sinon.assert

tmatsuo:
part of #1946

### DIFF
--- a/packages/header-checker-lint/test/header-checker-lint.ts
+++ b/packages/header-checker-lint/test/header-checker-lint.ts
@@ -15,7 +15,6 @@
 import {resolve} from 'path';
 // eslint-disable-next-line node/no-extraneous-import
 import {Probot, createProbot, ProbotOctokit} from 'probot';
-import {Octokit} from '@octokit/rest';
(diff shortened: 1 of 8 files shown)


packages/header-checker-lint/test/header-checker-lint.ts
```diff
@@ -15,7 +15,6 @@
 import {resolve} from 'path';
 // eslint-disable-next-line node/no-extraneous-import
 import {Probot, createProbot, ProbotOctokit} from 'probot';
-import {Octokit} from '@octokit/rest';
 import snapshot from 'snap-shot-it';
```
sofisl, line 18: just confirming, it seems like we weren't using this?

sofisl (approved):

packages/header-checker-lint/test/header-checker-lint.ts
tmatsuo: I used Octokit in the previous code, but it started to fail with sinon.assert. I changed it to ProbotOctokit and now it's passing.

PR #1966, opened 2021-06-08
chore(deps)(merge-on-green): update lock file

sofisl:


sofisl:
I think we're blocked [here](https://github.com/probot/octokit-auth-probot/issues/120).

JustinBeckwith (approved):

PR #1971, opened 2021-06-08
(chore)(deps): update lock file header-checker-lint

sofisl:


JustinBeckwith:
Looks like this needs an `npm run fix`

PR #1974, opened 2021-06-08
chore(merge-on-green): use Octokit from @octokit/rest

tmatsuo:


JustinBeckwith (approved):

packages/merge-on-green/src/merge-logic.ts
```diff
@@ -12,11 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// eslint-disable-next-line node/no-extraneous-import
-import {ProbotOctokit} from 'probot';
 import {logger} from 'gcf-utils';
 
-type OctokitType = InstanceType<typeof ProbotOctokit>;
+// eslint-disable-next-line node/no-extraneous-import
+import {Octokit} from '@octokit/rest';
```
danielbankhead, line 18: What are your thoughts on exporting `Octokit` from `gcf-utils` or using a utility like [`ReturnType`](https://www.typescriptlang.org/docs/handbook/utility-types.html#returntypetype) to dynamically get the `Octokit` we want? This may prevent the linter disable line and potential issues if multiple packages in `package.json` import `@octokit/rest` at different versions

tmatsuo: Probably a good idea

tmatsuo: But there's a downside of deep dependencies.

For example, in the future we may have another library for cloud run. I don't know if we'll keep using gcf-utils, but just depending on `@octokit/rest` sounds a lighter and nicer dependency.

Personally I keep it light because what we win is to reduce a tiny annoyance of having lint comment.

sofisl (approved):

danielbankhead (approved):

Issue #1955, opened 2021-06-07
auto-label: cron job for auto-label failed on Jun 7


sofisl:
<img width="1633" alt="Screen Shot 2021-06-07 at 12 25 52 PM" src="https://user-images.githubusercontent.com/55454395/121063102-845cfc00-c78b-11eb-864d-d9767580bc5a.png">

tmatsuo:
I found some log for 500 response of scheduler proxy:

```
{
  "insertId": "60c0750700043a39afa7dc1d",
  "httpRequest": {
    "requestMethod": "POST",
    "requestUrl": "https://serverless-scheduler-proxy-c7mcaca6sa-uc.a.run.app/v0",
    "requestSize": "1427",
    "status": 500,
    "responseSize": "1321",
    "userAgent": "Google-Cloud-Scheduler",
    "remoteIp": "35.187.132.166",
    "serverIp": "216.239.36.53",
    "latency": "6.928120958s",
    "protocol": "HTTP/1.1"
  },
  "resource": {
    "type": "cloud_run_revision",
    "labels": {
      "configuration_name": "serverless-scheduler-proxy",
      "service_name": "serverless-scheduler-proxy",
      "location": "us-central1",
      "project_id": "repo-automation-bots",
      "revision_name": "serverless-scheduler-proxy-00644-hiz"
    }
  },
  "timestamp": "2021-06-09T08:00:07.277049Z",
  "severity": "ERROR",
  "labels": {
    "instanceId": "00bf4bf02db02869557118393dcd9ed7180ec8f389c5b3adbc3bc6879e21a9fd28e7e9b93e888f59b6a728570b0c95b6f4bb871d311bac045baf"
  },
  "logName": "projects/repo-automation-bots/logs/run.googleapis.com%2Frequests",
  "trace": "projects/repo-automation-bots/traces/bfaa3d066f01eeea7d40780ce2d2e8e4",
  "receiveTimestamp": "2021-06-09T08:00:07.595822506Z"
}
```

I could not find any other useful logs.

tmatsuo:
Looking at the cloud functions log, I think auto-label scheduler job is called:

```
{
insertId: "000000-7bea856f-b058-49e2-bf2a-6de2081ad64e"
jsonPayload: {
trigger: {
github_delivery_guid: "59f06c0d-9ada-4623-9f57-fe84f57f422d"
trigger_type: "Cloud Task"
}
level: 30
message: "running for org googleapis"
}
[...]

PR #1986, opened 2021-06-09
fix(gcf-utils): create tasks in parallel

tmatsuo:
Part of #1955


tmatsuo:
I needed to pin "@octokit/auth-app": "3.4.0", so I filed:
https://github.com/googleapis/repo-automation-bots/issues/1987

sofisl (approved):

PR #2019, opened 2021-06-10
chore: silence snippet-bot

tmatsuo:


bcoe (commented):
It feels odd to me that someone would need to opt out of getting the message:

> No region tags are edited in this PR.

If no snippets were edited it feels a bit like spam to leave a message.

If we think there is value to having the message when no snippets are edited for some repos, I think people should opt into this rather than needing to opt out.

tmatsuo:
Ok, I'll go with a long term solution

Issue #2021, opened 2021-06-10
snippet-bot: pull request that has never touched region tags should not have comment added


bcoe:
When a pull request is made editing a file in a repository that is not related to samples, as an example updating configuration. It feels a bit strange that snippet bot leaves a comment (see #2020).

This creates additional confusion for repositories which have no samples, e.g.,

* gax-nodejs.
* google-auth-library.
* repo-automation-bots.
* release-please
* gaxios
* teeny-request
* discovery-artifact-manager
* code-suggester
* node-gtoken

It also creates confusion for handwritten libraries, where code being added may or may not be touching samples, e.g.,

* nodejs-spanner
* nodejs-storage

---

Perhaps rather than adding the snippet bot comment, we could add an emoji to the PR, like merge-on-green does?

Or perhaps the status check is enough?

bcoe:
Another case that I just saw:

https://github.com/googleapis/nodejs-datastore/pull/788

tmatsuo:
Does it really annoying to see the comment?
It adds
- a ui for refreshing
- certain information that the PR is not modifying any region tags

Also, this is tricky because the same PR might have modified region tags in the past.

In this case, we should update the comment.

Anyways, I'll add config files for those repositories tomorrow.

tmatsuo:
Maybe a reasonable fix is to only update the comment if no region tags are edited.

It'll need to ditch addOrUpdateIssueComment from gcf-utils.

tmatsuo:
Ah I can add an option to addOrUpdateIssueComment in gcf-utils.

PR #2024, opened 2021-06-11
feat(gcf-utils)!: add onlyUpdate option to addOrUpdateIssueComment

tmatsuo:
part of #2021

Then we can update snippet-bot to use this option when there's no region tag changes in a PR.


bcoe (approved):
thanks for doing this work 😊

PR #2027, opened 2021-06-11
feat(auto-approve): add logic to check incoming renovate-bot PRs for Node

sofisl:
This PR adds logic to check that incoming renovate-bot PRs conform to the following guidelines for auto-approving by:

1. Checking the authorship
2. Checking the title starts with 'chore(deps)' or 'fix(deps)'
3. Ensuring that files changed are only ones that contain dependencies (i.e., ending with package.json)
4. Ensuring that only max 2 files are changed (samples/package.json and package.json)
5. Ensuring that all files changed meet the following criteria:
     a. They only have 2 changes listed (one deletion, one addition)
     b. The dependency changed matches the original dependency listed, which also matches what is said to be changed in the title
     c. Versions are upgraded and not downgraded
     d. Versions are not upgraded to majors

A lot of the actual logic was already written for automerging releases. Most of the work done was adding 5b, and refactoring the logic so that we are now checking different processes (i.e., a process for releases, vs. dependency upgrades, etc.)

It should be very easy to expand this to other languages if the logic described above is the same. All we would need to do is add specific rules to the ./language-versioning-rules.json file (cc @parthea).


packages/auto-approve/src/check-pr.ts
```diff
@@ -65,12 +66,16 @@ export async function checkPRAgainstConfig(
...
-    const versionChecks = true;
+    let addtlRules = true;
+    const releasePRFiles: File[] = [];
 
     // Since there's only one allowed title per author right now, we don't need to
     // add complicated logic to see which title should match the incoming PR; but,
     // this could be logic we work into the future.
     if (!title.match(rulesToValidateAgainst.title)) {
+      logger.info(
+        `Info for ${repoOwner}/${repo}/${prNumber} title does not match what is allowed`
```
tmatsuo, line 77: BTW, you fetch repoOwner from `pr.pull_request.head.repo.owner.login`, but it seems wrong.

For example, if I create a PR from my personal fork, the repoOwner will be `tmatsuo`.

tmatsuo (commented):
Let me add 2 comments first. I'll review more shortly.

tmatsuo (commented):
adding one more comment, still reviewing

packages/auto-approve/src/check-pr.ts
tmatsuo: I guess this is why I only get auto-approve config check failure for PRs from the same repo and I don't get the failing check on PRs from my personal repo.

packages/auto-approve/src/valid-pr-schema.json
```diff
@@ -30,7 +30,8 @@
             "yoshi-automation",
             "release-please[bot]",
             "yoshi-code-bot",
-            "gcf-owl-bot[bot]"
```
tmatsuo, line 33: Not related to the changes in this PR, but I noticed several things in this schema definition, so I'm pointing out here.


- Can you add the version of the json schema spec in `$schema`?
- Can you make the `$id` something more unique (e.g. `@google-automations/auto-approve/v1`)?
- `rules` have `additionalProperties`, but it doesn't work for an array.
- I think you should add `additionalProperties`: false on the `aPR` object
- Then you need to add other properties to `properties` definition
- `author` is defined as `enum`, but I think it's too strict. Can it be just a string and validate the value in the app? I also see the benefit of making it very strict, so not a blocker.
- Question: `if` and `then`. Do you have a link to the documentation on `if` and `then`?
- You're restricting the value of the title with a const. This is problematic because when one of the bot (for example googleapis-publisher) changes it's PR title, we'll need to 1) update the schema definition, and 2) update the config files in multiple repositories? Also after we updated the config schema, all the PRs editing auto-approve.yaml will fail because of this schema change. Which seems pretty bad.
- The above point also applied to changedFiles. What if the bot changes the filename?

tmatsuo (commented):
This time I reviewed the config schema, still reviewing the other parts of the PR.

packages/auto-approve/src/valid-pr-schema.json
tmatsuo: added

- The name `aPR` seems not intuitive because it represents a condition for auto mergeable PRs. How about `autoApproveCondition` or sth?

packages/auto-approve/src/check-pr.ts
```diff
@@ -84,20 +89,82 @@ export async function checkPRAgainstConfig(
...
 
     // If we've found the file in this ruleset, let's run the additional checks
-    if (fileAndFileRule) {
-      // TODO: make the checks conditional based on the different kinds of conditions
-      // to be checked, i.e., runDependencyUpgradeValidation, etc.
-      if (!runVersioningValidation(fileAndFileRule)) return false;
+    // We're running each file through the check so that we can make customized checks
+    // for each file. This way, we can be as specific as possible as to how the file
+    // needs to be checked.
+    while (fileAndFileRule) {
```
tmatsuo, line 103: To be honest, I feel that reading this block of code little bit difficult to understand.

tmatsuo (commented):
Ok, I'm putting all the feedback at the moment.

I'll need to understand the code more to do the code review further.

packages/auto-approve/src/check-pr.ts
sofisl: Thanks, I corrected that!

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -48,19 +52,24 @@ export interface Versions {
...
   changedFiles: File[],
   author: string,
-  languageRules: FileSpecificRule[]
-): {file: File; fileRule: FileSpecificRule} | undefined {
+  languageRules: FileSpecificRule[],
+  searchAfter: number
+): {file: File; fileRule: FileSpecificRule; ithElement: number} | undefined {
   let file;
   let fileRule;
+  let ithElement = 0;
```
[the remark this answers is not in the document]
sofisl: Sure.

packages/auto-approve/src/check-pr.ts
```diff
@@ -65,12 +66,16 @@ export async function checkPRAgainstConfig(
...
 
     // Since there's only one allowed title per author right now, we don't need to
     // add complicated logic to see which title should match the incoming PR; but,
     // this could be logic we work into the future.
     if (!title.match(rulesToValidateAgainst.title)) {
+      logger.info(
+        `Info for ${repoOwner}/${repo}/${prNumber} title does not match what is allowed`
+      );
       return false;
     }
```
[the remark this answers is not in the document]
sofisl: Also corrected this.

packages/auto-approve/src/valid-pr-schema.json
sofisl: Marking for further discussion.

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -48,19 +52,24 @@ export interface Versions {
 export function getTargetFile(
   changedFiles: File[],
   author: string,
-  languageRules: FileSpecificRule[]
-): {file: File; fileRule: FileSpecificRule} | undefined {
+  languageRules: FileSpecificRule[],
+  searchAfter: number
+): {file: File; fileRule: FileSpecificRule; ithElement: number} | undefined {
```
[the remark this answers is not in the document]
sofisl: Marking for further discussion.

packages/auto-approve/src/check-pr.ts
```diff
@@ -65,12 +66,16 @@ export async function checkPRAgainstConfig(
     // changedFiles and maxFiles are not set in the JSON schema
     let filePathsMatch = true;
     let fileCountMatch = true;
-    const versionChecks = true;
```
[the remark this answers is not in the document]
sofisl: Perhaps we should leave this as future work? I just don't want to cause any unforseen issues if it breaks.

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -48,19 +52,24 @@ export interface Versions {
...
   let file;
   let fileRule;
+  let ithElement = 0;
+  let keepSearching = true;
```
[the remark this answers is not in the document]
sofisl: Thanks for this. I don't think I can use a shortened loop signature while initializing the variable (from searchAfter), but I was able to name the loop

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -103,27 +114,69 @@ export function getVersions(
...
+
+/**
+ * This function checks whether the dependency stated in a given title was the one that was changed
+ *
+ * @param versions the Versions object that contains the old dependency name and new dependency name and versions
+ * @param dependencyRegex the regular exp to find the dependency within the title of the PR
+ * @param title the title of the PR
+ * @returns whether the old dependency, new dependency, and dependency in the title all match
+ */
+export function doesDependencyMatchTarget(
```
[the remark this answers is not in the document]
sofisl: Sure.

packages/auto-approve/src/utils-for-pr-checking.ts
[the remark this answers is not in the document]
tmatsuo: Yeah, for the outer loop, you can keep using the index, but for the inner code, you don't need the index.

packages/auto-approve/src/check-pr.ts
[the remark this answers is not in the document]
tmatsuo: I don't think it breaks. I use `Octokit` in many many places in our libraries and bots.

But I'm open to defer.

packages/auto-approve/src/valid-pr-schema.json
sofisl: Per our conversation yesterday, here is an [issue](https://github.com/googleapis/repo-automation-bots/issues/2090).

packages/auto-approve/src/check-pr.ts
```diff
@@ -84,20 +89,76 @@ export async function checkPRAgainstConfig(
...
+        if (additionalRules === false) {
+          return false;
+        }
+      }
+    }
+
+    // This additional check confirms that there were no files changed in the PR
+    // that weren't checked by the additional files rule if it's from renovate-bot
+    if (prAuthor === 'renovate-bot') {
+      for (let i = 0; i < changedFiles.length; i++) {
```
tmatsuo, line 151: Also nit: I prefer `for (const thing of things) {`, but it might be just a taste.

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -19,14 +19,34 @@ import utc from 'dayjs/plugin/utc';
 import timezone from 'dayjs/plugin/timezone';
 dayjs.extend(utc);
 dayjs.extend(timezone);
+
```
tmatsuo, line 22: I will be glad if you add jsdoc for this interface too.

tmatsuo (commented):
I think the code is now much easier to read, thank you!

Adding initial comments.
I'll read the code more closely and add another review later today.

packages/auto-approve/src/check-pr.ts
```diff
@@ -84,20 +89,76 @@ export async function checkPRAgainstConfig(
...
 
     // If we've found the file in this ruleset, let's run the additional checks
-    if (fileAndFileRule) {
-      // TODO: make the checks conditional based on the different kinds of conditions
-      // to be checked, i.e., runDependencyUpgradeValidation, etc.
-      if (!runVersioningValidation(fileAndFileRule)) return false;
+    // We're running each file through the check so that we can make customized checks
+    // for each file. This way, we can be as specific as possible as to how the file
+    // needs to be checked.
+    if (fileAndFileRules.length !== 0) {
```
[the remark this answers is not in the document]
sofisl: Whoops, this was a remnant from the while loop. Thanks.

packages/auto-approve/src/check-pr.ts
sofisl: I think we need an iterator since I'm also iterating through the elements in another array at the same time.

packages/auto-approve/src/language-versioning-rules.ts
```diff
@@ -0,0 +1,48 @@
...
+  {
+    prAuthor: 'renovate-bot',
+    process: 'dependency',
+    targetFile: 'package.json',
+    dependency:
+      '^(fix\\(deps\\)|chore\\(deps\\)): update dependency (@?\\S*) to v(\\S*)$',
+    // This would match: -  "version": "^2.3.0"
+    oldVersion: '-[\\s]*"(@?\\S*)":[\\s]"([\\^0-9]*)*\\.([0-9]*\\.[0-9]*)",',
+    // This would match: +  "version": "^2.3.0"
+    newVersion: '\\+[\\s]*"(@?\\S*)":[\\s]"([\\^0-9]*)*\\.([0-9]*\\.[0-9]*)",',
```
tmatsuo, line 34: I suspect that most of the cases, this is fine.
However, what happens if the repo is using `~` instead of `^`, or something like `=>`?

I'm totally fine with starting with the current regex, but I'll be happy if we document this limitation somewhere.

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -45,34 +70,29 @@ export interface Versions {
  * @returns an object containing the specific file to be scrutinized and the rules to scrutinize it by,
```
tmatsuo, line 70: Can you update the return type? It still says `undefined`.

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -45,34 +70,29 @@ export interface Versions {
...
-        fileRule = languageRules[j];
-        break;
+        console.log({file: changedFile, fileRule: rule});
+        targetFiles.push({file: changedFile, fileRule: rule});
       }
     }
   }
 
   // If we didn't find a match, return undefined, but this shouldn't
   // stop execution of the rest of the tests
```
tmatsuo, line 93: The comment is now stale. Maybe just delete it?

packages/auto-approve/test/check-pr.test.ts
```diff
@@ -276,5 +277,223 @@ describe('check pr against config', async () => {
...
+        'events',
+        'pull_request_opened_special_validation_dependency_update'
+      ));
+
+      const scopes = listChangedFilesPR(200, [
+        {
+          filename: 'package.json',
+          sha: '1234',
+          patch:
+            '@@ -1,7 +1,7 @@\n' +
```
tmatsuo, line 300: non-blocking suggestions:
How about to use file based fixtures for these diffs?

packages/auto-approve/src/check-pr.ts
tmatsuo: Ah Ok.

Non blocking comment. I think there is an interesting case we will need to think about.

Renovate creates dependency update.
Owl bot postprosessor run `npm run fix` then add a commit.

tmatsuo (changes requested):
I have read the code closely.

I hope this is the last round trip. I added some comments including non-blocking and blocking comment. Please address some blocking (but it should be easy) comments.

Thanks!

packages/auto-approve/src/check-pr.ts
```diff
@@ -84,20 +89,74 @@ export async function checkPRAgainstConfig(
...
 
     // If we've found the file in this ruleset, let's run the additional checks
-    if (fileAndFileRule) {
-      // TODO: make the checks conditional based on the different kinds of conditions
-      // to be checked, i.e., runDependencyUpgradeValidation, etc.
-      if (!runVersioningValidation(fileAndFileRule)) return false;
+    // We're running each file through the check so that we can make customized checks
+    // for each file. This way, we can be as specific as possible as to how the file
+    // needs to be checked.
+    for (const fileAndFileRule of fileAndFileRules) {
```
bcoe, line 102: thank you for making this edit 😄

bcoe (approved):

packages/auto-approve/src/check-pr.ts
sofisl: Does the postprocessor run on forks? https://github.com/googleapis/repo-automation-bots/pull/2102

tmatsuo: Yeah I think so, you likely need to add `owlbot:run` label or something

packages/auto-approve/src/language-versioning-rules.ts
sofisl: I haven't seen this operator in Node: `=>`
But, I have written out logic for `~`, `^`

packages/auto-approve/src/utils-for-pr-checking.ts
sofisl: I'm not sure what you're referring to. This function returns either a `Versions` object, or `undefined`.

packages/auto-approve/test/check-pr.test.ts
sofisl: I think the patches are small enough to be in-line. I also feel it makes debugging a bit simpler. Perhaps if the patches get larger.

packages/auto-approve/src/utils-for-pr-checking.ts
tmatsuo: I'm talking about the jsdoc `@returns` on getTargetFiles.

Now it returns `fileAndMetadata[]` and it will return an empty array, right?
But the jsdoc still says `or undefined if not found.`

packages/auto-approve/test/check-pr.test.ts
tmatsuo: ok

packages/auto-approve/src/utils-for-pr-checking.ts
sofisl: I see, I have updated.

sofisl:
> comment update?

I've updated all the comments accordingly.

tmatsuo (approved):
Thanks for your patience. It looks good now!

PR #2031, opened 2021-06-11
fix(snippet-bot): do not add a useless comment

### LINKED ISSUE #2021: snippet-bot: pull request that has never touched region tags should not have comment added

tmatsuo:
fixes #2021

When the PR is not modifying any region tags, the bot will not add
a new comment, but if there's already a comment, it will update it.


packages/snippet-bot/test/snippet-bot.ts
```diff
@@ -603,15 +595,7 @@ describe('snippet-bot', () => {
...
-        .reply(200, [])
-        .post(
-          '/repos/tmatsuo/repo-automation-bots/issues/14/comments',
-          body => {
-            snapshot(body);
-            return true;
-          }
-        )
-        .reply(200);
+        .reply(200, []);
```
bcoe, line 598: Would it be possible to add a test for the two new cases:

"no comment added if no samples modified"

"existing comment updated, if comment existed and samples no longer updated".

bcoe (changes requested):

packages/snippet-bot/test/snippet-bot.ts
tmatsuo: Added 2 tests, PTAL

bcoe (approved):
ship it 👌

Issue #2040, opened 2021-06-12
Failed to update package-lock.json for some packages.


tmatsuo:
- [ ] gcf-utils: 'npm i' with ncu failed
- [ ] owl-bot: 'npm run test' with ncu failed
- [ ] owl-bot: 'npm run test'  failed
- [ ] policy: 'npm run test' with ncu failed

For auto-approve and owl-bot, test failed both with or without the major bump.

For gcf-utils, `npm i` (and compile) failed with major bump, but it successfully update the package-lock.json without the major bump.

For policy, `npm run test` failed with major bump, but it successfully update the package-lock.json without the major bump.

tmatsuo:
The failure log is too big, so created [a gist](https://gist.github.com/tmatsuo/751db2d4806394a6f5bd4800a7d625fb)

tmatsuo:
I run the script again today. auto-approve is ok now.

Here are some relevant logs:

gcf-utils-npm-i-with-ncu.log
```

> gcf-utils@10.0.0 prepare /usr/local/google/home/tmatsuo/work/repo-automation-bots/packages/gcf-utils
> npm run compile


> gcf-utils@10.0.0 compile /usr/local/google/home/tmatsuo/work/repo-automation-bots/packages/gcf-utils
> tsc -p .

node_modules/@types/pino/index.d.ts:197:8 - error TS2709: Cannot use namespace 'SonicBoom' as a type.

197     ): SonicBoom;
           ~~~~~~~~~

node_modules/@types/pino/index.d.ts:205:57 - error TS2709: Cannot use namespace 'SonicBoom' as a type.

205     function extreme(fileDescriptor?: string | number): SonicBoom;
                                                            ~~~~~~~~~

src/bin/genkey.ts:63:22 - error TS2339: Property 'keyfile' does not exist on type '{ [x: string]: unknown; keyfile: string; verbose: boolean
; project: string; bot: string; id: string; secret: string; _: (string | number)[]; $0: string; } | Promise<{ [x: string]: unknown; keyfile: string; verbose: boolean; project:
 string; ... 4 more ...; $0: string; }>'.
  Property 'keyfile' does not exist on type 'Promise<{ [x: string]: unknown; keyfile: string; verbose: boolean; project: string; bot: string; id: string; secret: string; _: (s
tring | number)[]; $0: string; }>'.

63 const keyfile = argv.keyfile || 'key.pem';
                        ~~~~~~~

[...]

PR #2041, opened 2021-06-12
build: add npm-check-updates to update-package.locks.sh

### LINKED ISSUE #2036: feat: we want a maintenance script for updating package.json
Background:
`npm update` seems [to stop updating package.json](https://github.com/npm/cli/issues/708).

We want to routinely update our package.json, even optionally bumping major versions too.

tmatsuo:
I changed the name of the script to `update-dependencies.sh`.
You need to install npm-check-updates in order to use this script now.

```
npm i -g npm-check-updates
```

The script will try to update everything to the latest version (including major bump) with `ncu -u`, then run
`npm i`, `npm audit fix`, `npm run fix`, then `npm run test`.

When the first trial fails, it falls back to do the same thing without
updating `package.json` file.

If you give arguments to this script, it only iterates the given directories.
Example:

```
$ scripts/update-dependencies.sh packages/snippet-bot packages/bot-config-utils
```

fixes #2036 

#2040 and #2042 are example results.

This solves some problems with the current renovate solutions.

- Gives flexibility of how many packages in one single PR
- This script only preserve changes when test passes, so likely no more test failures on package update PRs
- This script also help us to track the failed update attempts (even if it's manual)

Future possible enhancements (we don't need them, but they are handy):
- automatic creation of PRs (limiting the number of packages in a single PR)
- automatic reporting the failed attempts

### DIFF
--- a/scripts/update-dependencies.sh
+++ b/scripts/update-dependencies.sh
@@ -14,23 +14,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Update the package-lock.json files and run tests.
+# To use this script, you need to install npm-check-updates
+# npm i -g npm-check-updates
+
+# This script will first try to update package.json with major bumps
+# with npm-check-updates, then run the following:
+# - npm i
+# - npm run fix
+# - npm run test
 
 # When the test fails, it stores the test log and the summary in the
-# temp directory, and revert the changes in the sub directory.
+# temp directory, and revert the changes in the sub directory. The
+# script then try to do the same thing without the major bumps.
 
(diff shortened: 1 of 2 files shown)


sofisl:
@tmatsuo, I definitely can't give feedback on the bash syntax itself, but the steps make sense to me. 

If we really really wanted to make this process more efficient, I think it would be really great to build the following workflow:

1. A package-lock update comes in, and kicks off your script,
2. The script opens a PR with successful updates, and
3. Opens issues for unsuccessful updates, 
4. Comments on original package-lock updates with PRs and issues created.

sofisl (approved):

PR #2052, opened 2021-06-14
fix: skip task queue retries if payload expired

### LINKED ISSUE #2049: Limit number of task retries
Some jobs in our task queues are on their 90th+ retry. After failing that many times, it's highly unlikely that task will succeed. 

Related to #527

Perhaps something like retry N times, then open an issue.

chingor13:
Fixes #2049


packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -271,6 +271,62 @@ describe('GCFBootstrapper', () => {
...
+        .get('/storage/v1/b/tmp/foo/o/%2Fbucket%2Ffoo?alt=media')
+        .reply(404);
+
+      await handler(req, response);
+
+      delete process.env.WEBHOOK_TMP;
+      sinon.assert.calledOnce(configStub);
+      sinon.assert.notCalled(sendStatusStub);
+      sinon.assert.calledOnceWithMatch(sendStub, {statusCode: 200});
+      sinon.assert.notCalled(spy);
```
tmatsuo, line 298: Nit: naming suggestion: `handerStub`?
Not a blocker.

tmatsuo (approved):

packages/gcf-utils/test/gcf-bootstrapper.ts
chingor13: These were the existing stub names, we can rename them in a follow-up.

PR #2062, opened 2021-06-14
chore(generated-files-bot): fix warning from dependabot

tmatsuo:
Somehow npm i resolved to older versions.
I needed to add explicit dev dependencies.


tmatsuo:
weird, locally it is fixed without explicit dependency.

chingor13 (approved):

PR #2063, opened 2021-06-15
chore: remove package-lock.json for libraries

chingor13:
Deployed apps/functions should lock their dependencies. This ensures that we run in production what we're testing against.

Libraries should not use `package-lock.json` files as its dependencies are resolved by the depending program and may differ from what we're testing against. This may cause tests on these libraries to fail if a dependency update breaks our behavior, but we will catch it earlier in our tests. In these cases, we should pin the dependency version harder in the `package.json`


tmatsuo (commented):
It seems like cron-utils, mono-repo-publish, and release-brancher are standalone app.

Do you intend to remove package-lock.json for those standalone apps too?

tmatsuo:
> as its dependencies are resolved by the depending program and may differ from what we're testing against.

I think this problem is not solved by removing package-lock.json because package-lock.json is not published to npm.

I think having package-lock.json provides some benefit. It shows the combination of versions of dependencies which passed the test.

chingor13:
Ok, I agree now that we kinda use the package-lock.json renovate PR as our CI.

Issue #2068, opened 2021-06-16
bug(flakybot): scheduler request is handled by the bot directly


tmatsuo:
I created scheduler jobs for flakybot for syncing labels in #2065.

It looks like the scheduler job is directly executed by the bot.

I suspect this is because flakybot configured as `background`: false.
https://github.com/googleapis/repo-automation-bots/blob/21138aac7b187fc3876aa3a0e2ddfcee9710bf5c/packages/flakybot/src/app.ts#L20

The code goes through this line:
https://github.com/googleapis/repo-automation-bots/blob/21138aac7b187fc3876aa3a0e2ddfcee9710bf5c/packages/gcf-utils/src/gcf-utils.ts#L335

PR #2069, opened 2021-06-16
fix(gcf-utils): correctly handle scheduler task for non-background apps

tmatsuo:
part of #2068


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -294,6 +294,12 @@ export class GCFBootstrapper {
         if (triggerType === TriggerType.UNKNOWN) {
           response.sendStatus(400);
           return;
+        } else if (triggerType === TriggerType.SCHEDULER) {
+          // TODO: currently we assume that scheduled events walk all repos
+          // managed by the client libraries team, it would be good to get more
+          // clever and instead pull up a list of repos we're installed on by
+          // installation ID:
```
chingor13, line 301: This comment is not correct, but it's moved from below. We do pull the list of repositories by installation id.

chingor13 (approved):

PR #2072, opened 2021-06-16
chore(owl-bot): update dependencies

tmatsuo:
I just started to use node 14 and the test passes.


tmatsuo:
The only concern is that, locally the test fails with node12. The test passes with node14.
On GCF, we deploy owl-bot with node12 runtime...

However, considering that the test fails even with the current HEAD with node12, so it might not be a problem.

PR #2091, opened 2021-06-17
fix(owl-bot): set credentials earlier in the build

### LINKED ISSUE #2088: owl bot checks fail on private repos
The owl bot checks currently fail on private repos. As an example, see https://github.com/googleapis/nodejs-life-sciences/pull/1 , which has a [failing check](https://github.com/googleapis/nodejs-life-sciences/pull/1/checks?check_run_id=2845775931).

The error from the build log for the failing check in `nodejs-life-sciences` is as follows:
```
2021-06-17T05:11:32.137487374ZStarting Step #0
Info
2021-06-17T05:11:32.138898561ZStep #0: Already have image (with digest): gcr.io/cloud-builders/git
Info
2021-06-17T05:11:32.639123089ZStep #0: Cloning into 'nodejs-life-sciences'...
Info
2021-06-17T05:11:34.377585974ZStep #0: fatal: could not read Username for 'https://github.com': No such device or address
Info
2021-06-17T05:11:34.584731600ZFinished Step #0

```

You can also find other repos in this list:
https://github.com/googleapis?q=is%3Aprivate&type=&language=&sort=

tmatsuo:
fixes #2088

### DIFF
--- a/packages/owl-bot/cloud-build/update-pr.yaml
+++ b/packages/owl-bot/cloud-build/update-pr.yaml
@@ -13,6 +13,16 @@
 # limitations under the License.
 
 steps:
+  # Populate credentials for cloning from private repo
+  - name: 'bash'
+    args:
+      - '-eEuo'
+      - 'pipefail'
+      - '-c'
+      - |-
+        echo https://x-access-token:$$GITHUB_TOKEN@github.com >> $${HOME}/.git-credentials
+    env:
+      - 'GITHUB_TOKEN=$_GITHUB_TOKEN'
   # Clone user/branch that created PR:
   - name: 'gcr.io/cloud-builders/git'
     args:
--- a/packages/owl-bot/cloud-build/update-pr.yaml
+++ b/packages/owl-bot/cloud-build/update-pr.yaml
@@ -13,23 +13,10 @@
 # limitations under the License.
 
 steps:
-  # Populate credentials for cloning from private repo
-  - name: 'bash'
-    args:
-      - '-eEuo'
-      - 'pipefail'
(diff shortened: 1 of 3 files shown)


packages/owl-bot/cloud-build/update-pr.yaml
```diff
@@ -13,6 +13,16 @@
...
+  # Populate credentials for cloning from private repo
+  - name: 'bash'
+    args:
+      - '-eEuo'
+      - 'pipefail'
+      - '-c'
+      - |-
+        echo https://x-access-token:$$GITHUB_TOKEN@github.com >> $${HOME}/.git-credentials
+    env:
+      - 'GITHUB_TOKEN=$_GITHUB_TOKEN'
```
SurferJeffAtGoogle, line 25: Can we now remove lines 51 through 60?  Have you tried running this build manually?

tmatsuo: Actually no. Is there a guide for running it locally?

Keeping the git credentials in the git clone is a safe guard for being non disruptive (it will continue to work even if the first step is doing it wrong).

I'm happy to remove if it's not necessary.

SurferJeffAtGoogle (approved):

tmatsuo:
@SurferJeffAtGoogle 

I think I figured out how to set the credentials earlier. Initial trial failed because the ${HOME} is not persisted among cloud build steps. So I think we need to set HOME=/workspace for preserving all the git configs.

I managed to run this locally with my personal repo and it went through the git clone phase, but I haven't tested it with a real private repo.

PTAL

PR #2093, opened 2021-06-17
fix!: verify payload signatures

chingor13:
* For all requests, verify the signature before taking any actions.
* When queuing a task, re-sign the request body (which is not necessarily the webhook payload)

Not really a breaking change, but forcing a major version bump.

### DIFF
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -57,7 +57,6 @@ interface Scheduled {
 
 interface EnqueueTaskParams {
   body: string;
-  signature: string;
   id: string;
   name: string;
 }
@@ -65,6 +64,7 @@ interface EnqueueTaskParams {
 export interface WrapOptions {
   background: boolean;
   logging: boolean;
+  skipVerification?: boolean;
 }
 
 export const logger = new GCFLogger();
@@ -286,8 +286,14 @@ export class GCFBootstrapper {
     return async (request: express.Request, response: express.Response) => {
       wrapOptions = wrapOptions ?? {background: true, logging: false};
 
+      // default skip verification for testing
+      // to test signatures, explicitly set skipVerification to false
+      if (wrapOptions.skipVerification === undefined) {
+        wrapOptions.skipVerification = RUNNING_IN_TEST;
+      }
+
       this.probot =
-        this.probot || (await this.loadProbot(appFn, wrapOptions?.logging));
+        this.probot || (await this.loadProbot(appFn, wrapOptions.logging));
 
       const {name, id, signature, taskId} =
         GCFBootstrapper.parseRequestHeaders(request);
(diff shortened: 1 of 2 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -57,14 +57,14 @@ interface Scheduled {
...
   body: string;
-  signature: string;
   id: string;
   name: string;
 }
 
 export interface WrapOptions {
   background: boolean;
   logging: boolean;
+  skipVerification?: boolean;
```
SurferJeffAtGoogle, line 67: Please add a comment for `skipVerification`.  Comments for `background` and `logging` would also be appreciated.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -232,6 +232,18 @@ export class GCFBootstrapper {
     }
   }
 
+  private static parseSignatureHeader(request: express.Request): string {
```
SurferJeffAtGoogle, line 235: In a comment for this function, explain why it returns 'unset'.  I would have expected it to return the empty string when a signature was not found in the request.

SurferJeffAtGoogle (changes requested):

packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -640,6 +640,115 @@ describe('GCFBootstrapper', () => {
...
+        });
+        req.body = {
+          installation: {id: 1},
+        };
+        req.headers = {};
+        req.headers['x-github-event'] = 'issues';
+        req.headers['x-github-delivery'] = '123';
+        req.headers['x-cloudtasks-taskname'] = 'my-task';
+        req.headers['x-hub-signature'] =
+          'sha1=c012c260559a04cf285e05d67e1ecedcad71b931';
```
tmatsuo, line 710: I'll be glad if there's a comment describing how to calculate this value.

tmatsuo (approved):
Looks good
Please address comments from @SurferJeffAtGoogle tho

packages/gcf-utils/src/gcf-utils.ts
chingor13: Added comments for each

packages/gcf-utils/src/gcf-utils.ts
chingor13: Done

packages/gcf-utils/test/gcf-bootstrapper.ts
chingor13: Added a comment: `echo -n '{"installation":{"id":1}}' | openssl dgst -sha1 -hmac "foo"`

tmatsuo (approved):

SurferJeffAtGoogle (approved):

PR #2101, opened 2021-06-18
fix: fix serverless proxy signing

### LINKED ISSUE #2092: scheduler-proxy: signatures should be prepended with algorithm
https://docs.github.com/en/developers/webhooks-and-events/webhooks/securing-your-webhooks

The `x-hub-signature` header should be prepended with `sha1=`, example: `sha1=03c67a1465cf31173dde015eb20a50abb0dcd7f1`

chingor13:
Fixes 2 issues with the signing:
1. We are not prepending the `sha1=`
2. We are not using the correct secret value. The secret from Secret Manager is actually a JSON blob and the secret is one value from it that we need to parse out. (previously, we were using the entire secrets blob as the secret)
3. The signatures should be hex encoded, not base64

Fixes #2092

### DIFF
--- a/serverless-scheduler-proxy/main.go
+++ b/serverless-scheduler-proxy/main.go
@@ -182,7 +182,7 @@ func botContainerProxy(cfg botConfig) http.HandlerFunc {
 }
 
 type SecretConfig struct {
-	PrivateKey string `json:"privateKey`
+	PrivateKey string `json:"privateKey"`
 	AppId      int    `json:"appId"`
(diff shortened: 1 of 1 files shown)


chingor13:
I tested this locally against a built version of the proxy that redirected to localhost which was running the canary bot with verification turned on.

lesv (approved):

tmatsuo (approved):

orthros (approved):

PR #2106, opened 2021-06-18
fix(owlbot): add owbot:run for trusted contributors

### LINKED ISSUE #1714: OwlBot checks should be made to be required checks
I think the problem is that we're not calling the github API to announce that the check was kicked off.  We only call the github API when the check has completed.  There's a window of time when the user isn't aware that the Owl Bot check is running, and may merge.

For example, this PR was merged before Owl Bot post processor completed.
https://github.com/googleapis/python-security-private-ca/pull/35

CC @parthea

bcoe:
Along with running `kokoro:run` for trusted contributors, we need to make sure to run `owlbot:run`.

Refs #1714

### DIFF
--- a/packages/owl-bot/src/core.ts
+++ b/packages/owl-bot/src/core.ts
@@ -474,6 +474,32 @@ async function hasOwlBotLoop(
   return false;
 }
 
+/*
+ * Return whether or not the last commit was from OwlBot.
+ *
+ * @param owner owner of repo.
+ * @param repo short repo name.
+ * @param prNumber PR to check for commit.
+ * @param octokit authenticated instance of octokit.
+ * @returns Promise was the last commit from OwlBot?
+ */
+async function lastCommitFromOwlBot(
+  owner: string,
+  repo: string,
+  prNumber: number,
+  octokit: Octokit
+): Promise<boolean> {
+  const commits = (
+    await octokit.pulls.listCommits({
+      pull_number: prNumber,
+      owner,
+      repo,
+    })
+  ).data;
+  const commit = commits[commits.length - 1];
+  return commit?.author?.login === OWLBOT_USER;
+}
+
 /**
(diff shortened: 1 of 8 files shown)


tmatsuo (changes requested):
Please update the src/config-shcema.json as well.

tmatsuo (approved):

tmatsuo:
The code itself looks good, but maybe we should use another type of label? `owlbot:run-from-fork`, or `owlbot:approve`?

tmatsuo:
Per conversation on the chat.

Maybe we should add annotations only for PRs from a fork.

tmatsuo (changes requested):
Can you also make the bot only add the annotations for PRs from a fork?

tmatsuo (approved):
Ok, I was wrong. We still needs annotations for PRs from the same repo.

packages/owl-bot/src/core.ts
```diff
@@ -474,6 +474,32 @@ async function hasOwlBotLoop(
...
+ * @returns Promise was the last commit from OwlBot?
+ */
+async function lastCommitFromOwlBot(
+  owner: string,
+  repo: string,
+  prNumber: number,
+  octokit: Octokit
+): Promise<boolean> {
+  const commits = (
+    await octokit.pulls.listCommits({
```
tmatsuo, line 493: Probably it's alright, but there's an edge case where there are more than 100 commits.

tmatsuo (approved):

packages/owl-bot/src/core.ts
bcoe: I don't think this API allows for more than 100 commits, maybe just add a comment.

bcoe: I'm wrong, it can potentially support pagination, I will add note.

tmatsuo (approved):

PR #2108, opened 2021-06-18
fix: only provide the installationId if set

### LINKED ISSUE #2107: gcf-utils: error creating installation-less octokit instance
```
message: "[@octokit/auth-app] installationId is set to a falsy value"   
  stack: "Error: [@octokit/auth-app] installationId is set to a falsy value
```

chingor13:
`@octokit/auth-app` distinguishes between unset and falsy. We need to make sure we do not set installation id if we want an installationless octokit instance.

Fixes #2107


tmatsuo (approved):

tmatsuo:
You can add system test that ensures installationid less octokit calling actual method, but I'm fine with this.

PR #2116, opened 2021-06-18
feat: add max retries config options

chingor13:
* Introduces 3 new configuration parameters:
  * `maxRetries` - max retries for GitHub webhook events. defaults to 10
  * `maxCronRetries` - max retries for cron events. defaults to 0
  * `maxPubSubRetries` - max retries for pubsub events. defaults to 0
* All handlers now go through Cloud Tasks - no requests are handled directly by the webhook source (Pub/Sub, Cron, or GitHub webhook).
* Deprecates the `background` configuration option. It now sets retries to 0 for all payload types. This option was originally there when the payload size was too large for Cloud Tasks, but that was resolved by writing the payload to the GCS.

### DIFF
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -64,16 +64,42 @@ interface EnqueueTaskParams {
 export interface WrapOptions {
   // Whether or not to enqueue direct GitHub webhooks in a Cloud Task
   // queue which provides a retry mechanism. Defaults to `true`.
-  background: boolean;
+  // Deprecated. Please use `maxRetries` and `maxCronRetries` instead.
+  background?: boolean;
 
   // Whether or not to automatically log Octokit requests. Defaults to
   // `false`.
-  logging: boolean;
+  logging?: boolean;
 
   // Whether or not to skip verification of request payloads. Defaults
   // to `true` in test mode, otherwise `false`.
   skipVerification?: boolean;
+
+  // Maximum number of attempts for webhook handlers. Defaults to `10`.
+  maxRetries?: number;
+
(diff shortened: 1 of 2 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -64,17 +64,43 @@ interface EnqueueTaskParams {
...
 
   // Whether or not to skip verification of request payloads. Defaults
   // to `true` in test mode, otherwise `false`.
   skipVerification?: boolean;
+
+  // Maximum number of attempts for webhook handlers. Defaults to `10`.
+  maxRetries?: number;
+
+  // Maximum number of attempts for cron handlers. Defaults to `0`.
+  maxCronRetries?: number;
```
tmatsuo, line 82: Little bit worried about the default value of 0.
Do we need to set it properly when migrating? If so, what's the equivalent to the current behavior?

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -64,17 +64,43 @@ interface EnqueueTaskParams {
...
+
+  // Maximum number of attempts for cron handlers. Defaults to `0`.
```
tmatsuo, line 84: The comment has copy paste error

tmatsuo (approved):

packages/gcf-utils/src/gcf-utils.ts
chingor13: Currently, if `background: true`, the cron request is handled directly (so no retries). Otherwise, it's unlimited retries (until the payload expires after a day).

tmatsuo: Ok

PR #2119, opened 2021-06-18
refactor: use new cron types for merge-on-green

chingor13:
* Updates merge-on-green to gcf-utils v11.
* Switches to use our new cron types.
  * The merger will run per-installation and fetch watched PRs by installation id
  * The find hanging PRs job runs per installation and only on org installations
  * The cleanup job runs globally
  * The sync labels cron run per-repository
* Consolidates cron job's to allow gcf-utils to fanout installations (so we don't need to hardcode installation ids)

*Note: before deploying, delete all the m-o-g scheduled tasks and let them rebuild*


tmatsuo (approved):

chingor13:
Would like @sofisl to review before merging

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -511,49 +512,81 @@ handler.scanForMissingPullRequests = async function scanForMissingPullRequests(
...
-  //open issue for a fix: https://github.com/octokit/webhooks.js/issues/277
-  app.on('schedule.repository' as '*', async context => {
-    if (context.payload.syncLabels === true) {
-      const owner = context.payload.organization.login;
-      const repo = context.payload.repository.name;
-      await syncLabels(context.octokit, owner, repo, MERGE_ON_GREEN_LABELS);
+  // This scheduled job iterates through the PR database and removes PRs
+  // That are closed or do not have an applicable label anymore.
+  app.on('schedule.global' as '*', async context => {
+    if (context.payload.cleanUp !== true) {
```
sofisl, line 518: So, switching the logic this way will make it so that the regular cron jobs (not the clean-up ones) will not execute. This isn't equivalent to 
`if (context.payload.cleanUp === true)`

chingor13: It's a guard clause in case we have multiple `global` cron events in the future. The other non-global crons won't attempt to run this block.

chingor13: The regular merge ones are now on `cron_type: installation` which runs in a `schedule.installation` handler.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -511,49 +512,81 @@ handler.scanForMissingPullRequests = async function scanForMissingPullRequests(
...
-    if (context.payload.cleanUp === true) {
-      logger.info('Entering clean up cron job');
-      await handler.cleanDatastoreTable(watchedPRs, app, context);
+    await handler.cleanDatastoreTable(watchedPRs, app, context);
+  });
+
+  // This scheduled job looks for PRs that have an applicable label
+  // but are not in the database for whatever reason (missed webhook).
+  app.on('schedule.installation' as '*', async context => {
+    if (context.payload.findHangingPRs !== true) {
```
sofisl, line 530: Same thing here regarding the cron jobs.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -511,49 +512,81 @@ handler.scanForMissingPullRequests = async function scanForMissingPullRequests(
...
+      context.octokit,
+      context.payload.cron_org
+    );
+    return;
+  });
+
+  // This scheduled job is the main recurring job that attempts to merge
+  // mergeable PRs.
+  app.on('schedule.installation' as '*', async context => {
+    if (context.payload.performMerge !== true) {
```
sofisl, line 553: Is this a new variable? I'm not sure what `performMerge` does.

sofisl:
@chingor13, thanks for the explanation and the work!

sofisl (approved):

PR #2135, opened 2021-06-21
fix: don't invoke the post processor for pull requests with the

SurferJeffAtGoogle:
owl-bot-ignore label


tmatsuo:
@SurferJeffAtGoogle Do you have a link to an issue?

packages/owl-bot/test/owl-core.ts
```diff
@@ -113,10 +126,27 @@ describe('core', () => {
         trigger: 'abc123',
       });
       assert.ok(triggerRequest);
-      assert.strictEqual(build.conclusion, 'success');
-      assert.strictEqual(build.summary, 'successfully ran 1 steps 🎉!');
+      assert.strictEqual(build!.conclusion, 'success');
+      assert.strictEqual(build!.summary, 'successfully ran 1 steps 🎉!');
+    });
+
+    it("doesn't trigger build when labled with owl-bot-ignore", async () => {
```
tmatsuo, line 133: Can you add a test that ensures that owl-bot adds success status?

tmatsuo (changes requested):

tmatsuo:
A suggestion to the label name:

Can you use `owl-bot:ignore`?
We already has `owl-bot:run`, so it's more consistent that way.

Also please consider syncing its labels with label-utils as other bots started doing so.

bcoe:
This is a great idea, I hadn't thought of this solution 😄

SurferJeffAtGoogle:
> @SurferJeffAtGoogle Do you have a link to an issue?

There's no issue open for this.

SurferJeffAtGoogle:
> A suggestion to the label name:
> 
> Can you use `owl-bot:ignore` instead of `owl-bot-ignore`?
> We already has `owl-bot:run`, so it's more consistent that way.
> 

I'll support both for now, because owl-bot-ignore is in use some places.  We can drop support for owl-bot-ignore later.

> Also please consider syncing its labels with label-utils as other bots started doing so.

Are there instructions for how to do this?

tmatsuo:
> I'll support both for now, because owl-bot-ignore is in use some places. We can drop support for owl-bot-ignore later.

SG

> Are there instructions for how to do this?

Currently:
https://github.com/googleapis/repo-automation-bots/blob/master/packages/label-utils/README.md

With the new cron type, we should update the guide too.

I'm happy to create a PR this time.

SurferJeffAtGoogle:
> > I'll support both for now, because owl-bot-ignore is in use some places. We can drop support for owl-bot-ignore later.
> 
> SG
> 
> > Are there instructions for how to do this?
> 
> Currently:
> https://github.com/googleapis/repo-automation-bots/blob/master/packages/label-utils/README.md
> 
> With the new cron type, we should update the guide too.
> 
> I'm happy to create a PR this time.

I like syncing labels, but I think it's outside the scope of this pull request.  Please open an issue describing what we need to do.

tmatsuo:
> I like syncing labels, but I think it's outside the scope of this pull request. Please open an issue describing what we need to do.

Yeah sounds good.

FYI there's already an issue:
https://github.com/googleapis/repo-automation-bots/issues/1673

packages/owl-bot/test/owl-bot.ts
```diff
@@ -379,6 +379,64 @@ describe('owlBot', () => {
...
+      .stub(core, 'hasOwlBotLoop')
+      .resolves(false);
+    const createCheckStub = sandbox.stub(core, 'createCheck');
+    await probot.receive({
+      name: 'pull_request.synchronize',
+      payload,
+      id: 'abc123',
+    });
+    sandbox.assert.calledOnce(triggerBuildStub);
+    sandbox.assert.calledOnce(createCheckStub);
```
tmatsuo, line 436: Can you also check the conclusion is `success`?

tmatsuo (changes requested):

packages/owl-bot/test/owl-bot.ts
SurferJeffAtGoogle: Done.

tmatsuo (approved):

PR #2137, opened 2021-06-21
chore: use the new cron type for snippet-bot cron

tmatsuo:


chingor13 (approved):
Make sure to delete the unnecessary scheduler jobs.

PR #2141, opened 2021-06-21
feat(owl-bot): start syncing labels

tmatsuo:
part of #1673


packages/owl-bot/cron.yaml
```diff
@@ -0,0 +1,7 @@
+cron:
+  - name: owl-bot-sync-label
+    schedule: 10 3 * * *
+    description: owl-bot syncing labels
+    params:
+      cron_type: "repository"
+      syncLabels: true
```
SurferJeffAtGoogle, line 7: Why isn't there an existing entry for `scanGithubForConfigs()`?

tmatsuo: My guess is `cron.yaml` is relatively a new thing. The scheduler job `owl-bot-scan-googleapis` might have created before `cron.yaml` is a thing.

I can create the entry if you like.

tmatsuo: Added an entry

packages/owl-bot/src/owl-bot.ts
```diff
@@ -150,6 +150,13 @@ export function OwlBot(
   // Run periodically, to ensure that up-to-date configuration is stored
   // for repositories:
   app.on('schedule.repository' as '*', async context => {
+    if (context.payload.syncLabels === true) {
+      // Sync labels.
+      const owner = context.payload.organization.login;
+      const repo = context.payload.repository.name;
+      await syncLabels(context.octokit, owner, repo, OWL_BOT_LABELS);
+      return;
```
SurferJeffAtGoogle, line 158: I don't understand all the contexts in which this function invoked.  I surprised to see a return statement here.  I would expect to see something like `if (context.payload.scanGithubForConfigs) {` before line 160.

Paging @bcoe who set up the original cron job.

tmatsuo: Added a parameter `scanGithubForConfigs`, also added some comment for clarification

tmatsuo: @SurferJeffAtGoogle Thanks, PTAL, and please let me know if the code is still unclear.

bcoe: I think what might make it clearer is if we made a `schedule.sync-labels` which would be run if the payload contains the `payload.syncLabels` standard field.

This could be follow up work, an we could track with issue?

---

With regards to having a `return` early, this is okay, there's no response sent during handler invocation, so an early return will just complete the promise.

SurferJeffAtGoogle (approved):

PR #2150, opened 2021-06-22
feat: allow configuration of GCFBootstrapper via constructor arguments

### LINKED ISSUE #2139: gcf-bootstrapper should accept configuration at instantiation
Currently, a bunch of functionality is set via environment variables (e.g. whether/where to write payloads to GCS, GCP Project Ids, etc.) These should be configurable via constructor arguments. This will allow for much easier testing. We can also fail early if we are missing configuration variables.

We can maintain behavior by having the server entrypoint default these configurations to the environment variables.

chingor13:
* Changes the constructor argument from positional args to an options object (technically breaking, but the only usage is in our tests).
* Default options to reading from the environment variables
* Throws errors early if required options are not set (rather than crashing later or having unexpected behavior)
* Refactors the tests to only set env variables for the duration of a single test.

Fixes #2139

### DIFF
--- a/packages/gcf-utils/test/gcf-bootstrapper.ts
+++ b/packages/gcf-utils/test/gcf-bootstrapper.ts
@@ -59,6 +59,67 @@ describe('GCFBootstrapper', () => {
     process.env = {...storedEnv};
   });
 
+  describe('configuration', () => {
+    it('requires a projectId to be set', () => {
+      assert.throws(
+        () => {
+          new GCFBootstrapper({
+            functionName: 'my-function',
+            location: 'my-location',
+          });
+        },
+        {
+          message: /Missing required `projectId`/,
+        }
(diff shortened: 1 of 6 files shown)


packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -58,8 +52,72 @@ function nockListInstallations(fixture = 'app_installations.json') {
 const sandbox = sinon.createSandbox();
 
 describe('GCFBootstrapper', () => {
+  // Save original env varas and restore after each test
+  const storedEnv = {...process.env};
```
tmatsuo, line 56: How about to use:
https://www.npmjs.com/package/mocked-env
?

tmatsuo (approved):

PR #2159, opened 2021-06-23
feat: allow queuing Cloud Tasks to hit Cloud Run service

### LINKED ISSUE #1817: infra: bot template/config for running on Cloud Run
In the future, we will want extra dependencies available to our bot. We should still be able to run our probot-based bots easily with our credentials and tasks shims in `gcf-utils` within a Docker container on Cloud Run.

- [x] Figure out how to run in Cloud Build
- [x] Configure canary-bot to run on Cloud Build
- [x] Configure scheduler-proxy to target a Cloud Build service
- [x] Configure gcf-utils to queue tasks to target the Cloud Build service
- [x] Configure generate-bot to templatize deployment for a new bot

chingor13:
* Uses the Cloud Run apiary client to get the url for the named service
* Provides a new `taskTargetEnvironment` configuration option for `GCFBootstrapper`.

Fixes #1817


tmatsuo (approved):

tmatsuo (changes requested):
I think we should be able to distinguish the app running on Cloud Function and the app running on Cloud Run, especially with the scheduler requests.

I'll add more thoughts tomorrow.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -885,12 +892,59 @@ export class GCFBootstrapper {
...
+    projectId: string,
+    location: string,
+    botName: string
+  ): Promise<string | null> {
+    // Cloud Run service names can only use dashes
+    const serviceName = botName.replace('_', '-');
+    const auth = new GoogleAuth({
+      scopes: ['https://www.googleapis.com/auth/cloud-platform'],
+    });
+    const authClient = await auth.getClient();
```
JustinBeckwith, line 905: You can actually just drop this line, and directly pass `auth` instead of fetching a client.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -885,12 +892,59 @@ export class GCFBootstrapper {
...
+    const client = await run({
+      version: 'v1',
+      auth: authClient,
+    });
+    const name = `projects/${projectId}/locations/${location}/services/${serviceName}`;
+    const res = await client.projects.locations.services.get({
+      name,
+    });
+
+    if (res.data.status?.address?.url) {
```
JustinBeckwith, line 915: If undefined is ok, you could return `res.data.stats?.address?.url` here and get the same effect.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -885,12 +892,59 @@ export class GCFBootstrapper {
     };
   }
 
+  private async getCloudRunUrl(
```
JustinBeckwith, line 895: Adding some doc annotations here would be nice :)

packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -1217,4 +1217,168 @@ describe('GCFBootstrapper', () => {
...
+      const createTask = sandbox
+        .stub(bootstrapper.cloudTasksClient, 'createTask')
+        .resolves();
+      await bootstrapper.enqueueTask({
+        body: JSON.stringify({installation: {id: 1}}),
+        id: 'some-request-id',
+        name: 'event.name',
+      });
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore - https://github.com/DefinitelyTyped/DefinitelyTyped/issues/36436
```
JustinBeckwith, line 1237: Instead of using `@ts-ignore` can we use a cast to `any`?  `@ts-ignore` comes with a bunch of other side effects I think we want to avoid in this case.

JustinBeckwith: Here and below

JustinBeckwith (approved):
LGTM w/ nits.  This is awesome.

tmatsuo (changes requested):
I think, for some period, we want to deploy the same bot both to Cloud Function and Cloud Run at the same time.

Can we auto detect the taskTargetEnvironment so that gcf-utils can automatically choose the right target URL?

tmatsuo:
Also I think we should address #2183 too.

chingor13:
> I think, for some period, we want to deploy the same bot both to Cloud Function and Cloud Run at the same time.

I had envisioned setting this target environment in the constructor which is in a different entrypoint script (example: https://github.com/googleapis/repo-automation-bots/pull/2142/files#diff-88bee7c686a71c57b64a40dc2ad31e8ae18510a3ce554dd2f16f30458e7640deR18)

> Can we auto detect the taskTargetEnvironment so that gcf-utils can automatically choose the right target URL?

It looks like Cloud Run only [provides a few environment variables](https://cloud.google.com/run/docs/reference/container-contract#env-vars) and they look to be K-Native standard variables so it wouldn't distinguish between any other K-Native environment and Cloud Run. We could also interrogate the metadata server. Alternatively, we could look for GCF's special environment variables (e.g. `FUNCTION_TARGET`).

packages/gcf-utils/src/gcf-utils.ts
chingor13: The response field is `string | null` so it'd be weird to return `string | null | undefined`. I'd prefer to limit to one missing type but maybe return null and undefined is preferred?

packages/gcf-utils/src/gcf-utils.ts
chingor13: I removed, then put it back because it's really hard to mock the auth here for environments that have ADC set up (like locally) and not (like GitHub actions CI).

chingor13:
@tmatsuo I think this behavior is fine for now (default to `function` which is the current behavior with an optional explicit override). Opened #2194 as a feature request to improve the default setting.

tmatsuo (approved):

PR #2161, opened 2021-06-23
fix(conventional-commit-lint): allow long footer line lengths

parthea:
This PR overrides the [default footer maximum line length](https://github.com/conventional-changelog/commitlint/blob/master/%40commitlint/config-conventional/index.js#L7) rule to allow a longer footer in the commit message. This PR also removes the [rule that prevents a full stop in the subject](https://github.com/conventional-changelog/commitlint/blob/master/%40commitlint/config-conventional/index.js#L15).

fix(conventional-commit-lint): allow full stop in subject


bcoe (approved):

bcoe:
@parthea thanks for contributing this 👍 sorry for the slow review.

Issue #2098, opened 2021-06-17
owlbot: _DEFAULT_BRANCH not set to default GitHub branch when new container submitted


bcoe:
When the `update-pr.yaml` job runs, it populates a variable `_DEFAULT_BRANCH` which is used by templates to set an appropriate default branch for templates.

This does not appear to happen when a new `OwlBot.lock` update is created:

https://github.com/googleapis/nodejs-local-auth/pull/94

Which results in the templates reverting the branch name back to the default `master`.

---

_Note: I noticed that if I manually add `owlbot:run` it fixes the issue, not 100% sure this didn't happen initially on #94._

PR #2170, opened 2021-06-24
feat: pass default branch name to post processor

SurferJeffAtGoogle:
during an update-lock build via DEFAULT_BRANCH_PATH environment variable

Fixes https://github.com/googleapis/repo-automation-bots/issues/2098


tmatsuo:
Question:

Is it possible to detect the default branch name in the owl-bot bot code and set it as a substitution?

SurferJeffAtGoogle:
> Question:
> 
> Is it possible to detect the default branch name in the owl-bot bot code and set it as a substitution?

Yes, it's possible, but that would require another call to the github API.  That's a network RPC which has a small, but non-zero chance of failing.  With this solution, there's no additional network RPC, so it reduces the chance of failure.

tmatsuo (approved):

PR #2172, opened 2021-06-24
build(owlbot): turn off signature verification until we can debug payload

bcoe:
The signature validation is failing for pubsub messages.

### DIFF
--- a/packages/gcf-utils/package.json
+++ b/packages/gcf-utils/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gcf-utils",
-  "version": "11.3.0",
+  "version": "11.3.0-debug.4",
   "description": "An extension for running Probot in Google Cloud Functions",
   "scripts": {
     "compile": "tsc -p .",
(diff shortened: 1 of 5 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -448,27 +449,34 @@ export class GCFBootstrapper {
         !wrapConfig.skipVerification &&
         !this.probot.webhooks.verify(request.body, signature)
       ) {
-        response.send({
+        logger.info('we send a statusCode 400');
+        /*response.send({
```
bcoe, line 453: signature validation appears to hiccup on Pubsub messages.

packages/gcf-utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "gcf-utils",
-  "version": "11.3.0",
+  "version": "11.3.0-debug.2",
```
bcoe, line 3: @SurferJeffAtGoogle the fact that I can do this is very useful, but something we should think about when we discuss BCID, I published a debugging version of client library to npm, which you have to opt into.

bcoe:
Turning off this setting is confirmed to work, I'm digging more into the underlying issue.

bcoe:
I would like to push out one more debug build of this, I'm very curious about what the signature is, and I'm going to print it out for testing.

SurferJeffAtGoogle (approved):

PR #2176, opened 2021-06-25
fix(gcf-utils): use raw body for signature verification

tmatsuo:
The `req.body` is a parsed json object. We should use the raw body
for signature verification.


bcoe (approved):
Looks great to me, I'm hopeful that this will fix pubsub.

A couple follow on issues we might want to open:

1. adding a pubsub check to canary bot, so that it updates an issue upon an event to a pubsub queue.
2. adding the same rawBody behavior to our Cloud Run server.ts version of gcf-utils.

tmatsuo:
Created followup issues:
#2182 
#2183

PR #2195, opened 2021-06-28
fix: express server should save rawBody

### LINKED ISSUE #2183: infrastructure: add rawBody on Cloud Run too
With #2176 gcf-utils started to use rawBody for signature verification on Cloud Function.
We should support the same thing on Cloud Run too.

chingor13:
Fixes #2183

### DIFF
--- a/packages/gcf-utils/test/server.ts
+++ b/packages/gcf-utils/test/server.ts
@@ -188,7 +188,10 @@ describe('getServer', () => {
       response: express.Response
     ) => {
       if (request.rawBody) {
-        response.sendStatus(200);
+        response.send({
+          status: 200,
+          body: request.rawBody.toString('utf-8'),
+        });
       } else {
         response.sendStatus(400);
       }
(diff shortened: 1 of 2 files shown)


packages/gcf-utils/test/server.ts
```diff
@@ -168,3 +175,50 @@ describe('GCFBootstrapper', () => {
...
+    const response = await gaxios.request({
+      url: `http://localhost:${TEST_SERVER_PORT}/`,
+      headers: {
+        'x-github-delivery': '123',
+        'x-cloudtasks-taskname': 'test-bot',
+        'x-github-event': 'issues',
+        'content-type': 'application/json',
+      },
+      method: 'POST',
+      body: JSON.stringify(payload),
```
tmatsuo, line 220: Can you add indent here and verify the buffer retains the indent?

tmatsuo:
Can we also use this on Cloud Function safely too? FWIW, Cloud Function already automatically inject the rawBody.

chingor13:
> Can we also use this on Cloud Function safely too? FWIW, Cloud Function already automatically inject the rawBody.

Cloud Functions runs its own wrapper server (as does functions-framework) which does this. We only need to do this for our wrapper `http.Server` implementation (for Cloud Run right now).

packages/gcf-utils/test/server.ts
chingor13: Modified the test server to return the rawBody in the response and we assert that the body indentation has not changed.

tmatsuo (approved):

PR #2227, opened 2021-07-02
feat(owl-bot): deploy frontend server in Cloud Run.

tmatsuo:
This change itself should not break anything.

After the deployment, we test the flow from Cloud Run frontend to
the Cloud Tasks. If it works, we can change the Github bot's webhook
URL.


tmatsuo:
Oops I should use the different filename for the Dockerfile

chingor13 (approved):

PR #2229, opened 2021-07-02
fix: track PR author in auto-approve metric

sofisl:


packages/auto-approve/src/auto-approve.ts
```diff
@@ -259,6 +259,14 @@ export function handler(app: Probot) {
               repo: `${owner}/${repo}`,
               pr: prNumber,
             });
+            if (pr.pull_request.user.login === 'renovate-bot') {
+              logger.metric('auto_approve.approved_tagged', {
+                repo: `${owner}/${repo}`,
+                pr: prNumber,
+                renovate_bot: true,
```
chingor13, line 266: this will duplicate the raw count of `auto_approve.approved_tagged`.

WDYT about adding `pr_author` to all the metrics?

chingor13 (changes requested):

packages/auto-approve/src/auto-approve.ts
sofisl: Good call!

chingor13 (approved):

PR #2240, opened 2021-07-03
fix: return a 400 on requests with missing or bad signatures

### LINKED ISSUE #2224: gcf-utils: the server returns 200 when the signature verification fails
I'm trying to deploy a new bot on Cloud Run and observed that the server returns the status code 200 even when the signature verification failed.

chingor13:
Returns a 400 on signature verification errors.

Fixes #2224

### DIFF
--- a/packages/gcf-utils/test/server.ts
+++ b/packages/gcf-utils/test/server.ts
@@ -38,140 +38,258 @@ const sandbox = sinon.createSandbox();
 
 describe('GCFBootstrapper', () => {
   describe('server', () => {
-    let server: http.Server;
-    const bootstrapper = new GCFBootstrapper({
-      projectId: 'test-project',
-      functionName: 'test-bot',
-      location: 'some-location',
-    });
-    const issueSpy = sandbox.stub();
-    const repositoryCronSpy = sandbox.stub();
-    const installationCronSpy = sandbox.stub();
-    const globalCronSpy = sandbox.stub();
-    const pubsubSpy = sandbox.stub();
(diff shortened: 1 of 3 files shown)


packages/gcf-utils/test/server.ts
```diff
@@ -38,140 +38,258 @@ const sandbox = sinon.createSandbox();
...
+            const client = await auth.fromAPIKey('abc123');
+            return client.request(opts);
+          }
+        );
+        enqueueTask = sandbox.stub(bootstrapper.cloudTasksClient, 'createTask');
+        sandbox
+          .stub(bootstrapper, 'getAuthenticatedOctokit')
+          .resolves(new Octokit());
+
+        server = bootstrapper
```
bcoe, line 221: nit: I wonder if it would be worth pulling this into a helper method, as our `before` is ending up with a log of logic (might be nice to have method names that describe it).

bcoe (approved):
Non blocking suggestion.

tmatsuo:
To be honest, I don't think we need the special treatment for requests with TriggerType Task. I would just return 400 for all cases. Less code, less test.

Theoretically all the tasks in Cloud Tasks should have correct signature. Even if we have tasks with bad signature, it will probably get retried up to 10 by default. Those requests should be quick.

For the cases where someone is spoofing requests from Cloud Tasks, the status code doesn't matter.

However, I'm still fine merging this.

tmatsuo (approved):

tmatsuo (approved):

Issue #2246, opened 2021-07-07
snippet-bot: Check does not run when a PR only renames files


busunkim96:
See https://github.com/GoogleCloudPlatform/python-docs-samples/pull/6267

This PR changes a directory name and does not change content inside the files. snippet-bot doesn't report status to the PR, even with the `snippet-bot:force-run` label.

tmatsuo:
@busunkim96 Thanks for the report. I found the log. The bot dies with a large blob (in this case, `train.csv`).

```
"    HttpError: This API returns blobs up to 1 MB in size. The requested blob is too large to fetch via the API, but you can use the Git Data API to request blobs up to 100 MB in size.: {"resource":"Blob","field":"data","code":"too_large"}"
```

For a band aid, we can skip those files (ignoring those large files, it's better than dying). For a long term solution, we can use the Git Data API to support those large files.

tmatsuo:
Maybe I'll downgrade this to P2 once #2248 lands.

busunkim96:
Thanks Takashi!

tmatsuo:
Now the bot should survive large files.
If files larger than 1MB has region tags, there will be a problem, but pragmatically I don't think there's a file like that. I may be wrong, let me know if you find such a file.

tmatsuo:
Filed: https://github.com/googleapis/repo-automation-bots/issues/2703

PR #2248, opened 2021-07-07
fix(snippet-bot): skip errors when fetching renamed files

tmatsuo:
part of #2246

This is not a complete solution, but pragmatically it will work well.
The bot used to die when there is a file larger than 1MB.

With this PR, the bot ignores such an error. Most of the cases, those
large files are data files and they normally don't include region tags.

### DIFF
--- a/packages/snippet-bot/src/region-tag-parser.ts
+++ b/packages/snippet-bot/src/region-tag-parser.ts
@@ -19,6 +19,7 @@ import {isFile} from './utils';
 import {Octokit} from '@octokit/rest';
 import parseDiff from 'parse-diff';
 import fetch from 'node-fetch';
+import {logger} from 'gcf-utils';
 
 /**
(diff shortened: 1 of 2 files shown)


packages/snippet-bot/src/region-tag-parser.ts
```diff
@@ -97,65 +98,75 @@ export async function parseRegionTagsInPullRequest(
...
-        'base64'
-      ).toString('utf8');
+      try {
+        const blobBeforeRename = await octokit.repos.getContent({
+          owner: owner,
+          repo: repo,
+          path: file.from,
+          ref: sha,
+        });
+        if (!isFile(blobBeforeRename.data)) {
```
sofisl, line 108: I don't understand why you'd need to specify this case, if you're simply continuing. Why wouldn't you assert the opposite?

sofisl (approved):

PR #2253, opened 2021-07-08
fix(gcf-utils)!: upgrade to probot 12

tmatsuo:
Making this a major bump.


tmatsuo:
The linkinator job fails because it rewrites only the `branch` portion of the URL. If the PR is coming from another repo (like personal fork), it needs to rewrites the `owner` of the repository as well.

tmatsuo:
@JustinBeckwith FYI
I pushed probot-12 branch to this repo so that the docs job will pass.

bcoe (approved):

PR #2254, opened 2021-07-09
feat(canary): add Pub/Sub handler

### LINKED ISSUE #2182: Add Pub/Sub handler  to canary-bot
Currently canary-bot has a normal github webhook hander and several cron handlers. It's great if there's Pub/Sub handler too.

tmatsuo:
also add corresponding code to serverless proxy

fixes #2182

### DIFF
--- a/packages/canary-bot/src/canary-bot.ts
+++ b/packages/canary-bot/src/canary-bot.ts
@@ -13,9 +13,11 @@
 // limitations under the License.
 
 /* eslint-disable-next-line node/no-extraneous-import */
-import {Probot} from 'probot';
+import {Probot, Logger} from 'probot';
 /* eslint-disable-next-line node/no-extraneous-import */
 import {components} from '@octokit/openapi-types';
+/* eslint-disable-next-line node/no-extraneous-import */
+import {Octokit} from '@octokit/rest';
 import * as fs from 'fs';
 import {resolve} from 'path';
 import {logger, addOrUpdateIssueComment} from 'gcf-utils';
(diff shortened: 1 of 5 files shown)


packages/canary-bot/src/canary-bot.ts
```diff
@@ -40,6 +42,14 @@ const cronIssueTitle = 'A canary is chirping';
 const myRepositoryName = 'repo-automation-bots';
 const myOrganizationName = 'googleapis';
 
+interface PubSubContext {
+  github: Octokit;
+  readonly event: string;
+  log: Logger;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  payload: any;
```
JustinBeckwith, line 50: Would it be possible to make this interface `<T = any>` so consumers can attach a type to it?

packages/canary-bot/src/canary-bot.ts
```diff
@@ -117,4 +127,12 @@ export = (app: Probot) => {
       );
     }
   });
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  app.on('pubsub.message' as any, async (context: PubSubContext) => {
```
JustinBeckwith, line 132: I've found you can cast to `as '*'` and get the same effect without the `any`.

serverless-scheduler-proxy/main.go
```diff
@@ -117,6 +118,23 @@ func rewriteBotContainerURL(c botConfig) func(*http.Request) {
 	}
 }
 
+func rewriteBotCanaryURL(c botConfig) func(*http.Request) {
```
JustinBeckwith, line 121: I had to stare at this code for longer than I'd care to admit before I realized I wasn't looking at JavaScript.

JustinBeckwith (approved):

packages/canary-bot/src/canary-bot.ts
tmatsuo: Sure

packages/canary-bot/src/canary-bot.ts
tmatsuo: Unfortunately we can not. flakybot and owl-bot both use `any` for the pubsub handler.

PR #2272, opened 2021-07-14
feat: make it possible to disable failure checker for the given repo

alexander-fenster:
Let's allow setting
```yaml
disableFailureChecker: true
```
in `release-please.yml` to disable the failure checker for some repositories where it spams us for no reason. E.g. in gapic-generator-typescript we do have `releaseType: node` but never publish to NPM, so the checker gets triggered for each release.


tmatsuo:
I think you'll need to modify these files too:

https://github.com/googleapis/repo-automation-bots/blob/master/packages/release-please/src/config-schema.json

https://github.com/googleapis/repo-automation-bots/blob/master/packages/release-please/src/config-constants.ts

alexander-fenster:
Thanks Takashi, I did that (also fixed some syntax errors and formatting in the JSON schema - not all tools properly understand trailing commas).

tmatsuo:
@alexander-fenster 

> Thanks Takashi, I did that (also fixed some syntax errors and formatting in the JSON schema - not all tools properly understand trailing commas).

Thank you for fixing the syntax!

BTW, this schema change raises another question. 

Do we want `disableFailureChecker` in each branch, or on top level config?

If we want to have `disableFailureChecker` in each branch, shouldn't failurechecker also respect it?

If we only want to have the top level `disableFailureChecker`, I think we should define the field directly in [`ConfigurationOptions`](https://github.com/googleapis/repo-automation-bots/blob/767a65430bc259e6c11e1c695629e6170152de83/packages/release-please/src/config-constants.ts#L36) alongside with `primaryBranch`.

WDYT?

alexander-fenster:
I think the `disableFailureChecker` functionality is on the same level as `releaseType` functionality, so what I did was listing the `disableFailureChecker` everywhere where `releaseType` is listed. As I can see from the code, the failure checker does not care about any other configuration options.

So, in general, I agree that it makes sense for it to be defined in the top level `ConfigurationOptions`, but at the same time, it makes sense for these two parameters to live together. Not sure what is the right way here - what would you suggest? :) 

Thanks!

tmatsuo:
@alexander-fenster 

In general I agree that `disableFailureChecker` should be on the same level as `releaseType`, but in failurechecker implementation, it only look at the top level `disableFailureChecker`.

My concern is that if people configure `disableFailureChecker` in any branches, that configuration will be ignored, and it's a confusing behavior.

My suggestion here is to (temporarily) put `disableFailureChecker` only at the top level.

My second suggestion is to allow `disableFailureChecker` in branch config and also make a change in failurechecker.

alexander-fenster:
@tmatsuo Thank you, for now I moved the `disableFailureChecker` option to the top level config as you suggested.

tmatsuo (approved):

PR #2274, opened 2021-07-15
fix: release-please delay 5 seconds on pushes to allow GitHub indexing to catch up

chingor13:
Fixes https://github.com/googleapis/release-please/issues/826

### DIFF
--- a/packages/release-please/src/release-please.ts
+++ b/packages/release-please/src/release-please.ts
@@ -57,6 +57,7 @@ interface GitHubAPI {
 }
 
 const DEFAULT_API_URL = 'https://api.github.com';
+const DELAY_TIME = process.env.NODE_ENV === 'test' ? 500 : 5000;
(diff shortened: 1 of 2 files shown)


packages/release-please/src/release-please.ts
```diff
@@ -232,6 +237,10 @@ export = (app: Probot) => {
     setLogger(logger);
 
     logger.info(`push (${repoUrl})`);
+
+    // wait 5 seconds for potential GitHub indexing race conditions
+    await delay(DELAY_TIME);
```
bcoe, line 242: Is there a way for us to catch this bug in action? I wonder if we could log a metric before the delay? Check our table in bigquery to see if this delay ever stops the behavior observed.

chingor13:
closing in favor of https://github.com/googleapis/release-please/pull/977

PR #2288, opened 2021-07-20
feat: implement renovate-bot checks for Python and Java

sofisl:


packages/auto-approve/src/auto-approve.ts
```diff
@@ -162,7 +163,7 @@ export function handler(app: Probot) {
       const prConfig = await getBlobFromPRFiles(
         context.octokit,
         owner,
-        repo,
+        repoBase,
```
chingor13, line 166: I *think* we want to keep this one as the head repo - we are trying to get the file from the head (repo that has the changes for the pull request) repo/sha.

packages/auto-approve/src/auto-approve.ts
```diff
@@ -143,13 +143,14 @@ export function handler(app: Probot) {
     async (context: Context<'pull_request'>) => {
       const pr = context.payload;
       const owner = pr.repository.owner.login;
-      const repo = pr.pull_request.head.repo.name;
+      const repoBase = pr.pull_request.base.repo.name;
+      console.log(`REPO BASE: ${repoBase}`);
```
chingor13, line 147:
```suggestion
      logger.info(`REPO BASE: ${repoBase}`);
```
or debug

packages/auto-approve/src/auto-approve.ts
```diff
@@ -172,7 +173,7 @@ export function handler(app: Probot) {
         const codeOwnersFile = await getBlobFromPRFiles(
           context.octokit,
           owner,
-          repo,
+          repoBase,
```
chingor13, line 176: Same on this one.

packages/auto-approve/src/auto-approve.ts
```diff
@@ -243,35 +244,36 @@ export function handler(app: Probot) {
               process.env.PROJECT_ID || '',
               'yoshi-approver'
             );
+            console.log(`SECOND REPO BASE: ${repoBase}`);
```
chingor13, line 247:
```suggestion
            logger.info(`SECOND REPO BASE: ${repoBase}`);
```

chingor13 (changes requested):
I would probably separate into 2 PRs to keep the diffs clean

packages/auto-approve/src/auto-approve.ts
sofisl: Oh right. I can't readily tell from the documentation, but it's also not erroring, so I'll leave it as is.

packages/auto-approve/src/auto-approve.ts
sofisl: Sorry, these were leftover accidentally.

sofisl:
@chingor13, thanks for the quick review: https://github.com/googleapis/repo-automation-bots/pull/2292

packages/auto-approve/src/language-versioning-rules.ts
```diff
@@ -58,4 +58,52 @@ export const languageVersioningRules = [
...
+    ),
+    // This would match: -  google-cloud-storage==1.39.0
+    oldVersion: new RegExp(/-[\s]?(@?[^=]*)==([0-9])*\.([0-9]*\.[0-9]*)/),
+    // This would match: +  google-cloud-storage==1.40.0
+    newVersion: new RegExp(/\+[\s]?(@?[^=]*)==([0-9])*\.([0-9]*\.[0-9]*)/),
+  },
+  {
+    prAuthor: 'renovate-bot',
+    process: 'java-dependency',
+    targetFile: 'pom.xml',
```
lesv, line 77: We should also allow `build.gradle` files.

lesv: @kurtisvg PTAL

packages/auto-approve/test/fixtures/config/valid-schemas/valid-schema6.yml
```diff
@@ -0,0 +1,6 @@
+rules:
+- author: "renovate-bot"
+  title: "^(fix\\(deps\\)|chore\\(deps\\)):"
+  changedFiles: 
+  - "/pom\\.xml$"
+  maxFiles: 3
```
lesv, line 6: We should also allow `build.gradle`

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -80,7 +80,13 @@ export function getTargetFiles(
...
+        changedFile.filename === rule.targetFile
+      ) {
+        targetFiles.push({file: changedFile, fileRule: rule});
+      } else if (
+        rule.prAuthor === author &&
+        changedFile.filename.endsWith(rule.targetFile) &&
+        rule.process === 'java-dependency'
       ) {
         targetFiles.push({file: changedFile, fileRule: rule});
       }
```
lesv, line 92: Should there also be a regex case here?

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -170,12 +191,20 @@ export function getVersions(
...
   const titleRegex = title.match(dependencyRegex);
+
   if (titleRegex) {
     dependencyName = titleRegex[2];
+
+    if (process === 'java-dependency') {
+      if (!dependencyName.includes('com.google.')) {
+        return false;
+      }
+    }
```
lesv, line 207: LGTM - there are some `com.google` artifacts, but they were last released in 2014

packages/auto-approve/src/valid-pr-schema.json
```diff
@@ -199,13 +199,13 @@
                     "enum": [
                       "/package\\.json$",
                       "/pom\\.xml$",
```
lesv, line 201: `build.gradle`

lesv: Also `*.sbt`.

lesv (approved):
Generally approve - though I probably should review your version approval code again.

I left out `build.gradle` & `*.sbt` files.  We have them, but renovateBot hasn't been updating.  So consider that a future feature request after we fix renovateBot.

I've asked Kurtis to take a look.

packages/auto-approve/src/language-versioning-rules.ts
kurtisvg: I'll be honest, I don't really know what I'm looking at here. 

I'm not sure I understand why we are putting a config in the repo if we are also adding code somewhere else that only allows that config to be certain values.

packages/auto-approve/src/language-versioning-rules.ts
```diff
@@ -58,4 +58,52 @@ export const languageVersioningRules = [
...
+    // This would match: chore(deps): update dependency com.google.cloud:google-cloud-datacatalog to v1.4.2 or chore(deps): update dependency com.google.apis:google-api-services-policytroubleshooter to v1-rev20210319-1.32.1
+    dependency: new RegExp(
+      /^(fix\(deps\)|chore\(deps\)): update dependency (@?\S*) to v(\S*)$/
```
kurtisvg, line 80: This regex looks wrong to me - specifically the OR for fix/chore looks like it might resolve to`)|c`, but I think you want `(fix)|(chore)` - it would probably be more readable if it was`^((fix)|(chore))\(deps\)): update dependency (@?\S*) to v(\S*)$/`

packages/auto-approve/src/language-versioning-rules.ts
sofisl: The config is for repos to self-select if they want to use auto-approve. However, as it is a riskier process, we are starting off with a validation method in order to ensure that only certain types of PRs are considered automergeable.

For other cases, the config is pretty much sufficient for ensuring that the PR confirms to certain guidelines. For the renovate-bot case, we need to add many additional requirements as these PRs are coming from a third party.

packages/auto-approve/src/utils-for-pr-checking.ts
sofisl: I don't think we need a regex case here, but I may be misunderstanding. This method basically only ensures that we're grabbing the right file rule from `language-versioning-rules.ts`. Java is a bit different from other languages, since your path to pom.xml can look different, whereas with Node and Python it should be strict equality.

What use case are you thinking about specifically where we might need regex here?

packages/auto-approve/src/valid-pr-schema.json
sofisl: Before we add these, would you mind sending me some sample PRs so that I can confirm the rules/processes would look the same?

packages/auto-approve/src/language-versioning-rules.ts
sofisl: Sure, I've changed it to be more readable. I think both formulations are equivalent.

chingor13 (approved):
@sofisl and I will meet later to discuss possible cleanup refactoring

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -118,17 +125,31 @@ export function getVersions(
...
-    newDependencyName = newVersions[1];
-    newMajorVersion = newVersions[2];
-    newMinorVersion = newVersions[3];
+    if (process === 'java-dependency') {
+      newDependencyName = `${newVersions[1]}:${newVersions[2]}`;
+      newMajorVersion = newVersions[5] || newVersions[7];
+      newMinorVersion = newVersions[6] || newVersions[8];
+      console.log(newDependencyName);
+      console.log(newMajorVersion);
+      console.log(newMinorVersion);
```
parthea, line 147: Did you mean to include `console.log()` here?

parthea (approved):
Approving for python. Only 1 observation, otherwise LGTM.

packages/auto-approve/src/valid-pr-schema.json
sofisl: Per our discussion, we'll add these later on.

packages/auto-approve/src/utils-for-pr-checking.ts
sofisl: Good catch, thanks!

sofisl:
Want to wait for #2292 first.

packages/auto-approve/src/language-versioning-rules.ts
```diff
@@ -58,4 +57,52 @@ export const languageVersioningRules = [
...
+      <groupId>com.google.apis</groupId>
+      <artifactId>google-api-services-policytroubleshooter</artifactId>
+      -      <version>v1-rev20210319-1.31.5</version>
+      or
+      <groupId>com.google.apis</groupId>
+      <artifactId>google-api-services-policytroubleshooter</artifactId>
+-     <version>v1-rev20210319-1.31.5</version>
+    */
+    oldVersion: new RegExp(
+      /<groupId>([^<]*)<\/groupId>[\s]*<artifactId>([^<]*)<\/artifactId>[\s]*-[\s]*<version>(v[0-9]-rev[0-9]*-([0-9]*)\.([0-9]*\.[0-9])|([0-9]*)\.([0-9]*\.[0-9]*))<\/version>[\s]*/
```
tmatsuo, line 91: Not a blocker, but I recommend that we use an xml parser.

packages/auto-approve/src/language-versioning-rules.ts
```diff
@@ -58,4 +57,52 @@ export const languageVersioningRules = [
...
+-     <version>1.4.1</version>
++     <version>1.4.2</version>
+      or
+       <groupId>com.google.apis</groupId>
+       <artifactId>google-api-services-policytroubleshooter</artifactId>
+-      <version>v1-rev20210319-1.31.5</version>
++      <version>v1-rev20210319-1.32.1</version>
+    */
+    newVersion: new RegExp(
+      /<groupId>([^<]*)<\/groupId>[\s]*<artifactId>([^<]*)<\/artifactId>[\s]*-[\s]*<version>(v[0-9]-rev[0-9]*-[0-9]*\.[0-9]*\.[0-9]|[[0-9]*\.[0-9]*\.[0-9]*)<\/version>[\s]*\+[\s]*<version>(v[0-9]-rev[0-9]*-([0-9]*)\.([0-9]*\.[0-9])|([0-9]*)\.([0-9]*\.[0-9]*))<\/version>/
```
tmatsuo, line 105: I think this is a regex against a diff. I also recommend that we use a diff parser.

sofisl: Added in: https://github.com/googleapis/repo-automation-bots/issues/2311

PR #2292, opened 2021-07-20
fix(auto-approve): pull changed files from base repo, not head repo

### LINKED ISSUE #2283: auto-approve: Error: 404, Not Found for GoogleCloudPlatform/python-docs-samples-1/6274
We're seeing a lot of these exceptions.

The `python-docs-samples-1` is a clue that we're trying to use the wrong repo name.

sofisl:
Fixes #2283


chingor13 (approved):

chingor13 (commented):
Ideally, we should have a test case that demonstrates this fix (a pull request from a forked repo where the fork name differs)

packages/auto-approve/src/auto-approve.ts
```diff
@@ -162,7 +163,7 @@ export function handler(app: Probot) {
       const prConfig = await getBlobFromPRFiles(
         context.octokit,
         owner,
```
chingor13, line 165: If this is coming from a fork, this should also probably be the head branch's repo's owner as well.

chingor13 (changes requested):

sofisl:
I agree. I actually spent a good amount of time looking for an example of a webhook event for a fork, but the documentation is pretty sparse in that way (and I didn't have access to Cloud Functions). I can wait to merge if you prefer to add tests first.

chingor13:
The PR coming from a fork seems like a case we should ensure is working with test cases/fixtures.

sofisl:
@chingor13, thanks for providing the webhook event, I added some more tests.

bcoe (approved):

bcoe:
LGTM, let's wait for Jeff's feedback Monday though 👍

chingor13 (approved):

PR #1821, opened 2021-05-20
feat: support multiple .OwlBot.yamls Redux

SurferJeffAtGoogle:
I ran ran locally, scanned all the repositories, and it finished without error.  So, I think it was a space issue.

The first commit reverts the revert.  The second commit fixes the issue.

Issue #1573, opened 2021-03-25
owlbot's config files should be case-insensitive.


SurferJeffAtGoogle:
We should find all case variations of the following names:
.OwlBot.yaml
.OwlBot.lock.yaml
.OwlBot.yml
.OwlBot.lock.yml
owlbot.py

tmatsuo:
What should it do when there're multiple variants? Then we should have a determined order, then we should have it tested.

Rather, how about to create a suggesting PR for renaming the config file(s) to the correct ones, and always use once single file name?

tmatsuo:
Actually if we go that route, there might be a case where multiple variants exist.

In that case, we can file an issue and let the owner decide.

This way, we can always use the single file name, as well as catch most of the filename mistakes.

SurferJeffAtGoogle:
This is a nice to have, and would make the code a lot more complicated.  I'm closing this as "Will not fix."

PR #2310, opened 2021-07-29
feat(owl-bot): support multiple .OwlBot.yamls

tmatsuo:
This is basically rebasing #1821 to the current HEAD.

My plan is to use this branch for staging deployment and reproducing the memory issue and debugging.


packages/owl-bot/src/fetch-configs.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Fetches the configuration files from a github repo.
+ */
+export async function fetchConfigs(
+  octokit: OctokitType,
+  githubRepo: {
+    owner: string;
+    repo: string;
+    ref: string;
+  }
+): Promise<[OwlBotLock | undefined, OwlBotYamlAndPath[]]> {
```
bcoe, line 33: I like that the zip logic is encapsulated in a helper file, makes it easier to swap out if we ever wanted to.

packages/owl-bot/src/handlers.ts
```diff
@@ -257,69 +252,37 @@ export async function refreshConfigs(
...
-      await createIssueIfTitleDoesntExist(
-        octokit,
-        githubOrg,
-        repoName,
-        title,
-        body,
-        logger
-      );
+  try {
+    const [lock, yamls] = await fetchConfigs(octokit, {
```
bcoe, line 256: This will make it easier for us to support config files regardless of case sensitivity or `yaml` vs., `yml` suffix, see:

https://github.com/googleapis/repo-automation-bots/issues/1573

bcoe (commented):
LGTM, not going to approve yet until we've dug into memory bug.

tmatsuo:
superseded by #2420

Issue #1653, opened 2021-04-20
feat(owlbot): add "Update this PR" checkbox in the owlbot comment


tmatsuo:
In the owlbot comment, can it provide something like the following:

- [ ] Update this PR

Then the bot listens to the issue comment update, then the owlbot will start over from the current head, then push new commits to the PR?

SurferJeffAtGoogle:
cc @Neenu1995 

https://github.com/googleapis/java-dialogflow-cx/pull/240 

![image](https://user-images.githubusercontent.com/11778530/122996447-ba8ba580-d35f-11eb-8cf4-2d756078f869.png)
 
![image](https://user-images.githubusercontent.com/11778530/122996406-ae9fe380-d35f-11eb-8895-a1266091533c.png)

PR #2329, opened 2021-08-04
feat: checkbox to regenerate a pull request

SurferJeffAtGoogle:
Fixes https://github.com/googleapis/repo-automation-bots/issues/1653.


SurferJeffAtGoogle:
We'll need to pair up to deploy this pull request.  We need to create a new Cloud Build Trigger, for example.

packages/owl-bot/src/copy-code.ts
```diff
@@ -42,29 +42,67 @@ const readFileAsync = promisify(readFile);
...
+  return `Source-Link: ${sourceLink}\n`;
+}
+
+/**
+ * Finds the source link in a pull request body.  Returns the empty string if
+ * not found.
+ */
+export function findSourceHash(prBody: string): string {
+  const match =
+    /https:\/\/github.com\/googleapis\/googleapis-gen\/commit\/([0-9A-Fa-f]+)/.exec(
```
bcoe, line 62: nit: sometimes I find it's worth using named capture groups for clarity:


```js
/https:\/\/github.com\/googleapis\/googleapis-gen\/commit\/(?<hash>[0-9A-Fa-f]+)/
return match ? match.hash : '';
```

packages/owl-bot/src/copy-code.ts
```diff
@@ -42,29 +42,67 @@ const readFileAsync = promisify(readFile);
...
+      prBody
+    );
+  return match ? match[1] : '';
+}
+
+/**
+ * A return value indicating that code was copied into a local branch.
+ */
+interface LocalCopy {
+  kind: 'LocalCopy';
```
bcoe, line 72: Also a nit, I believe that TypeScript has `enums`, which might be worth considering for these fields.

packages/owl-bot/src/copy-code.ts
```diff
@@ -116,19 +156,52 @@ ${err}`,
...
+  const destBranch = 'owl-bot-' + uuidv4();
+  const dest = await copyCodeIntoLocalBranch(
+    sourceRepo,
+    sourceRepoCommitHash,
+    destRepo,
+    destBranch,
+    octokitFactory,
+    logger
+  );
+  if (dest.kind === 'CreatedGithubIssue') {
```
bcoe, line 189: If I'm understanding correctly, `dest.kind` is used to create a state machine where we react to the responses of several steps in the pull request creation process that could have multiple outcomes, e.g., opening an issue rather than updating a PR?

packages/owl-bot/src/handlers.ts
```diff
@@ -335,3 +342,95 @@ Please fix this as soon as possible so that your repository will not go stale.`;
...
+export async function handlePullRequestEdited(
+  appId: number,
+  privateKey: string,
+  project: string,
+  trigger: string,
+  payload: PullRequestEditedEvent,
+  octokit: OctokitType,
+  logger = console
+) {
+  const head = payload.pull_request.head.repo.full_name;
```
bcoe, line 359: I would be tempted to do this data munging in `owl-bot.ts`, then `handlePullRequestEdited` could just accept the parameters that it needs, and it doesn't need to know about the full `payload` structure.

packages/owl-bot/test/scan-googleapis-gen-and-create-pull-requests.ts
```diff
@@ -220,3 +225,45 @@ describe('scanGoogleapisGenAndCreatePullRequests', function () {
...
+      abcRepo,
+      sourceHash,
+      destRepo,
+      'test-branch',
+      factory
+    );
+    // Confirm new pull request was pushed to test-branch.
+    const destDir = destRepo.getCloneUrl();
+    const gitLog = cmd('git log test-branch', {cwd: destDir}).toString('utf-8');
+    assert.match(
```
bcoe, line 262: I can help add a test for the route in owl-bot.ts too, after we land this; if we go with my suggestion of not passing payload into the handler, then all that test would exercise would be passing the proper parameters into the handler.

packages/owl-bot/src/handlers.ts
```diff
@@ -335,3 +342,95 @@ Please fix this as soon as possible so that your repository will not go stale.`;
...
+    createComment(`Owl Bot could not regenerate this pull request because the body is missing a source hash.
+
+A source hash in the source link looks like this:
+${sourceLine}`);
+    return;
+  }
+
+  try {
+    const cb = core.getCloudBuildInstance();
+    await cb.runBuildTrigger({
```
bcoe, line 413: I believe this will return a long running operation which we then need to await completion on, this is what's happening from line `134` on here:

https://github.com/googleapis/repo-automation-bots/blob/ae6615f4dfc002bdd105f58e79f85927d8834c1c/packages/owl-bot/src/core.ts#L134

What we could perhaps do is pull this into a `cloud-build.ts` wrapper, that handles this nuisance for us.

bcoe (changes requested):
This approach is looking great to me, left mostly nits.

One bug I noticed was that running cloud build returns immediately, but the job itself would need to be awaited, if we want to provide feedback about the success or failure of the job, I linked an example of how to do this.

packages/owl-bot/src/copy-code.ts
SurferJeffAtGoogle: Agreed.  This regular expression has a single capture group, so I don't see much potential for confusion.

packages/owl-bot/src/copy-code.ts
SurferJeffAtGoogle: I thought about that too, but I don't see an advantage in this particular case.  A typo in any instance of these literal strings would trigger a compile error.

packages/owl-bot/src/copy-code.ts
SurferJeffAtGoogle: `.kind` is just used to distinguish the two possible return values.  I'm following the typescript convention described here: https://www.typescriptlang.org/docs/handbook/2/narrowing.html

packages/owl-bot/src/handlers.ts
SurferJeffAtGoogle: Ok.  Will do.

packages/owl-bot/test/scan-googleapis-gen-and-create-pull-requests.ts
SurferJeffAtGoogle: I'll add the test.

packages/owl-bot/src/handlers.ts
SurferJeffAtGoogle: I think I got confused because I'd expect functions named `handlePullRequestLabeled()` and `handlePullRequestEdited()` to live in a file called `handlers.ts`.

packages/owl-bot/src/handlers.ts
SurferJeffAtGoogle: Done.

packages/owl-bot/src/handlers.ts
SurferJeffAtGoogle: In this case, there is no github check that we need to update.  The build job does the work of updating the pull request with either a success or failure.  What would we do after we waited for the result here?

packages/owl-bot/test/scan-googleapis-gen-and-create-pull-requests.ts
SurferJeffAtGoogle: Done.

SurferJeffAtGoogle:
This pull request is ready for another look.

bcoe (approved):

bcoe (changes requested):

PR #2334, opened 2021-08-05
Revert "deps(release-please): release please with fix for pre-releases"

chingor13:
We're seeing a ton of exceptions `PathNotFoundError: Could not find requested path: google-api-client/src/main/java/com/google/api/client/googleapis/GoogleUtils.java` starting 18 hours ago which matches up with this release

See #2335


alexander-fenster (approved):
stamped

PR #2360, opened 2021-08-12
fix(owlbot): allow a certificate to be passed in

bcoe:
Allow a certificate to be passed in to Octokit factory, rather than always reading from disk.


bcoe:
@SurferJeffAtGoogle in app, certificate is in environment variable, rather than in a file on disk.

tmatsuo (approved):

PR #2441, opened 2021-09-04
fix(owl-bot): Fallback to Pull, Push when Push Fails When Creating PR

### LINKED ISSUE #2307: bug: identify and silence owl-bot post-processor failures when PRs attempt to push to a branch on which the post-processor has already run
See #2304 and #2306 

```
To https://github.com/googleapis/python-billing.git
 ! [rejected]        release-v1.3.2 -> release-v1.3.2 (fetch first)
error: failed to push some refs to 'https://github.com/googleapis/python-billing.git'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
```

This happens when a PR is created, then the post-processor runs on it, then the PR author attempts to push another commit to that branch without first pulling down the changes. However, ultimately, the post-processor runs fine on Github since the user can force-push in order to retrigger the post-processor. So, perhaps there is a way of silencing these errors, or perhaps a way of delaying the post-processor for a few seconds to ensure another commit pushed to that branch recently does not cause this error.

danielbankhead:
Fixes #2307


packages/owl-bot/cloud-build/scan-googleapis-gen-and-create-pull-requests.yaml
```diff
@@ -33,6 +33,12 @@ steps:
       - --global
       - user.name
       - 'Owl Bot'
+  - name: 'gcr.io/cloud-builders/git'
```
tmatsuo, line 36: My guess is we need to set $HOME for correctly preserving the configs.

(Also applies to the steps above)
@SurferJeffAtGoogle

packages/owl-bot/cloud-build/update-lock-branch.yaml
```diff
@@ -49,6 +49,12 @@ steps:
       - --global
       - user.name
       - 'Owl Bot'
+  - name: 'gcr.io/cloud-builders/git'
+    args:
+      - config
+      - --global
+      - pull.rebase
+      - 'false'
```
tmatsuo, line 57: ditto

danielbankhead: Thanks, updated

packages/owl-bot/src/create-pr.ts
```diff
@@ -70,7 +70,16 @@ export async function createPullRequestFromLastCommit(
...
   cmd(`git remote set-url origin ${pushUrl}`, {cwd: localRepoDir});
-  cmd(`git push origin ${branch}`, {cwd: localRepoDir});
+
+  try {
+    cmd(`git push origin ${branch}`, {cwd: localRepoDir});
+  } catch (e) {
+    logger.info('Falling back to pull, push', e);
+
+    // In case we're missing commits from an existing branch
+    cmd(`git pull origin ${branch}`, {cwd: localRepoDir});
```
tmatsuo, line 80: Question.

Sometimes the PR author force pushing to the PR head branch. What will happen in such cases?

danielbankhead: That case would be handled in the follow flow: the `push` in the try block would fail, which we would then attempt to `pull` then `push`

tmatsuo: @SurferJeffAtGoogle 

I think when this code runs, the local git repo may have some changes.
If the remote branch is force pushed and containing different history, it may fail.

WDYT?

packages/owl-bot/cloud-build/scan-googleapis-gen-and-create-pull-requests.yaml
SurferJeffAtGoogle: It looks to me like HOME is being set on lines 39 and 47.

SurferJeffAtGoogle (approved):

packages/owl-bot/cloud-build/copy-code-again.yaml
```diff
@@ -37,6 +37,14 @@ steps:
       - 'Owl Bot'
     env:
       - 'HOME=/workspace'
+  - name: 'gcr.io/cloud-builders/git'
```
SurferJeffAtGoogle, line 40: Because the current branch is being force pushed, there's no need to pull here.

SurferJeffAtGoogle:
After sleeping on it, I think the only place we need to pull before pushing is in update-pr.yaml.  In these other places, I think it's preferable that the push fail.  Either way, it's not urgent.  Maybe by Monday, I'll change my mind again...

PR #2471, opened 2021-09-10
docs: add README and links for release-trigger

chingor13:


bcoe (commented):
If configured would there potentially be a race condition between autorelease and this automation, until we decommission  release-tool?

Would it be of any concern if autorelease beat this automation to the punch?

bcoe (approved):

chingor13:
> If configured would there potentially be a race condition between autorelease and this automation, until we decommission release-tool?
> 
> Would it be of any concern if autorelease beat this automation to the punch?

It is possible, but the automations do the same thing (follow the same labeling scheme). In the worst case, we trigger the release job twice and one fails.

PR #2480, opened 2021-09-10
build: upgrade to TypeScript 4.4.3

JustinBeckwith:
Well that was fun.  This upgrades us to the latest version of TypeScript across all the bots, and does some lightweight cleanup as required along the way.


packages/auto-approve/src/check-config.ts
```diff
@@ -14,11 +14,12 @@
 
 // This file manages the logic to check whether a given config file is valid
 
-// eslint-disable-next-line node/no-extraneous-import
 import yaml from 'js-yaml';
 import Ajv from 'ajv';
 import {Octokit} from '@octokit/rest';
 import {Configuration} from './auto-approve';
+// eslint-disable-next-line node/no-extraneous-import
+import {RequestError} from '@octokit/request-error';
```
feywind, line 22: It's weird that eslint would see this as an extraneous import, if you're actually using it below O_o

Issue #2339, opened 2021-08-06
OwlBot Failing due to too many subsequent updates


bcoe:
There's an error happening on OwlBot:

```
 Error: too many OwlBot updates created in a row for googleapis/java-aiplatform at runPostProcessor (/workspace/build/src/owl-bot.js:267:15) at async Object.handlePullRequestLabeled (/workspace/build/src/owl-bot.js:173:5) at async /workspace/build/src/owl-bot.js:57:9 at async Promise.all (index 0) at async /workspace/node_modules/gcf-utils/build/src/gcf-utils.js:371:21
Error: too many OwlBot updates created in a row for googleapis/java-aiplatform at runPostProcessor (/workspace/build/src/owl-bot.js:267:15) at async Object.handlePullRequestLabeled (/workspace/build/src/owl-bot.js:173:5) at async /workspace/build/src/owl-bot.js:57:9 at async Promise.all (index 0) at async /workspace/node_modules/gcf-utils/build/src/gcf-utils.js:371:21
```

As a result of the PR:

https://github.com/googleapis/java-aiplatform/pull/449/commits

We have a safety in a place that doesn't allow three subsequent updates from OwlBot, I believe this safety is being incorrectly triggered on this PR, even though it's had new updates since OwlBot's changes.

CC: @Neenu1995

parthea:
@danielbankhead  I'll keep this PR in the failed state in case it helps with development:
https://github.com/googleapis/python-datastream/pull/26 .

sofisl:
It logs a failure, but also doesn't re-run the container, so it leaves the PR in a pending state.

and doesn't re-run templates, if someone was purposefully running it again.

SurferJeffAtGoogle:
Saw it again here:
https://github.com/googleapis/nodejs-dialogflow/pull/871
![image](https://user-images.githubusercontent.com/11778530/132383930-2c26340d-f96f-420c-94ac-2d5a98415d27.png)

bcoe:
This is cropping up very frequently on Java repos, e.g.,

https://github.com/googleapis/java-artifact-registry/pull/142

I think we need to prioritize this work sooner rather than later.

bcoe:
@danielbankhead opened for now while we debug.

PR #2482, opened 2021-09-10
fix(owl-bot): Determine loop by the last few commits

danielbankhead:
Fixes: #2339


packages/owl-bot/src/core.ts
```diff
@@ -502,13 +502,24 @@ async function hasOwlBotLoop(
...
   ).data;
-  let count = 0;
-  for (const commit of commits) {
-    if (commit?.author?.login === OWLBOT_USER) count++;
-    else count = 0;
-    if (count >= circuitBreaker) return true;
+
+  // get the most recent commits (limit by circuit breaker)
+  const lastFewCommits = commits
+    .sort((a, b) => {
```
danielbankhead, line 508: @bcoe we may not need this sort, however I didn't see any documented guarantees `octokit.pulls.listCommits` returns sorted by date

tmatsuo:
Can we take another approach?

If a human add a `owlbot:run` label, maybe the bot should always run the task.

tmatsuo (approved):

Issue #1692, opened 2021-04-27
FR: Allow Docker entry-point or cmd to be specified in .OwlBot.yaml


jskeet:
Currently, I need to create a Docker image for OwlBot to run.

All I want my Docker image to do is run a bash script. Currently that means I have to have a Dockerfile of the form:

```sh
FROM mcr.microsoft.com/dotnet/core/sdk:3.1-bionic
CMD ["tools/owlbot-postprocess.sh"]
```

... then build that as a Docker image and publish it.

I'd really prefer to be able to skip the Dockerfile, and have a .OwlBot.yaml file of:

```yaml
docker:
  image: mcr.microsoft.com/dotnet/core/sdk:3.1-bionic
  cmd: tools/owlbot-postprocess.sh
```

(I'm aware there are [differences between ENTRYPOINT and CMD](https://www.ctl.io/developers/blog/post/dockerfile-entrypoint-vs-cmd/) but I don't have strong opinions - I'd be happy enough with an entry-point array specified in the YAML instead. The exact format doesn't bother me nearly as much as not having to have a Dockerfile and publish a docker image somewhere.)

SurferJeffAtGoogle:
Jon, there's another bug that demands a standard entry point into the docker image:
https://github.com/googleapis/repo-automation-bots/issues/1700

To fix issue 1700, I proposed adding an another standard entry point `print-changes.sh` that would print changes since the last release of the docker image.  That would then require a custom docker image.  Once we require a custom docker image, then this bug is moot.

Thoughts?

jskeet:
@SurferJeffAtGoogle: That feels like it's saying "In order to support this feature you don't need if you're using a non-custom Docker image, you need to use a custom Docker image" :) (If the Docker image isn't changing, why do we need a way of saying what's changed?)

Perhaps more constructively, I'd suggest:

- Allow a print-changes-cmd to be specified as well as a "regular" cmd
- If cmd is present, but print-changes-cmd isn't, but the lock file *is* updated, the PR doesn't have useful information. (If you want useful information printed automatically, you do need the custom image

I'd really, really prefer to avoid requiring a custom Docker image if possible.

SurferJeffAtGoogle:
Ok.  I'll make in configurable some way.

PR #2522, opened 2021-09-15
WIP: add a CMD parameter to .OwlBot.yaml 

SurferJeffAtGoogle:
Will fix https://github.com/googleapis/repo-automation-bots/issues/1692

Google Cloud Build does not have any conditional statements.  Therefore, supporting a CMD in .OwlBot.yaml requires maintaining a copy of update-pr.yaml that is only one line different, and also requires creating and maintaining a separate build trigger.

### DIFF
--- a/packages/owl-bot/cloud-build/update-pr.yaml
+++ b/packages/owl-bot/cloud-build/update-pr.yaml
@@ -12,6 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# This cloud build gets invoked every time a new pull request is created or
+# a new commit is added to a pull request that's being watched by Owl Bot.
+# It's triggered programmatically by event handlers in Owl Bot.
+# It runs the post processor specified in the environment variable 
+
 steps:
   # Check git version
   - name: 'gcr.io/cloud-builders/git'
--- a/packages/owl-bot/cloud-build/update-pr-cmd.yaml
+++ b/packages/owl-bot/cloud-build/update-pr-cmd.yaml
@@ -19,7 +19,7 @@
 
 ##############################################################################
 # If you make changes to this file, you mast make the same changes to
-# update-pr-cmd.yaml.
+# update-pr.yaml.
 ##############################################################################
 
 steps:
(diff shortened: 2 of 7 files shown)


SurferJeffAtGoogle:
@jskeet I think this PR is and its future maintenance are growing far more complex than maintaining a two-line Docker file.  Thoughts?

jskeet:
If it's this vs *just* maintaining a `Dockerfile`, I agree. If it's this vs maintaining a `Dockerfile` *and the docker image* (making it available somewhere, rebuilding it periodically in case the base has changed etc) - and that being done by every user rather than the complexity here being maintained by a single team - then I think it's less clearcut.

packages/owl-bot/cloud-build/update-pr.yaml
```diff
@@ -12,6 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# This cloud build gets invoked every time a new pull request is created or
+# a new commit is added to a pull request that's being watched by Owl Bot.
+# It's triggered programmatically by event handlers in Owl Bot.
+# It runs the post processor specified in the environment variable 
+
+##############################################################################
```
bcoe, line 20: I wonder, could we have a step run conditionally ([there seem to be some hacky ways to do so](https://stackoverflow.com/questions/58235945/google-cloud-build-conditional-step)) and that way eliminate the need for a second cloud build file?

SurferJeffAtGoogle: I don't think the hacky solutions can be applied to our situation.

SurferJeffAtGoogle:
> If it's this vs _just_ maintaining a `Dockerfile`, I agree. If it's this vs maintaining a `Dockerfile` _and the docker image_ (making it available somewhere, rebuilding it periodically in case the base has changed etc) -

We'd make it publicly available in the GCP project cloud-devrel-public-resources.  We could set up cloud build to automatically rebuild and republish it nightly.

> and that being done by every user rather than the complexity here being maintained by a single team - then I think it's less clearcut.

I can't imagine a case when users would want to build the docker image themselves, rather than pull it from the public GCP project.

jskeet:
> We'd make it publicly available in the GCP project cloud-devrel-public-resources. We could set up cloud build to automatically rebuild and republish it nightly.

Okay. Let's talk about exactly how that would work - I suggest an actual meeting might be better than GitHub for that.

PR #2528, opened 2021-09-17
fix(flakybot): add a new comparator for finding existing issue

### LINKED ISSUE #2105: flakybot: keeps opening/closing new bugs for same issue
Have an issue in google-cloud-go that's essentially being triggered by a progressive rollout, so the issue is destined to be intermediately flaky for days to weeks.  This is manifesting as failures in a test.

Flakybot is flagging the failures as expected, but using a new issue each time and referencing a much older issue (3612) as the rationale for opening a new bug each time.

instances:
https://github.com/googleapis/google-cloud-go/issues/4277
https://github.com/googleapis/google-cloud-go/issues/4279
https://github.com/googleapis/google-cloud-go/issues/4285

tmatsuo:
fixes #2105

### DIFF
--- a/packages/flakybot/test/flakybot.test.ts
+++ b/packages/flakybot/test/flakybot.test.ts
@@ -1293,52 +1293,6 @@ describe('flakybot', () => {
         scopes.forEach(s => s.done());
       });
 
-      it('reopens the original flaky issue when there is a duplicate', async () => {
-        getConfigWithDefaultStub.resolves(DEFAULT_CONFIG);
-        const payload = buildPayload('one_failed.xml', 'golang-samples');
-
-        const title = formatTestCase({
-          package:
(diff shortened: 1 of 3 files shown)


packages/flakybot/test/flakybot.test.ts
```diff
@@ -1293,52 +1293,6 @@ describe('flakybot', () => {
         scopes.forEach(s => s.done());
       });
 
-      it('reopens the original flaky issue when there is a duplicate', async () => {
```
sofisl, line 1296: Would we want to ensure here that we're reopening the latest flaky issue?

tmatsuo: I don't think so.

This is one way or another. Fix the reported bug, or try to preserve the existing behavior.

I don't think the existing behavior is valuable enough to abandon the fix.

sofisl (approved):

PR #2531, opened 2021-09-18
feat: new secret-rotator service

sofisl:


packages/secret-rotator/cloudbuild.yaml
```diff
@@ -0,0 +1,65 @@
...
+    id: "build-docker"
+    waitFor: ["-"]
+    dir: packages/secret-rotator
+    args:
+      - "build"
+      - "-f"
+      - "Dockerfile"
+      - "-t"
+      - "gcr.io/$PROJECT_ID/secret-rotator"
+      - "."
```
chingor13, line 35: We should pick whether we're doing the compile in a bash script or the Dockerfile, not both.

IMO, it should be a multi-stage docker image like https://github.com/googleapis/repo-automation-bots/blob/main/packages/release-trigger/Dockerfile

packages/secret-rotator/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+`npm i`
+
+## Running tests:
+
+`npm test`
+
+## Contributing
+
+If you have suggestions for how `release-brancher` could be improved, or want to report a bug, open an issue! We'd love all and any contributions.
```
chingor13, line 14:
```suggestion
If you have suggestions for how `secret-rotator` could be improved, or want to report a bug, open an issue! We'd love all and any contributions.
```

packages/secret-rotator/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "nock": "^13.1.3",
+    "sinon": "^11.1.2"
+  },
+  "dependencies": {
+    "@google-cloud/secret-manager": "^3.10.0",
+    "@googleapis/iam": "^0.2.0",
+    "express": "^4.17.1",
+    "gcf-utils": "^13.0.2",
+    "google-auth-library": "^7.9.1",
+    "typescript": "^4.4.3"
```
chingor13, line 49: `typescript` is probably a `devDependency`

packages/secret-rotator/src/app.ts
```diff
@@ -0,0 +1,70 @@
...
+    version: 'v1',
+    auth,
+  });
+}
+
+async function authenticateSecretManager(
+  auth: GoogleAuth
+): Promise<SecretManagerServiceClient> {
+  return new SecretManagerServiceClient({auth});
+}
```
chingor13, line 41: I don't feel like these functions add much vs. inlining the one line of code.

packages/secret-rotator/src/helper.ts
```diff
@@ -0,0 +1,135 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {iam_v1} from '@googleapis/iam';
+import {SecretManagerServiceClient} from '@google-cloud/secret-manager';
+import {logger} from 'gcf-utils';
+
+export class Helper {
```
chingor13, line 19: Maybe rename to `SecretRotator`? `Helper` is not very descriptive here.

packages/secret-rotator/src/helper.ts
```diff
@@ -0,0 +1,135 @@
...
+    serviceAccountEmail: string
+  ) {
+    const name = `projects/${serviceAccountProjectId}/serviceAccounts/${serviceAccountEmail}`;
+    const key = await client.projects.serviceAccounts.keys.create({
+      name,
+      requestBody: {
+        privateKeyType: 'TYPE_GOOGLE_CREDENTIALS_FILE',
+      },
+    });
+    console.log(key.data.privateKeyData);
```
chingor13, line 55: Probably don't want to log any private key info

packages/secret-rotator/src/helper.ts
```diff
@@ -0,0 +1,135 @@
...
+    serviceAccountEmail: string,
+    secretManagerProjectId: string,
+    secretName: string
+  ) {
+    this.iamClient = iamClient;
+    this.secretManagerClient = secretManagerClient;
+    this.serviceAccountProjectId = serviceAccountProjectId;
+    this.serviceAccountEmail = serviceAccountEmail;
+    this.secretManagerProjectId = secretManagerProjectId;
+    this.secretName = secretName;
```
chingor13, line 40: These instance variables look unused - each other method is using their method arguments.

scripts/publish-cloud-run-secret-rotator.sh
```diff
@@ -0,0 +1,99 @@
...
+)
+
+echo "About to cloud run app ${serviceName}"
+gcloud beta run deploy "${serviceName}" "${deployArgs[@]}"
+
+echo "Adding ability for scheduler to execute the Function"
+gcloud run services add-iam-policy-binding "${serviceName}" \
+  --member="${serviceNameForInvoker}" \
+  --region "${region}" \
+  --role="roles/run.invoker"
```
chingor13, line 99: Consider managing who can invoke this Cloud Run service to terraform.

Also consider managing the concurrency, timeout, etc in terraform as well.

packages/secret-rotator/src/app.ts
```diff
@@ -0,0 +1,70 @@
...
+
+app.post('/', async (req: Request, res: Response) => {
```
chingor13, line 43: Consider using a more explicit endpoint than the root of the service.

packages/secret-rotator/src/app.ts
```diff
@@ -0,0 +1,70 @@
...
+  const serviceAccountProjectId = req.body.serviceAccountProjectId;
+  const serviceAccountEmail = req.body.serviceAccountEmail;
+  const secretManagerProjectId = req.body.secretManagerProjectId;
+  const secretName = req.body.secretName;
```
chingor13, line 47: Should we have some sort of validation on these user inputs?

chingor13 (changes requested):
Great start!

packages/secret-rotator/src/helper.ts
sofisl: hehehe... definitely... good catch!

packages/secret-rotator/cloudbuild.yaml
sofisl: Gotcha, so you're suggesting eliminating the build step in the cloudbuild.yaml file and adding it to the Dockerfile instead, right?

packages/secret-rotator/src/app.ts
sofisl: Done!

scripts/publish-cloud-run-secret-rotator.sh
sofisl: Done!

sofisl:
@chingor13, thanks for your comments! Ready for the second round :)

packages/secret-rotator/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# https://hub.docker.com/_/node
+FROM node:14-alpine
+
+# Create and change to the app directory.
+WORKDIR /usr/src/app
+
+COPY package*.json ./
+
+# Install production dependencies.
+RUN npm install
```
chingor13, line 25: Take another look at the release-trigger Dockerfile. The way this is implemented (single stage), the final image will contain the node devDependencies which are not necessary at runtime.

packages/secret-rotator/cloudbuild.yaml
```diff
@@ -0,0 +1,65 @@
+# Copyright 2020 Google LLC
```
chingor13, line 1:
```suggestion
# Copyright 2021 Google LLC
```

packages/secret-rotator/src/app.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  if (!serviceAccountProjectId) {
+    throw new Error('No service account project ID specified');
+  } else if (!serviceAccountEmail) {
+    throw new Error('No service account email specified');
+  } else if (!secretManagerProjectId) {
+    throw new Error('No secret manager project ID specified');
+  } else if (!secretName) {
+    throw new Error('No secret name specified');
```
chingor13, line 37: throwing an Error will cause the server to return a 500 error and likely print a stacktrace. Instead, we likely want to log an error/warning and return a 4xx status code with an error message.

packages/secret-rotator/src/secret-rotator.ts
```diff
@@ -0,0 +1,122 @@
...
+      secretName,
+      serviceAccountKey
+    );
+    logger.info(`updated secret: ${version}`);
+
+    await this.deleteExpiredServiceAccountKeys(
+      this.iamClient,
+      serviceAccountProjectId,
+      serviceAccountEmail
+    );
```
chingor13, line 120: Should we delete expired keys first? If we're at the max keys already, we may never get here if creating a service account key fails due to being at the limit.

packages/secret-rotator/test/test.app.ts
```diff
@@ -0,0 +1,137 @@
...
+        'test-service-account-email',
+        'test-secret-project-manager-Id',
+        'test-secret-name'
+      )
+    );
+    assert.deepStrictEqual(response.status, 200);
+  });
+
+  it('should throw an error if service account is falsy', async () => {
+    assert.rejects(async () => {
```
chingor13, line 66: I believe we need to await on `assert.rejects`.

Also, we should verify the type of error thrown (here and for each of the assert.rejects)

packages/secret-rotator/src/app.ts
```diff
@@ -0,0 +1,59 @@
...
+import express, {Request, Response} from 'express';
+import {GoogleAuth} from 'google-auth-library';
+import {SecretRotator} from './secret-rotator';
+import {iam} from '@googleapis/iam';
+import {SecretManagerServiceClient} from '@google-cloud/secret-manager';
+export const app = express();
+
+app.use(express.json());
+
+app.post('/rotate-secret', async (req: Request, res: Response) => {
```
chingor13, line 24: nit: name this `rotate-service-account-key`? The service is named secret-rotator, so we're always rotating a secret, but in the future, you could rotate some other kind of secret.

packages/secret-rotator/cloudbuild.yaml
```diff
@@ -0,0 +1,65 @@
...
+    waitFor: ["build-docker"]
+    args: ["push", "gcr.io/$PROJECT_ID/secret-rotator"]
+
+  - name: gcr.io/cloud-builders/gcloud
+    id: "deploy-cloud-run"
+    waitFor: ["push-docker"]
+    entrypoint: bash
+    args:
+      - "-e"
+      - "./scripts/publish-cloud-run-secret-rotator.sh"
```
chingor13, line 48: non blocking suggestion: since this script is not shared, remove unnecessary options (like bucket, key rings) and skip parameterizing the script (move the other values into the script).

chingor13 (changes requested):

packages/secret-rotator/Dockerfile
chingor13: This can be done in a follow-up PR if desired.

packages/secret-rotator/src/app.ts
chingor13: This could also be done in a follow-up PR.

sofisl: Done!

packages/secret-rotator/src/secret-rotator.ts
sofisl: Good catch.

packages/secret-rotator/cloudbuild.yaml
sofisl: I missed that from updating the publish cloud run script, thanks.

sofisl:
@chingor13, addressed all your comments, LMK if I got the Dockerfile right, that one is a bit new to me.

packages/secret-rotator/Dockerfile
sofisl: I think I addressed this... let me know if I missed something.

packages/secret-rotator/src/app.ts
```diff
@@ -17,24 +17,29 @@ import {GoogleAuth} from 'google-auth-library';
...
+    logger.error('No service account email specified');
+    res.sendStatus(400).send('No service account email specified');
   } else if (!secretManagerProjectId) {
-    throw new Error('No secret manager project ID specified');
+    logger.error('No secret manager project ID specified');
+    res.sendStatus(400).send('No secret manager project ID specified');
   } else if (!secretName) {
-    throw new Error('No secret name specified');
+    logger.error('No secret name specified');
+    res.sendStatus(400).send('No secret name specified');
```
chingor13: These should probably return from the function as well.

chingor13 (changes requested):
One last small fix

chingor13 (approved):

PR #2577, opened 2021-09-24
fix(owlbot): break loop if bot account added label

bcoe:
I worry that we still might end up with infinite loop if trusted contributor bot keeps adding `owlbot:run` and we're in a looping update situation.

### DIFF
--- a/packages/owl-bot/src/owl-bot.ts
+++ b/packages/owl-bot/src/owl-bot.ts
@@ -290,7 +290,7 @@ export async function handlePullRequestLabeled(
       defaultBranch: payload?.repository?.default_branch,
     },
     octokit,
-    false
+    isBotAccount(payload.sender.login) ? true : false
   );
   logger.metric('owlbot.run_post_processor');
 }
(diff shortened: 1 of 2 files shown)


packages/owl-bot/src/owl-bot.ts
```diff
@@ -290,7 +290,7 @@ export async function handlePullRequestLabeled(
       defaultBranch: payload?.repository?.default_branch,
     },
     octokit,
-    false
+    isBotAccount(payload.sender.login) ? true : false
```
chingor13, line 293: Do you need this ternary?

chingor13 (approved):

packages/owl-bot/src/owl-bot.ts
bcoe:
```suggestion
    isBotAccount(payload.sender.login)
```

bcoe: No, left over from original logic I was writing.

danielbankhead (approved):

PR #2578, opened 2021-09-24
feat(snippet-bot): allow api_shortname for samplegen samples

tmatsuo:
With this PR, snippet-bot will start allowing api_shortname as the
prefix for region tags generated by samplegen.


packages/snippet-bot/test/violations.ts
```diff
@@ -25,13 +29,74 @@ import {
...
+import {ApiLabels} from '../src/api-labels';
+import * as apiLabelsModule from '../src/api-labels';
 import {resolve} from 'path';
 import assert from 'assert';
 import {describe, it} from 'mocha';
 import * as sinon from 'sinon';
 
 const fixturesPath = resolve(__dirname, '../../test/fixtures');
 
+describe('checkProductPrefixViolations', async () => {
```
bcoe, line 41: no need for `async` here 👍 I think you can use `async`, if you do some async initialization:

```
desceribe('checkProductPrefixViolations', async () => {
  const sharedConfig = await loadConfig();
});
```

But I'd only do this if needed.

bcoe:
```suggestion
describe('checkProductPrefixViolations', () => {
```

bcoe (changes requested):
One small nit, LGTM otherwise.

packages/snippet-bot/test/violations.ts
tmatsuo: Sure, removed!

tmatsuo:
Sorry, I'll create a PR from the same repo, not from my fork

PR #2628, opened 2021-09-30
feat(do-not-merge): add configuration for `alwaysCreateStatusCheck`

### LINKED ISSUE #2446: Make it possible have "do not merge" a required check
I would like to be able to add the "do not merge" bot as a required check in repos so that if the label is there, then it blocks the PR from merging. But currently this is not possible because the check only provides a status if the label is there. If the label is not there, then the status hangs.

"do not merge" can be a required check.

Not having it as a required check is also fine. But we have accidentally merged in PRs before when we shouldn't have.

chingor13:
With this new configuration, you can enable `alwaysCreateStatusCheck` on your repo and then make that status check required.

This also allows the bot the create a status check for the config file format.

Fixes #2446


tmatsuo (approved):

tmatsuo:
It looks good.

Notes for future possible improvements. The bot can miss the webhook for the PR event thus there's no status check. It might be handy if there's a way for re-running the bot somehow instead of closing and reopening the PR.

PR #2633, opened 2021-10-02
fix: secret rotator deployment

sofisl:


packages/secret-rotator/cloudbuild-custom.yaml
```diff
@@ -22,13 +22,13 @@ steps:
       - "-f"
       - "Dockerfile"
       - "-t"
-      - "gcr.io/$PROJECT_ID/secret-rotator"
+      - "gcr.io/$_PROJECT_ID/secret-rotator"
```
chingor13, line 25: the un-underscored `$PROJECT_ID` is a built-in substitution. Do we actually want to make it a user substitution?

PR #2660, opened 2021-10-04
feat(policy): track issues for policy failures

JustinBeckwith:


packages/policy/src/issue.ts
```diff
@@ -0,0 +1,93 @@
...
+  } else {
+    if (!isValid) {
+      await octokit.issues.create({
+        title: '[Policy Bot] found one or more issues with this repository.',
+        owner: result.org,
+        repo: result.repo,
+        body: message,
+      });
+    }
+  }
```
chingor13, line 92: Note: we can probable extract this into a helper library if other bots find this maintain issue functionality helpful.

packages/policy/test/test.issue.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  it('should close an open issue if suddenly valid', async () => {
+    const scope = nock(githubHost)
+      .get(`/repos/${org}/${repo}/issues?state=open`)
+      .reply(200, [
+        {
+          title: '[Policy Bot] hello',
+        },
+      ])
+      .patch(`/repos/${org}/${repo}/issues/`)
```
chingor13, line 80: This response and update url should have an issue number, otherwise they don't look valid. This happens to pass because it's generating a url with a `undefined` issue number.

packages/policy/test/test.cli.ts
```diff
@@ -112,4 +113,25 @@ describe('cli', () => {
     await cli.main(m);
     assert.ok(exportStub.calledOnce);
   });
+
+  it('should attempt to file an issue if asked nicely', async () => {
+    process.env.GH_TOKEN = 'token';
```
chingor13, line 118: Will this cause weirdness with future tests?

chingor13 (changes requested):

packages/policy/test/test.cli.ts
JustinBeckwith: It doesn't, but just to be safe I came back around and cached the original value, and reset it after the tests run.

packages/policy/test/test.issue.ts
JustinBeckwith: Heh, you're right - this worked by accident.  Addressed!

chingor13 (approved):

PR #2664, opened 2021-10-04
fix: only trigger on label event if the label added was owlbot's

### LINKED ISSUE #2663: owlbot post processor running on label added too many times
GitHub sends one `pull_request.labeled` event for each label added to a pull request.

The bot currently looks at the list of labels that are present on the pull request rather than looking to see if the event is for the added label.

Example:
A person adds 4 labels, one of which is `owlbot:run`. We will trigger 4 Cloud Build jobs.

chingor13:
Fixes #2663


bcoe (approved):
Thanks for the fix 👌

PR #2665, opened 2021-10-05
fix: prevent post processor running infinite times

SurferJeffAtGoogle:
The problem is that "git commit --amend --no-edit" always replaces
the most recent commit even when no changes are pending.

### DIFF
--- a/packages/owl-bot/src/bin/commands/commit-post-processor-update.ts
+++ b/packages/owl-bot/src/bin/commands/commit-post-processor-update.ts
@@ -45,6 +45,12 @@ export async function commitPostProcessorUpdate(repoDir = ''): Promise<void> {
   }
   // Add all pending changes to the commit.
   cmd('git add -A .', {cwd: repoDir});
+  const status = cmd('git status --porcelain', {cwd: repoDir}).toString(
+    'utf-8'
+  );
+  if (!status) {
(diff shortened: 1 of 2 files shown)


SurferJeffAtGoogle:
cc @amanda-tarafa @jskeet

jskeet (approved):
Thanks!

bcoe (approved):

packages/owl-bot/src/bin/commands/commit-post-processor-update.ts
```diff
@@ -45,6 +45,12 @@ export async function commitPostProcessorUpdate(repoDir = ''): Promise<void> {
   }
   // Add all pending changes to the commit.
   cmd('git add -A .', {cwd: repoDir});
+  const status = cmd('git status --porcelain', {cwd: repoDir}).toString(
```
bcoe, line 48: might be worth adding a comment that `// git status --porcelain returns empty string if no local changes`.

I'm going ahead and merging, since it seems like this bug would be pretty serious for `.NET` folks.

SurferJeffAtGoogle: Thank you for merging.  Comment added in https://github.com/googleapis/repo-automation-bots/pull/2666.

PR #2737, opened 2021-10-18
save for now

sofisl:

### DIFF
--- a/packages/merge-on-green/src/merge-on-green.ts
+++ b/packages/merge-on-green/src/merge-on-green.ts
@@ -32,6 +32,7 @@ const TABLE = 'mog-prs';
 const datastore = new Datastore();
 const MAX_TEST_TIME = 1000 * 60 * 60 * 6; // 6 hr.
 const WORKER_SIZE = 4;
+const MAX_ENTRIES = 100;
 
 handler.allowlist = [
   'googleapis',
@@ -44,7 +45,7 @@ handler.allowlist = [
   'firebase',
 ];
 
-interface DatastorePR {
+export interface DatastorePR {
   number: number;
   repo: string;
   owner: string;
@@ -55,6 +56,7 @@ interface DatastorePR {
   url: string;
   reactionId: number;
   installationId?: number;
+  created: string;
 }
 
 interface IncomingPR {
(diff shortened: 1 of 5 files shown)


packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -32,6 +32,7 @@ const TABLE = 'mog-prs';
 const datastore = new Datastore();
 const MAX_TEST_TIME = 1000 * 60 * 60 * 6; // 6 hr.
 const WORKER_SIZE = 4;
+const MAX_ENTRIES = 100;
```
bcoe, line 35: I think `100` is potentially enough entries that it would start to hit some API limits, I would be tempted to pick a threshold like `25`.

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -73,6 +75,28 @@ interface Label {
   name: string;
 }
 
+/**
+ * Reduces the list of PRs to search by half if there are more than 100 entries at a time
+ * @returns an array of Datastore PRs
+ */
+handler.maybeReducePRList = function maybeReducePRList(prs: DatastorePR[]) {
+  if (prs.length >= MAX_ENTRIES) {
+    if (Date.now() % 2 === 0) {
```
bcoe, line 84: Rather than hard-coding `2`, which I think will quickly bump into scaling issues again as we continue to scale, we'd do better to instead pick a work size, like `25`, and then use this as our modulus:

```
const modulus = parseInt(prs.length / MAX_ENTRIES)
prs = prs.filter(x => new Date(x.created!).getTime() % modulus === 0);
logger.info(
    `Too many entries in Datastore table; examining first ${prs.length}`
);
```

packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -73,6 +75,28 @@ interface Label {
...
+      prs = prs.filter(x => new Date(x.created!).getTime() % 2 === 0);
+      logger.info(
+        `Too many entries in Datastore table; examining first ${prs.length}`
+      );
+    } else {
```
bcoe, line 89: No need for the `else`, the modulus step will grab the appropriate sized chunk of work, time is essentially just being treated as a random value for us.

PR #2790, opened 2021-10-26
chore(refactor): break up languages into classes for easier implementation of new use cases by languge

sofisl:
Per our conversation, refactoring auto-approve into classes by language to make new use cases easier to implement by language.


packages/auto-approve/src/language-rules/java.ts
```diff
@@ -0,0 +1,117 @@
...
+           <artifactId>google-api-services-policytroubleshooter</artifactId>
+    -      <version>v1-rev20210319-1.31.5</version>
+    +      <version>v1-rev20210319-1.32.1</version>
+        */
+    newVersion:
+      /<groupId>([^<]*)<\/groupId>[\s]*<artifactId>([^<]*)<\/artifactId>[\s]*-[\s]*<version>(v[0-9]-rev[0-9]*-[0-9]*\.[0-9]*\.[0-9]|[[0-9]*\.[0-9]*\.[0-9]*)<\/version>[\s]*\+[\s]*<version>(v[0-9]-rev[0-9]*-([0-9]*)\.([0-9]*\.[0-9])|([0-9]*)\.([0-9]*\.[0-9]*))<\/version>/,
+  },
+];
+
+export class Rules {
```
chingor13, line 60: Rather than each language creating a `Rules` class. We should have a base `Rules` interface with something like:

```
interface Rules {
  checkPR(): Promise<boolean>;
}
```

Each language would implement the `Rules` interface. Then the calling code wouldn't care about any implementation details of where the rules came from -- all language rules would be treated the same way.

chingor13 (changes requested):
As implemented, this looks like duck-typing rather than using interfaces. Since we're using typescript, we should leverage true typing.

sofisl:
@chingor13, thanks for the review, PTAL

sofisl:
Closing in lieu of #2949

Issue #2796, opened 2021-10-30
bug(owl-bot): owlbot tries to run on archived repo 


parthea:
https://github.com/googleapis/python-identity-toolkit is archived (read-only). Owlbot shouldn't run on repos that are read-only.

See [build log](https://pantheon.corp.google.com/cloud-build/builds;region=global/ad428821-3625-418f-9254-8345081556c1?project=repo-automation-bots).

parthea:
Also see https://pantheon.corp.google.com/cloud-build/builds;region=global/745ceeda-98e8-4d6f-b032-166c757ec2cb;step=11?project=repo-automation-bots

SurferJeffAtGoogle:
I haven't figured out why this build was triggered.  The corresponding branch doesn't exist and I find no pull request that would have triggered it.

There is logic in Owl Bot to not create pull requests on archived repos.

PR #2827, opened 2021-11-01
fix: owl bot endlessly retried failed owl-bot-lock builds

SurferJeffAtGoogle:
Fixes https://github.com/googleapis/repo-automation-bots/issues/2796

### DIFF
--- a/packages/owl-bot/src/scan-and-retry-failed-lock-updates.ts
+++ b/packages/owl-bot/src/scan-and-retry-failed-lock-updates.ts
@@ -20,6 +20,8 @@ const twentyFourHours =
   60 * // minutes
   24; // hours.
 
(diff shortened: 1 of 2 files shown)


packages/owl-bot/src/scan-and-retry-failed-lock-updates.ts
```diff
@@ -20,6 +20,8 @@ const twentyFourHours =
   60 * // minutes
   24; // hours.
 
+const seventyTwoHours = 3 * twentyFourHours;
```
chingor13, line 23: nit: I'd probably have named this constant what it's used for rather than the value with a comment describing the human readable value.

chingor13 (approved):

Issue #2792, opened 2021-10-28
bug(owl-bot): New postprocessor publication of owlbot-php doesn't trigger a pull request for lock update


tmatsuo:
The target repo:
https://github.com/googleapis/google-cloud-php

The lock file:
https://github.com/googleapis/google-cloud-php/blob/master/.github/.OwlBot.lock.yaml

When I modify the docker tag in the container registry, I see the following in OwlBot log:

```
New docker image published: gcr.io/cloud-devrel-public-resources/owlbot-php:latest.  Digest: sha256:07ba38b7dbfeedc36833bc7334348681fdea48c1cec8d129c24fc4eab75edbfc"
```

However, I still don't see any pull requests for updating the lock file.

tmatsuo:
I took a look at the Firestore database and noticed there is no `lock` element in the data. This is why the pubsub message doesn't trigger the cloud build.

```
branchName: "master"
(string) 
commitHash: "386172a6d3ff625e20354befd2892cf6feecd5ab"
(string) 
installationId: 14695777
(number) 
yamls
(array) 
0
(map) 
path: "Asset/.OwlBot.yaml"
(string) 
yaml
(map) 
deep-copy-regex
(array) 
0
(map) 
dest: "/owl-bot-staging/Asset/v1/$1"
(string) 
source: "/google/cloud/asset/v1/.*-php/(.*)"
```

PR #2828, opened 2021-11-01
fix: fetch repo contents via git clone instead of downloading a zip

SurferJeffAtGoogle:
The problem is, .githubattributes can prevent files from being packed into a .zip file, like this:
https://github.com/googleapis/google-cloud-php/blob/393c614f9cff7f0fb9a0bbc870735f784a8ef293/.gitattributes#L13

Fixes https://github.com/googleapis/repo-automation-bots/issues/2792


tmatsuo (commented):
The code looks good, but I see the test is failing.

tmatsuo (approved):

Issue #2863, opened 2021-11-09
OwlBot builds are failing to "JavaScript heap out of memory"


tmatsuo:
Many OwlBot's cloud builds are failing today.

An example build:
https://pantheon.corp.google.com/cloud-build/builds;region=global/aa1591fa-c561-421c-88ea-a4da1f43d9e9;step=3?project=repo-automation-bots

Log:

```
<--- Last few GCs --->

[1:0x7f9a720262c0]  2562673 ms: Mark-sweep (reduce) 1967.1 (1971.0) -> 1965.7 (1971.9) MB, 2036.3 / 0.0 ms  (+ 122.4 ms in 136 steps since start of marking, biggest step 8.9 ms, walltime since start of marking 2337 ms) (average mu = 0.481, current mu = 0.[1:0x7f9a720262c0]  2564998 ms: Mark-sweep (reduce) 1966.7 (1973.9) -> 1966.7 (1974.7) MB, 2248.7 / 0.0 ms  (+ 69.5 ms in 6 steps since start of marking, biggest step 31.1 ms, walltime since start of marking 2322 ms) (average mu = 0.298, current mu = 0.00

<--- JS stacktrace --->

FATAL ERROR: MarkCompactCollector: young object promotion failed Allocation failed - JavaScript heap out of memory
```

bcoe:
We are running the CLI with Node's default heap allocatoin, which I believe is 2Gb. We can increase this with:

We can try to increase this with:

```
  - name: 'gcr.io/repo-automation-bots/owlbot-cli'
    args:
      - scan-googleapis-gen-and-create-pull-requests
      - --pem-path
      - /workspace/private-key.pem
      - --app-id
      - ${_APP_ID}
      - --installation
      - ${_INSTALLATION}
      - --firestore-project
      - ${_FIRESTORE_PROJECT}
  env:
    - NODE_OPTIONS='--max-old-space-size=8192'
```

If this doesn't work, we may need to update docker container setting this env or flag.

PR #2865, opened 2021-11-09
fix: reduce memory consumed by issue cache

SurferJeffAtGoogle:
Fixes https://github.com/googleapis/repo-automation-bots/issues/2863

### DIFF
--- a/packages/owl-bot/test/copy-code.ts
+++ b/packages/owl-bot/test/copy-code.ts
@@ -21,6 +21,7 @@ import {
   copyTagFrom,
   findCopyTag,
   findSourceHash,
+  newRepoHistoryCache,
   sourceLinkFrom,
   stat,
   unpackCopyTag,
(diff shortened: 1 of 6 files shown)


tmatsuo (approved):

packages/owl-bot/test/copy-code.ts
```diff
@@ -69,6 +70,44 @@ Copy-Tag: ${copyTag}
...
+        octokit as unknown as OctokitType,
+        destRepo,
+        'abc123',
+        1000,
+        cache
+      )
+    );
+
+    // Confirm its in the cache.
+    assert.strictEqual(
```
tmatsuo, line 99: These asserts look similar, do we need both?

SurferJeffAtGoogle: The first assert is optional.  If it were to fail, it be more obvious what had failed, so I left it in.

Issue #2867, opened 2021-11-09
bug(owl-bot): rmDirSync failing in owlbot-scan-googleapis-gen-and-create-pull-requests


tmatsuo:
[Example build](https://pantheon.corp.google.com/cloud-build/builds;region=global/16caa941-6eaf-4f7d-b5ef-9e80110e51bb?project=repo-automation-bots)

I saw many errors like this:
```
Step #3: Error: ENOTEMPTY: directory not empty, rmdir '/tmp/tmp-1-qJpzK6jjGDva/dest/google-cloud-shell/src/main/java/com/google/cloud'
Step #3:     at Object.rmdirSync (fs.js:916:10)
Step #3:     at copyDirs (/usr/src/app/build/src/copy-code.js:374:16)
Step #3:     at copyCode (/usr/src/app/build/src/copy-code.js:301:5)
Step #3:     at copyCodeIntoLocalBranch (/usr/src/app/build/src/copy-code.js:174:55)
Step #3:     at async copyCodeAndCreatePullRequest (/usr/src/app/build/src/copy-code.js:194:18)
Step #3:     at async scanGoogleapisGenAndCreatePullRequests (/usr/src/app/build/src/scan-googleapis-gen-and-create-pull-requests.js:113:9)
```

SurferJeffAtGoogle:
I'll prepare a PR to improve the error message.

SurferJeffAtGoogle:
Actually, I think I see the issue.  Inner directories must be removed before parent directories.  I'll fix the code to sort the directories before trying to remove them.

PR #2874, opened 2021-11-12
fix: correctly remove nested directories

SurferJeffAtGoogle:
Fixes https://github.com/googleapis/repo-automation-bots/issues/2867


bcoe (approved):
sneaky, I like this solution.

Issue #2751, opened 2021-10-22
feat: auto-label should mark stale pull requests


losalex:
**Is your feature request related to a problem? Please describe.**
Our team deal with lots of repos containing high volume of pull requests (PRs). In order to prioritize some PRs over others we need a staleness indicator, thus older requests can be looked first. 
 **Describe the solution you'd like**
Older PRs should be labeled when PRs last modification date exceeds 2 weeks ago from today. Ideally the labeling rule should be configurable and could have multiple, timespan-based staleness indicators.
 **Describe alternatives you've considered**
There is [Probot Stale](https://github.com/probot/stale) bot, however this would be yet another bot to maintain and auto-label is already used with our repos.
 **Additional context**
There is related feature request [#2303](https://github.com/googleapis/repo-automation-bots/issues/2303) which could be a plus

PR #2906, opened 2021-11-17
feat(auto-label): support for labeling stale pull requests

losalex:
Runs on `'schedule.repository'` event through all pull requests in repo and updates stale PRs with "stale: old" or "stale: critical" labels respectively based on configuration

Fixes [#2751](https://github.com/googleapis/repo-automation-bots/issues/2751) 🦕


packages/auto-label/src/auto-label.ts
```diff
@@ -249,15 +245,102 @@ handler.autoLabelOnPR = async function autoLabelOnPR(
...
+  owner: string,
+  repo: string,
+  config: Config
+) {
+  // If user has turned on stale labels by configuring {staleness: {pullrequest: true, old: 60, critical: 120}}
+  // By default, this feature is turned off
+  if (!config.staleness?.pullrequest) {
+    return;
+  }
+  const response = await context.octokit.rest.pulls.list({
```
bcoe, line 283: By derfault, this will only return the first 25 pull requests, a better approach could be to use an async itetertor, like this:

```js
for await (const response of octokit.paginate.iterator(
  octokit.rest.issues.listForRepo,
  {
    owner: "octokit",
    repo: "rest.js",
  }
)) {
  // do whatever you want with each response, break out of the loop, etc.
}
```

Then you will automatically page through all pull requests.

packages/auto-label/src/auto-label.ts
```diff
@@ -249,15 +245,102 @@ handler.autoLabelOnPR = async function autoLabelOnPR(
...
+    return;
+  }
+
+  let old = config.staleness?.old || helper.DEFAULT_DAYS_TO_STALE;
+  let critical = config.staleness?.critical || helper.DEFAULT_DAYS_TO_STALE * 2;
+  let label = null;
+
+  // Make sure that config is right and if not, set defaults
+  if (old > critical || critical > helper.MAX_DAYS || old < 0) {
+    logger.info(
```
bcoe, line 298: I would be tempted to simply log an error here, and `return`, as it seems presumtuous to try to fix the value.

Then you can make `critical` and `old`, `const`, and it reads a bit better.

packages/auto-label/test/auto-label.ts
```diff
@@ -567,6 +567,144 @@ describe('auto-label', () => {
       });
       ghRequests.done();
     });
+
+    it('does not updates staleness labels when same', async () => {
```
bcoe, line 571: Tests are looking great, I think it would be worth adding one additional assertion that, kif `staleness` is not set in config, labels are not added to a PR that would otherwise have them added with the default settings.

bcoe (changes requested):
Left a few nits.

All and all this is looking really solid, thanks for the contribution.

packages/auto-label/src/auto-label.ts
```diff
@@ -249,15 +245,102 @@ handler.autoLabelOnPR = async function autoLabelOnPR(
...
+        );
+        context.octokit.issues.removeLabel({
+          owner,
+          repo,
+          issue_number: pull.number,
+          name: staleLabel.name!,
+        });
+      }
+
+      logger.info(`Adding ${label} to ${owner}/${repo}/${pull.number}...`);
```
bcoe, line 334: It would be interesting to emit a metric here, so that we can make a dashboard of stale PRs over time, here's an example from merge-on-green:

```js
    logger.metric('merge_on_green.mergeability_sample', {
      repo: `${owner}/${repo}/`,
      number: pr,
      mergeable: prInfo.mergeable,
      mergeable_state: prInfo.mergeable_state,
      checkReviews,
      checkStatus,
    });
```

packages/auto-label/src/auto-label.ts
losalex: Yeah, I was thinking to isolate a call to pull requests only since correct me if I wrong, but call to `octokit.rest.issues.listForRepo` will return all issues in repo, including pull requests. We have already call to `context.octokit.issues.listForRepo.endpoint.merge` inside the `'schedule.repository'` event - I can patch it to process 5 issues counted by `labelWasNotAddedCount` and then only process pull requests when `labelWasNotAddedCount > 5` - what do you think?

packages/auto-label/src/auto-label.ts
losalex: Sure, will fix it

packages/auto-label/src/auto-label.ts
losalex: yep, makes sense

packages/auto-label/src/auto-label.ts
bcoe: I don't think we need to optimize yet, we have `12,000` events an hour to work with, and we're only going to be enabling this on a few repos to start.

I would keep processing PRs in your own helper, and just switch to an async iterator, so you process them all.

losalex: OK, thanks

packages/auto-label/test/auto-label.ts
losalex: thanks - added tests

packages/auto-label/test/language-path-tests.ts
```diff
@@ -94,13 +94,6 @@ describe('language-and-path-labeling', () => {
         .get('/repos/testOwner/testRepo/pulls/12/files')
         .reply(200, pr_files_payload)
 
-        // Mock issues.addlabels adding path_label
-        .post('/repos/testOwner/testRepo/issues/12/labels', body => {
```
bcoe, line 98: Do you know why this assertion ended up needing to be dropped, this concerns me slightly?

bcoe (changes requested):
Everything is looking good to me, thank you for addressing my code review.

One slight issue that concerns me is that we've had to remove one mock from a prior test, is this expected?

bcoe (approved):

packages/auto-label/test/language-path-tests.ts
losalex: Yes, sorry about this - there are 2 redundant calls to `context.octokit.issues.addLabels` within same function, I just united those into single call... It might be confusing, I just get used to approach to fix things once you encounter them.

losalex:
Thanks!

PR #2914, opened 2021-11-18
feat(auto-approve): add toggle RELEASE_FREEZE

bcoe:
Add environment variable toggle for release freezes.


packages/auto-approve/src/auto-approve.ts
```diff
@@ -142,6 +142,14 @@ export function handler(app: Probot) {
       'pull_request.synchronize',
     ],
     async (context: Context<'pull_request'>) => {
+      // During codefreeze, simply set the RELEASE_FREEZE environment variable.
+      if (process.env.RELEASE_FREEZE) {
```
chingor13, line 146: How are we planning to set this env variable on deploy?

chingor13: Also do we only want to clock auto approve for release PRs (not dependency updates)?

chingor13 (changes requested):

packages/auto-approve/src/auto-approve.ts
bcoe: > Also do we only want to clock auto approve for release PRs (not dependency updates)?

This would be the ideal, but I think it's hard to distinguish today since the PRs we merge are based on title, committer, etc., not type of PR.

This will be easier in the future with some lovely refactoring @sofisl has planned.

bcoe:
> How are we planning to set this env variable on deploy?

@chingor13 in the immediate term, just set the environment variable manually, I don't believe that environment variables are removed on deployment, just updated.

packages/auto-approve/src/auto-approve.ts
```diff
@@ -149,6 +149,18 @@ export function handler(app: Probot) {
       const repo = pr.pull_request.base.repo.name;
       const prNumber = pr.number;
 
+      // During codefreeze, simply set the RELEASE_FREEZE environment variable.
+      // if a PR is from release-please, it will not be merged:
+      if (
+        process.env.RELEASE_FREEZE &&
```
chingor13, line 155: I would have this explicitly check `=== "true"` otherwise, if set to "false" it would probably pass.

chingor13 (changes requested):

packages/auto-approve/src/auto-approve.ts
bcoe: I tend to prefer just treating a non empty string as `true`, since people might expect to be able to set `0`/`1` too, at which point yo're writing a bunch of logic for boolean parsing.

But happy to change if this is a blocker for you.

bcoe:
```suggestion
        process.env.RELEASE_FREEZE === 'true' &&
```

chingor13 (approved):

sofisl:
I think, in the future, I'd maybe just hard code the potential release-freezes,since I think that will be easier than people remembering to add the env variable. But, I think it's a great fix since the freeze happens tomorrow, thanks for doing this Ben!

sofisl (approved):

parthea:
@sofisl The freeze for client libraries has already started.

PR #2943, opened 2021-11-22
feat(bot-config-utils): getConfigWithDefault now returns default on permission errors

### LINKED ISSUE #2739: do-not-merge is failing with "Resource not accessible by integration" on private repo
`do-not-merge` has been installed on a private repo and is receiving the error:

```
Resource not accessible by integration 
```

This is resulting in a backup in the task queue.

bcoe:
I believe what might be causing #2739 is a lack of repository level config,
which causes the automation to check the org level .github repository, which
the integration does not have permission to access.

Fixes #2739

### DIFF
--- a/packages/bot-config-utils/README.md
+++ b/packages/bot-config-utils/README.md
@@ -35,6 +35,8 @@ const config = await getConfig<Config>(
 
 You can use a similar method that supports default value.
 
+> Note: on permission errors `getConfigWithDefault` will return the default configuration.
+
 ```typescript
 import {
   getConfigWithDefault,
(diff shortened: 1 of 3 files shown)


tmatsuo (approved):

tmatsuo (changes requested):
Ah wait.

Ignoring 404 makes sense, but 403 should be caught because it is a configuration error.

tmatsuo:
Ok, if the bot is installed to a single repo, it doesn't have permission?

I'm still not sure if it is the right thing to ignore 403

bcoe:
@tmatsuo if the automation is installed on a single repository and we get a `404`, it seems like we should fallback to the default configuration right?

So it seems reasonable to only accept 404 erros when grabbing config from the repository itself, but when fetching config from `.github/foo.yaml`, shouldn't we ignore 403s, in case we do not have permissoin to read this repo?

PR #2947, opened 2021-11-23
fix(bot-config-util): decrease pagination size from 100 to 50, move to async iterator

### LINKED ISSUE #1949: Bug(some bots): Seems like some bots are dying with a big PR
An example:
https://github.com/googleapis/google-api-php-client-services/pull/363

- [ ] Function timeout while determining whether the config file was modified (whether or not to validate)
- [ ] GitHub API request timeout while paginating through pull request files

bcoe:
Reduces the number of items returned per operation. Moves to an async iterator,
in the hopes that this also helps take the load off GitHub API.

Fixes #1949 

----

This PR does two things in an effort to reduce the load on GitHub API.

1. reduces items from 100 to 50 (we can experiment with this # more over time). It will now take 60 API requests to pull in all results for a massive PR, but I think it's worth trying 60 smaller requests, vs., 30 larger requests, in an effort to circumvent timeout.
2. I've switched to using async iteration. This means that a chunk of data will be downloaded and then processed, before the next API request goes out to GitHub (_whereas the paginator would have made all requests sequentially_). It's my hope that this approach also takes some load off of the GitHub API.

I suggest we try these changes, and then see if this type of error crops up again in the future.


packages/bot-config-utils/src/bot-config-utils.ts
```diff
@@ -185,57 +185,62 @@ export class ConfigChecker<ConfigType> {
         owner: owner,
         repo: repo,
         pull_number: prNumber,
-        per_page: 100,
+        per_page: 50, // Currently 30 is GitHub's default.
```
tmatsuo, line 188: I don't think it's concurrency. I think it's the number of items per page.

bcoe: That's reasonable, this reduces the concurrency to 50 from 100.

I think that using an async iterator is still a good call, vs., auto pagination, as it can help take a load off an API (_we've been moving away from using auto-pagination for client libraries in favor of async iterators._).

tmatsuo: Can you update the title and the description accordingly?

sofisl (approved):

tmatsuo:
Title looks better. The description (the first comment) still looks incorrect.

Sorry to be picky, but this kind of incorrect description will confuse future readers.

tmatsuo:
I don't think "concurrency" is the right word.

That number specifies the number of items returned in a single api call.

tmatsuo (approved):

PR #2948, opened 2021-11-23
fix(policy): bubble error when fetching files

### LINKED ISSUE #1924: bug(policy-bot): created a PR adding a file that already exists
An example:
https://github.com/googleapis/go-genproto/pull/610

bcoe:
If there's an intermittent error loading config files, prior to this change this
was treated as a 404. Rather than doing this, bubble the error. This will allow
us to investigate how common of an occurrence this is.

Fixes #1924


tmatsuo:
Is it only for config file?

tmatsuo (approved):

Issue #2094, opened 2021-06-17
owlbot: add image creation datetime to .OwlBot.lock.yaml


chingor13:
If an owlbot post-processor image is updated often, it's possible that we merge them out of order and will get merge conflicts.

When resolving, it's difficult to compare to hashes to decide which is the correct one. It would be nice to add a comment with the docker image build time so we can easily pick the later image SHA

chingor13:
related #2207

PR #2978, opened 2021-11-30
feat: record docker image's creation timestamp in .OwlBot.lock.yaml

SurferJeffAtGoogle:
Fixes https://github.com/googleapis/repo-automation-bots/issues/2094


packages/owl-bot/test/docker-api.ts
```diff
@@ -0,0 +1,28 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {describe, it} from 'mocha';
+import * as docker from '../src/docker-api';
+import * as assert from 'assert';
+
+describe.skip('fetchConfig', () => {
```
JustinBeckwith, line 19: Checking in a test with a `.skip` don't feel quite right.  Do we really want a test that performs a network operation?  Could we use nock to return a mocked response here?

SurferJeffAtGoogle: Yes, I want a test that performs a network operation.  I'm implementing one half of a protocol.

If I were to drop the integration test and write a test with nock, then I'm very unlikely to catch a bug.  The test would only confirm that my (mis?)understanding of the client agrees with my (mis?)understanding of the server.  It would not confirm that either agrees with the defined protocol.

So yes, I want check-in this integration test, so that if I change the code later, I can remove the `.skip` and at least confirm that it works on my machine.

I am open to adding another test, a unit test that uses nock and is not skipped.  Would you like to see that?

chingor13: Can we create a separate integration-test folder and add a separate `npm run` command to run the integration tests?

SurferJeffAtGoogle: Done.

Ready for another look.

chingor13 (approved):

PR #2994, opened 2021-12-06
chore: Upgrade git in the Bazel Bot Dockerfile

jskeet:
I'm going to build this locally in a Dockerfile before submitting.


jskeet:
Humbug - no, that doesn't work because 2.20.1 is the latest version of git in Debian buster.

Will need to reconsider...

chingor13:
FYI: I opened https://github.com/googleapis/repo-automation-bots/issues/2995 so we can more easily test changes to this file.

PR #3001, opened 2021-12-06
fix: improve validation error messages

### LINKED ISSUE #2992: OwlBot not providing sufficient context to debug bad config
OwlBot is not providing sufficient context to debug config file failure

Refs: https://github.com/googleapis/java-pubsublite/issues/959

chingor13:
Improves collection of validation error messages for `.OwlBot.yaml`. We now collect them and throw a better error message. In the created issue, we now list all the collected error messages.

Fixes #2992


packages/owl-bot/src/fetch-configs.ts
```diff
@@ -41,6 +41,6 @@ export async function fetchConfigs(
     return collectConfigs(repoDir);
   } finally {
     // When running in cloud functions, space is limited, so clean up.
-    fs.rmdirSync(tmpDir, {recursive: true});
+    fs.rmSync(tmpDir, {recursive: true});
```
bcoe, line 44: one slight behaviorial change is that this will now throw if the path doesn't exist, this is probably fine, if there's a reason why we expect the folder not to exist we can add the `force` flag.

chingor13:
This includes the same fix that broke when we were running on node12 in prod.

chingor13:
> This includes the same fix that broke when we were running on node12 in prod.

That was fixed yesterday.

packages/owl-bot/src/config-files.ts
```diff
@@ -67,45 +67,89 @@ export interface OwlBotYaml {
 // The default path where .OwlBot.yaml is expected to be found.
 export const DEFAULT_OWL_BOT_YAML_PATH = '.github/.OwlBot.yaml';
 
-function validatePath(path: string, fieldName: string) {
-  if (path && path[0] !== '/') {
-    throw `I expected the first character of ${fieldName} to be a '/'.  Instead, I found ${path}.`;
+function validatePath(
+  path: string,
+  fieldName: string,
+  errorMessages: string[]
```
SurferJeffAtGoogle, line 73: I see how this code reports all the issues in a config file, but I don't see how it transforms the error message `[object Object]` into something more informative.

If line 72 were the error being thrown, then we would have seen the error message, not `[object Object]`.

SurferJeffAtGoogle (changes requested):

packages/owl-bot/src/config-files.ts
chingor13: This wasn't the `[object Object]` error message. The `throw validate.errors` from old line 101 was the `[object Object]` message (it was `ErrorObject<string, Record<string, any>, unknown>` from `ajv`). After this change, we collect better error messages and put into the `InvalidOwlBotConfigError` that is thrown for a better message (new lines 114-120).

We change this helper to add error messages to a list to make types easier (collecting error messages than collecting various Error objects).

packages/owl-bot/src/config-files.ts
```diff
@@ -67,45 +67,89 @@ export interface OwlBotYaml {
...
+ * @returns {OwlBotYaml} The parsed OwlBot config
+ * @throws {InvalidOwlBotConfigError} If there are parse or validation errors.
+ */
+export function owlBotYamlFromText(yamlText: string): OwlBotYaml {
+  const loaded = yaml.load(yamlText) ?? {};
+  const validate = new Ajv().compile<OwlBotYaml>(owlBotYamlSchema);
+  const errorMessages: string[] = [];
+  if (!validate(loaded)) {
+    for (const err of validate.errors as DefinedError[]) {
+      const message = err.message
```
SurferJeffAtGoogle, line 115: To alleviate my paranoia, would you mind making this
```
  const message = err?.message
```

packages/owl-bot/src/config-files.ts
```diff
@@ -67,45 +67,89 @@ export interface OwlBotYaml {
...
-      validatePath(deepCopy.dest, 'dest');
-      validatePath(deepCopy.source, 'source');
-      // Confirm it's a valid regular expression.
-      toFrontMatchRegExp(deepCopy.source);
-    }
-    for (const removePath of yaml['deep-remove-regex'] ?? []) {
-      validatePath(removePath, 'deep-remove-regex');
-      // Confirm it's a valid regular expression.
-      toFrontMatchRegExp(removePath);
+function validateRegex(regex: string, errorMessages: string[]) {
```
SurferJeffAtGoogle, line 82: Please add a /** */ comment for this function.

packages/owl-bot/src/config-files.ts
```diff
@@ -67,45 +67,89 @@ export interface OwlBotYaml {
 // The default path where .OwlBot.yaml is expected to be found.
 export const DEFAULT_OWL_BOT_YAML_PATH = '.github/.OwlBot.yaml';
 
-function validatePath(path: string, fieldName: string) {
-  if (path && path[0] !== '/') {
-    throw `I expected the first character of ${fieldName} to be a '/'.  Instead, I found ${path}.`;
+function validatePath(
```
SurferJeffAtGoogle, line 70: Please add a /** */ comment for this function.

SurferJeffAtGoogle (changes requested):

packages/owl-bot/src/config-files.ts
chingor13: Done

packages/owl-bot/src/config-files.ts
chingor13: Done

packages/owl-bot/src/config-files.ts
chingor13: Done

sofisl (approved):

SurferJeffAtGoogle (approved):

PR #3019, opened 2021-12-10
feat(repo-metadata-lint): wire up logging-only version of linter

bcoe:
Wires together the moving parts of the linter, but currently only logs to console.

### DIFF
--- a/packages/repo-metadata-lint/src/repo-metadata-lint.ts
+++ b/packages/repo-metadata-lint/src/repo-metadata-lint.ts
@@ -15,8 +15,8 @@
 /* eslint-disable-next-line node/no-extraneous-import */
 import {Probot} from 'probot';
 import {logger} from 'gcf-utils';
-
-// type IssueResponse = Endpoints['GET /repos/{owner}/{repo}/issues']['response'];
+import * as fileIterator from './file-iterator';
+import {Validate, ValidationResult} from './validate';
 
 /**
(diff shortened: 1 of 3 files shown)


packages/repo-metadata-lint/src/repo-metadata-lint.ts
```diff
@@ -26,8 +26,27 @@ export function handler(app: Probot) {
...
-    logger.info('schedule endpoint', owner);
+    const repo = context.payload.repository.name;
+    const iterator = new fileIterator.FileIterator(
+      owner,
+      repo,
+      context.octokit
+    );
+    const results: Array<ValidationResult> = [];
+    const validate = new Validate(context.octokit);
+    for await (const [path, metadata] of iterator.repoMetadata()) {
```
chingor13, line 37: nit: an async iterator certainly works here, but each file validation is an independent event and there's no need to stop iteration until we've validated every file. Instead, we could do a basic loop, collect promises, and `await Promise.all([promises])`. They are also more of a pain to stub for testing.

chingor13 (approved):

packages/repo-metadata-lint/src/repo-metadata-lint.ts
bcoe: Yeah I agree, okay if I refactor when I write some more tests, just want to test this end to end.

Issue #2753, opened 2021-10-22
feat: auto-label should mark pull requests with T-shirt size indicator


losalex:
**Is your feature request related to a problem? Please describe.**
Our team deal with lots of repos containing high volume of pull requests (PRs). In order to prioritize some PRs over others we need a PR size indicator, thus we can estimate amount of work required to invest in review
 **Describe the solution you'd like**
Every PR should be labeled by auto-label bot with T-shirt label (e.g. S/M/L/XL) based on calculated size. See [The Commit Size Distribution of Open Source Software](https://dirkriehle.com/wp-content/uploads/2008/09/hicss-42-csdistr-final-web.pdf) for more insights
 **Describe alternatives you've considered**
There is an existing bot [pull-request-size](https://github.com/noqcks/pull-request-size), however this would be yet another bot to maintain and auto-label is already used with our repos.

PR #3029, opened 2021-12-14
feat: auto-label should mark pull requests with T-shirt size indicator

losalex:
This is a feature to label pull requests with t-shirt sizing. The feature calculates pull request size based on amount of changes provided by GitHub and translates calculated size into t-shirt size mapping (see README.md for more info).

Fixes [#2753](https://github.com/googleapis/repo-automation-bots/issues/2753) 🦕


packages/label-sync/src/labels.json
```diff
@@ -244,6 +244,36 @@
       "name": "stale: extraold",
       "description": "Pull request is critically old and needs prioritization.",
       "color": "801904"
+    },
+    {
```
chingor13, line 248: Do we want to have the labels sync'ed everywhere? This will sync on more repos than auto-label is installed/enabled on.

An alternative is to use `label-utils` (https://github.com/googleapis/repo-automation-bots/tree/main/packages/label-utils#usage-1) in a repository cron which would also allow conditionally adding the labels iff that feature is enabled.

packages/auto-label/src/auto-label.ts
```diff
@@ -356,6 +356,70 @@ async function updateStalenessLabel(
   }
 }
 
+/**
+ * Function updating a T-shirt size label on pull request if configuration is enabled.
+ * Removes previous size label if exists
+ */
+async function updatePullRequestSizeLabel(
+  context: Context<'pull_request'>,
```
chingor13, line 364: nit: it's probably easier to pass in `octokit` rather than a context (you're not using any fields from the pull request).

losalex: I also use `context.payload.pull_request.labels` below

packages/label-sync/src/labels.json
losalex: OK, makes sense

chingor13: The extra calls are negligible compared to the number of webhook events we handle. If we for sure want these on all our repos, then we can do it, but we sometimes do get complaints for being heavy-handed in repositories that haven't necessarily opted in.

losalex: Makes complete sense - will add new code to `schedule.repository` event so it will be less `syncLabel` calls

packages/auto-label/src/helper.ts
```diff
@@ -48,8 +51,49 @@ export const DEFAULT_CONFIGS = {
...
+    enabled: false,
+  },
 };
 
+/**
+ * The list of all labels used to mark pull request size.
+ * Given a fact that by default pull request size labeling feature is off,
+ * we will update those labels in repo only when configuration is enabled
+ */
+export const PULL_REUEST_SIZE_LABELS = [
```
chingor13, line 64: nit: typo

```suggestion
export const PULL_REQUEST_SIZE_LABELS = [
```

packages/auto-label/src/auto-label.ts
```diff
@@ -378,7 +448,17 @@ export function handler(app: Probot) {
...
+    // We run it here since the scheduler runs once in a while, so we dont
+    // need to update labels on pull request change event
+    if (config?.requestsize?.enabled) {
+      await syncLabels(
+        context.octokit,
+        owner,
+        repo,
+        helper.PULL_REUEST_SIZE_LABELS
+      );
+    }
```
chingor13, line 461: non-blocking comment: it's also possible to set up multiple `app.on('schedule.repository')` hooks and they will run in parallel on the same webhook.

packages/auto-label/src/helper.ts
```diff
@@ -125,6 +191,7 @@ export interface Config {
   };
   language?: LanguageConfig;
   staleness?: StaleConfig;
+  requestsize?: PullRequestSizeConfig;
```
chingor13, line 194: nit: casing feels weird, but we already have weird casing for `pullrequest` and `labelprefix`.

chingor13 (changes requested):
One small nit on a typo otherwise LGTM

packages/auto-label/src/helper.ts
chingor13: Also need to adjust it's usage

packages/auto-label/src/helper.ts
losalex: Yeah, keeping it consistent

packages/auto-label/src/helper.ts
losalex: Added comment with PR sizes

chingor13 (approved):

packages/auto-label/src/auto-label.ts
chingor13: No issues, it's just if we want to split to avoid having a massive function doing 2 separate things.

PR #3031, opened 2021-12-14
fix(auto-approve): check PR against V2 config if the repo has a V2 config

sofisl:
Fixes 
Screen Shot 2021-12-14 at 11 16 48 AM

When I added the second version for the config in #2949, I did not add logic to make sure that the PR was checked against the V2 configuration in the case that the repository had the V2 configuration. Instead, it was always checking against the V1 configuration. So, the meat of this bug fix is the branching logic in  `src/auto-approve.ts`, line 253, where it first checks to see if the configuration in the repository is V1 or V2.


SurferJeffAtGoogle:
In the PR description, please describe how the code was flawed and how you fixed it.  Alternatively, describe the flaw in a bug and mark this as the fix.

packages/auto-approve/test/auto-approve.test.ts
```diff
@@ -123,6 +130,7 @@ describe('auto-approve', () => {
     checkAutoApproveStub.restore();
     checkCodeOwnersStub.restore();
     getSecretStub.restore();
+    checkPRAgainstConfigV2Stub.restore();
```
chingor13, line 133: This is much simplified if you use a sinon.sandbox and restore the entire sandbox.

packages/auto-approve/test/auto-approve.test.ts
```diff
@@ -140,7 +148,12 @@ describe('auto-approve', () => {
         ));
 
         const scopes = [
-          getConfigFile('testOwner', 'testRepo', 'fake-config', 200),
+          getConfigFile(
```
chingor13, line 151: In the future, I would call these `stubGetConfigFile` or `mockGetConfigFile` to clarify that it's stubbing, not actually getting a config file.

chingor13 (approved):

PR #3032, opened 2021-12-15
feat(trusted-contributor): only add owlbot:run if config found

bcoe:
Only adds `owlbot:run` label if OwlBot configuration is found.

CC: @kurtisvg

### DIFF
--- a/packages/trusted-contribution/src/trusted-contribution.ts
+++ b/packages/trusted-contribution/src/trusted-contribution.ts
@@ -103,25 +103,22 @@ export = (app: Probot) => {
         return;
       }
 
-      // Check if repository is configured for OwlBot, by fetching .github/.OwlBot.yaml,
-      // which exists in all OwlBot-enabled repositories:
-      let hasOwlBotConfig = false;
-      try {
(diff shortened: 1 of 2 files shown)


packages/trusted-contribution/src/trusted-contribution.ts
```diff
@@ -105,9 +102,37 @@ export = (app: Probot) => {
...
+      let hasOwlBotConfig = false;
+      try {
+        await context.octokit.rest.repos.getContent({
+          owner,
+          repo,
+          path: OWLBOT_CONFIG_PATH,
+        });
+        hasOwlBotConfig = true;
+      } catch (_err) {
+        const err = _err as {code: number};
```
sofisl, line 117: Is it safe to catch this error in all cases? Should we throw it in some cases?

sofisl (approved):

packages/trusted-contribution/src/trusted-contribution.ts
bcoe: That's a good question, it might be worth throwing on a non-404 (I think this should happen almost never).

PR #3041, opened 2021-12-23
refactor: upgrade to release-please v13

chingor13:
This updates the release-please bot to use v13 of release-please. We now expose more options like `versioning`, `changelogType`.

We also allow running on the all matching configurations for a branch so you can have more than one configuration per branch.


chingor13:
@bcoe Can probably still review, but we'll wait until the new year to merge when we'll for sure be around to debug.

bcoe:
@chingor13 probably won't be able to look until new year, I'm glad we have this queued up.

packages/release-please/test/release-please.ts
```diff
@@ -757,40 +834,4 @@ describe('ReleasePleaseBot', () => {
       });
     });
   });
-
-  it('should fetch default branch from GitHub if not provided', async () => {
-    const payload = require(resolve(fixturesPath, './push_to_main'));
```
bcoe, line 762: Do we have anythiung that confirms we default to the appropriate branch if not config provided? Is there a test for this in release-please itself?

bcoe (changes requested):
I'm a little worried about dropping the tests for choosing an appropriate branch by default, and allowing this to be overridden in config, is this behavior still caught anywhere.

bcoe (approved):

PR #3083, opened 2022-01-12
fix(repo-metadata-lint): always run opener

bcoe:
The opener method closes PRs if they no longer have problems, so it
should always be run.

### DIFF
--- a/packages/repo-metadata-lint/src/repo-metadata-lint.ts
+++ b/packages/repo-metadata-lint/src/repo-metadata-lint.ts
@@ -47,9 +47,9 @@ export function handler(app: Probot) {
       logger.info(
         `${results.length} validation errors found for ${owner}/${repo}`
       );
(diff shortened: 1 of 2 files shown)


packages/repo-metadata-lint/src/repo-metadata-lint.ts
```diff
@@ -47,9 +47,9 @@ export function handler(app: Probot) {
       logger.info(
         `${results.length} validation errors found for ${owner}/${repo}`
       );
-      const opener = new IssueOpener(owner, repo, context.octokit);
-      await opener.open(results);
     }
+    const opener = new IssueOpener(owner, repo, context.octokit);
```
chingor13, line 51: nit: this is a misnomer if it also closes the issues

chingor13 (approved):

PR #3085, opened 2022-01-12
fix: try to fix race condition between tagging releases and creating PRs

chingor13:
See https://github.com/googleapis/release-please/issues/1208


chingor13:
Going to wait to see if this happens again

PR #3092, opened 2022-01-13
fix(owlbot-cli): allow token to be passed to copy-code

bcoe:
Make it possiblet to continue running local copy of copy-code by passing in a user access token.

### DIFF
--- a/packages/owl-bot/src/cmd.ts
+++ b/packages/owl-bot/src/cmd.ts
@@ -19,12 +19,20 @@ export type Cmd = (
   options?: proc.ExecSyncOptions | undefined
 ) => Buffer;
 
+function sanitize(command: string) {
+  command = command.replace(
+    /x-access-token:([^/]+)/,
+    'x-access-token:REDACTED'
+  );
+  return command;
+}
+
 export function newCmd(logger = console): Cmd {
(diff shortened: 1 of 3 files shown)


packages/owl-bot/src/cmd.ts
```diff
@@ -19,12 +19,20 @@ export type Cmd = (
   options?: proc.ExecSyncOptions | undefined
 ) => Buffer;
 
+function sanitize(command: string) {
+  command = command.replace(
+    /x-access-token:([^/]+)/,
```
bcoe, line 24: I noticed the token shows up in logs, not a huge deal in cloud build since it's short lived ... but probabluy good to redact it.

chingor13 (approved):

chingor13:
Perhaps grabbing from the `GITHUB_TOKEN` env variable could be nice too

bcoe:
> Perhaps grabbing from the GITHUB_TOKEN env variable could be nice too

Good call, I'll push that update.

sofisl (approved):

SurferJeffAtGoogle:
I don't understand why this was necessary.  Please provide the sequence of commands that failed.

Issue #2190, opened 2021-06-26
owlbot: keep license header on .Owlbot.lock.yaml when file is updated


parthea:
I added a license header to `.Owlbot.lock.yaml` in `googleapis/python-dataflow-client` but [owl bot](https://github.com/googleapis/python-dataflow-client/commit/c5c95333083c9c805cdd09287b4c1fc3c8e005cb#diff-28faeab193819b6210664df33e757a79c3b066449bc2b5cce3ebe463ec1e4f3f) removed it. The license header is required by OSPO.

SurferJeffAtGoogle:
Acknowledged.

PR #3103, opened 2022-01-18
fix: add a copyright message to .OwlBot.lock.yaml files

SurferJeffAtGoogle:
Fixes https://github.com/googleapis/repo-automation-bots/issues/2190

### DIFF
--- a/packages/owl-bot/src/write-lock.ts
+++ b/packages/owl-bot/src/write-lock.ts
@@ -1,0 +1,47 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


packages/owl-bot/src/write-lock.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Adds a copyright statement and a comment with the docker image's timestamp
+ * to the text of an .OwlBot.lock.yaml file.
+ */
+export function adornOwlBotLockText(
+  text: string,
+  dockerImageTimestamp: string
+): string {
+  const year = new Date().getFullYear();
```
tmatsuo, line 23: Is this intended to always use the current year?

tmatsuo (approved):

packages/owl-bot/src/write-lock.ts
SurferJeffAtGoogle: Yes.

PR #3117, opened 2022-01-21
fix: open a pull request for lock updates that failed

### LINKED ISSUE #2709: FR(owlbot): Owlbot should open an issue on a repo if owlbot fails to open a PR due to an error
If owlbot fails to open a PR due to an error, the failure is currently not visible to repository maintainers. For example, if there is an issue when owlbot tries to run the post processor or configuration file in a repository, a PR will not be opened by owlbot. There is likely an entry in the owlbot log files somewhere about this error but It would be great if owlbot could open an issue in the repository itself to notify maintainers that owlbot failed so that the issue in the repository can be fixed. Ideally, owlbot should not open an issue for transient failures during the retry period where owlbot is still attempting to create the PR. It should only open an issue when owlbot has given up on creating the PR.

SurferJeffAtGoogle:
Fixes #2709

### DIFF
--- a/packages/owl-bot/cloud-build/force-update-lock-branch.yaml
+++ b/packages/owl-bot/cloud-build/force-update-lock-branch.yaml
@@ -1,0 +1,109 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


tmatsuo (approved):

tmatsuo:
Just a thought. Does it make sense to always create a pull request without running the container? It will give a similar effect.

Issue #2744, opened 2021-10-20
Release-trigger: Fails to trigger releases


Neenu1995:
Release-trigger fails to trigger releases when a number of releases are being initiated by auto-approve use(likely at the same time). This could be caused due to an overload(All the time stamps in the image shows same time). The release can be re-triggered and successfully published by removing and re-adding the `autorelease: triggered` label.
![image](https://user-images.githubusercontent.com/13271327/138160495-a0d3d075-c0ac-4464-82b4-63ddc8c985c9.png)

PR #3122, opened 2022-01-24
feat(canary-bot): split back/frontend deployment

chingor13:
This now has a `canary-bot-cloud-run` frontend that enqueues its worker tasks to target the `canary-bot-cloud-run-backend` Cloud Run service. This allows us the split configuration of the front-end from the back-end.

Towards #2744


chingor13:
After deployment, we should see webhook events hit the Cloud Run frontend and the actual tasks handled by the backend service

tmatsuo (approved):

PR #3127, opened 2022-01-25
fix: record pull request created to copy code from googleapis-gen int…

SurferJeffAtGoogle:
…o client library repositories

The first step towards fixing #2834.

All the new code paths are controlled by flags, so nothing will change when this PR is merged.

Next steps:
1. Switch on flag --track-builds-in-firestore.
2. Wait a couple days.
3. Switch flag --search-depth to zero.
4. Remove all the code that searched through git history.
5. Write code to create cloud builds rather than create pull requests directly.

### DIFF
--- a/packages/owl-bot/src/copy-code.ts
+++ b/packages/owl-bot/src/copy-code.ts
@@ -252,6 +252,7 @@ ${copyTagLine}`,
  * @param sourceRepo: the source repository, either a local path or googleapis/googleapis-gen
  * @param sourceRepoCommit: the commit from which to copy code. Empty means the most recent commit.
  * @param destRepo: the destination repository, either a local path or a github path like googleapis/nodejs-vision.
+ * @returns a url to a github issue or pull request.
  */
 export async function copyCodeAndCreatePullRequest(
   sourceRepo: string,
--- a/packages/owl-bot/src/copy-code.ts
+++ b/packages/owl-bot/src/copy-code.ts
@@ -82,6 +82,7 @@ interface LocalCopy {
 interface CreatedGithubIssue {
   kind: 'CreatedGithubIssue';
   issue: number;
+  link: string;
 }
 
 /**
@@ -222,6 +223,7 @@ ${copyTagLine}`,
     const result: CreatedGithubIssue = {
       kind: 'CreatedGithubIssue',
       issue: issue.data.number,
+      link: issue.data.html_url,
     };
     return result;
   }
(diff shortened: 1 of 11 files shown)


packages/owl-bot/src/copy-code.ts
```diff
@@ -257,7 +259,7 @@ export async function copyCodeAndCreatePullRequest(
   destRepo: AffectedRepo,
   octokitFactory: OctokitFactory,
   logger = console
-): Promise<void> {
+): Promise<string> {
```
chingor13, line 262: nit: can we add the `@return` jsdoc

packages/owl-bot/src/copy-state-store.ts
```diff
@@ -0,0 +1,34 @@
...
+ * been copied from googleapis-gen into the client library repos.
+ */
+export interface CopyStateStore {
+  /**
+   * Records a build id of the post-processor run that was started for
+   * code copied into a client library repository.
+   * @param a unique id for the copy operation.
+   * @param buildId the google cloud build id.
+   */
+  RecordBuildForCopy(copyTag: string, buildId: string): Promise<void>;
```
chingor13, line 26: nit: can we make this lower camel case for method names?

packages/owl-bot/src/copy-state-store.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Finds an existing branch for the copy operation.
+   * @param a unique id for the copy operation.
+   * @returns the empty string if none exists.
+   */
+  FindBuildForCopy(copyTag: string): Promise<string | undefined>;
```
chingor13, line 33: nit: same on lower camel case

packages/owl-bot/src/database.ts
```diff
@@ -207,3 +208,29 @@ export class FirestoreConfigsStore implements ConfigsStore {
     return result;
   }
 }
+
+export class FirestoreCopyStateStore implements CopyStateStore {
```
chingor13, line 212: We use a mocked `CopyStateStore` in tests. Is there a way to test this behavior because it could potentially be complicated queries? Is there possibly a firestore emulator or can we look into an integration test?

packages/owl-bot/src/copy-code.ts
SurferJeffAtGoogle: Done.

packages/owl-bot/src/copy-state-store.ts
SurferJeffAtGoogle: Done.

packages/owl-bot/src/copy-state-store.ts
SurferJeffAtGoogle: Done.

packages/owl-bot/src/database.ts
SurferJeffAtGoogle: Added a system-test, which I confirmed passes when run locally.

SurferJeffAtGoogle:
Ready for another look.

chingor13 (approved):

PR #3163, opened 2022-02-07
fix: run on PR titles in bazel bot

### LINKED ISSUE #3039: OwlBot PR / commit messages are long and get truncated
## Overview

See https://github.com/googleapis/google-cloud-php/pull/4851

The commit language is the concatenation of multiple CL messages. The result looks like this:

```
feat: add the AcknowledgeUserDataCollection operation which
acknowledges the terms of user data collection for the specified
property feat: add the new resource type DataStream, which is
planned to eventually replace WebDataStream, `IosAppDataStr...
```

## The Problem

1. concatenating multiple "conventional commits" is not (as far as I know) a valid way to use conventional commits
2. The "conventional commit" prefix comes from only the first one. If there was a `fix` followed by a `feat`, the PR would only be prefixed with `fix` (In the example above, a `fix!` commit is ignored in the update PR, so the BC-breaking flag is lost)
2. The commit message is too long (and extends past the PR title maximum length)
3. The commit message gets truncated, and extends into the message description, which then breaks formatting with the code-highlighting backticks (see the above linked issue)

## The Solution

When a PR contains multiple CL updates to the protos, make the PR title (and first line of the commit) generic, with the individual updates getting their own lines beneath it. Something like this:

```
[AnalyticsAdmin] feat: multiple updates
feat: add the AcknowledgeUserDataCollection operation which acknowledges the terms of user data collection for the specified property
feat: add the new resource type DataStream, which is planned to eventually replace WebDataStream, `IosAppDataStream`, `AndroidAppDataStream` resources
fix!: remove `GetEnhancedMeasurementSettings`, `UpdateEnhancedMeasurementSettingsRequest`, `UpdateEnhancedMeasurementSettingsRequest` operations from the API
feat: add `CreateDataStream`, `DeleteDataStream`, `UpdateDataStream`, `ListDataStreams` operations to support the new `DataStream` resource
feat: add `DISPLAY_VIDEO_360_ADVERTISER_LINK`,  `DISPLAY_VIDEO_360_ADVERTISER_LINK_PROPOSAL` fields to `ChangeHistoryResourceType` enum
feat: add the `account` field to the `Property` type docs: update the documentation with a new list of valid values for `UserLink.direct_roles` field
```
The parser would need to be updated to look at each commit and take the most meaningful semantic keyword. In this case, `fix!` would need to be combined with `feat` to create `feat!`.

SurferJeffAtGoogle:
Fixes #3039

### DIFF
--- a/packages/bazel-bot/docker-image/generate-googleapis-gen.sh
+++ b/packages/bazel-bot/docker-image/generate-googleapis-gen.sh
@@ -153,7 +153,7 @@ for (( idx=${#ungenerated_shas[@]}-1 ; idx>=0 ; idx-- )) ; do
         fi
 
         # Copy the commit message from the commit in googleapis.
-        git -C "$GOOGLEAPIS" log -1 --format=%s%n%n%b > commit-msg.txt
+        git -C "$GOOGLEAPIS" log -1 --format=%B > commit-msg.txt
         echo "Source-Link: https://github.com/googleapis/googleapis/commit/$sha" >> commit-msg.txt
         # Commit changes and push them.
         git -C "$GOOGLEAPIS_GEN" commit -F "$(realpath commit-msg.txt)"


tmatsuo (approved):

tmatsuo:
For the record, `%B`:
https://git-scm.com/docs/git-log#Documentation/git-log.txt-emBem

PR #3175, opened 2022-02-10
feat: cloud build job to create pull requests for copy branches

SurferJeffAtGoogle:
First step towards not opening pull requests that get immediately closed after the post processor runs because there are no changes.

This PR creates a cloud build job that runs the post processor and opens a pull request if there are changes.  It doesn't include code to actually invoke the cloud build job so the net effect will be zero.


packages/owl-bot/cloud-build/update-copy-branch.yaml
```diff
@@ -0,0 +1,90 @@
...
+  # Examine the changed files and possibly create a pull request.
+  - name: ${_OWL_BOT_CLI}
+    args:
+      - maybe-create-pull-request-for-copy
+      - --pem-path
+      - /workspace/owl-bot-github-key.pem
+      - --app-id
+      - '99011'
+      - --installation
+      - '14695777'
```
danielbankhead, line 87: Are the `_APP_ID` and `_INSTALLATION` variables available in this build? https://github.com/googleapis/repo-automation-bots/blob/cfe86a90724a2c01af3172e4b3c864eb53618457/packages/owl-bot/cloud-build/enqueue-copy-jobs.yaml#L38-L41

danielbankhead (approved):
Looks solid and well documented. 1 nit

packages/owl-bot/cloud-build/update-copy-branch.yaml
SurferJeffAtGoogle: We could make them available, but they never vary.  They're also hard coded like this in other yamls.

PR #3191, opened 2022-02-15
feat(auto-label): respect api N/A label

### LINKED ISSUE #1226: product-auto-label reapplied incorrect product label to issue
## Actual

In the issue https://github.com/googleapis/google-auth-library-python-oauthlib/issues/96, @JustinBeckwith removed the `api: run` label, the next day the product-auto-label bot reapplied it.

This label is correct based on the issue title pattern matching, but is an edge case (matching off a function name that happens to start with `run_`, as opposed to something associated with Cloud Run).

## Expected

If the label is removed, it stays removed.

In the case where the auto-label is re-run in bulk, it notes the most recent manual change to the label in question and defers to that instead of overriding.

sofisl:
Fixes #1226


packages/auto-label/README.md
```diff
@@ -4,11 +4,11 @@ The auto-label bot automatically label issues and pull requests.
 
 There are 3 types of labels available:
 1. Product labels (enabled by default): Product-specific repos are always labeled with the API label
-for that product. The product can also be auto-detected from the issue/PR title.
+for that product. The product can also be auto-detected from the issue/PR title. If you do not want an issue to have its api auto-detected by auto-label, simply label the issue with an "api: N/A" label.
```
SurferJeffAtGoogle, line 7: I always feel like an idiot when instructions tell me to 'simply do yadayada' and I find it difficult.  My $0.02 is to remove the word 'simply' from this sentence.

packages/auto-label/src/auto-label.ts
```diff
@@ -34,6 +34,7 @@ import {
 } from '@google-automations/bot-config-utils';
 import {syncLabels} from '@google-automations/label-utils';
 
+const API_NA_LABEL = 'api: N/A';
```
SurferJeffAtGoogle, line 37: Please add a comment describing that N/A means 'not applicable', and describe why users would add it.

packages/auto-label/src/auto-label.ts
```diff
@@ -91,6 +92,12 @@ handler.addLabeltoRepoAndIssue = async function addLabeltoRepoAndIssue(
       `There was no configured match for the repo ${repo}, trying to auto-detect the right label`
     );
     const apis = await handler.getDriftApis();
+    if (labelsOnIssue && labelsOnIssue?.find(x => x.name === API_NA_LABEL)) {
+      logger.info(
+        `${owner}/${repo}/${issueNumber} is marked as not applicable, skipping`
```
SurferJeffAtGoogle, line 97: Yay!  Smart logging.

packages/auto-label/src/auto-label.ts
tmatsuo, line 37: Can you add this label automatically to all the repos? You can use `syncLabels`.

The existing usage of syncLabels:
https://github.com/googleapis/repo-automation-bots/blob/0f9806afbc2a9bfd7260df817057d676bd0f2912/packages/auto-label/src/auto-label.ts#L451-L460

packages/auto-label/README.md
sofisl: will do!

packages/auto-label/src/auto-label.ts
sofisl: @tmatsuo, don't the changes in `packages/label-sync/src/labels.json` automatically add it to all repos?

tmatsuo: That works too, I think. The benefit of syncing the labels is that we can selectively add the label only to relevant repos (repos with the bot config, etc).

Sometimes, repo owners complain when we add labels that they don't want.

sofisl: Oh I see. The only problem I see is that this problem really only arises when auto-label tries to auto-detect a product, which I think happens when a repo does not have a config. So I think in this case its helpful for all repos to have it, but I might be wrong about the logic I describe above ^

tmatsuo: Got it, maybe we can add the label to `packages/label-sync/src/labels.json` then.

tmatsuo (approved):

PR #3210, opened 2022-02-22
chore: a job to force creating pull requests when post processor fails

SurferJeffAtGoogle:


bcoe:
@SurferJeffAtGoogle curious what problem this solves? should we just be trying to reduce the frequency of post-processor failrues?

SurferJeffAtGoogle:
> @SurferJeffAtGoogle curious what problem this solves? should we just be trying to reduce the frequency of post-processor failrues?

Yes, but because we (Github Automation team) don't own the post processor, it can fail for reasons beyond our control.

When the post processor does fail in a non-transient way (code that doesn't compile, etc), we need to report that failure to the repo maintainer.  **This Cloud Build job creates a pull request even when the post processor fails so that the repo maintainer can see the failure.**  Otherwise, the the repo maintainer will remain oblivious, and the code in the repo will fall behind the protos.

This follows the same pattern we use for lock updates.  
https://github.com/googleapis/repo-automation-bots/blob/main/packages/owl-bot/cloud-build/update-lock-branch.yaml
https://github.com/googleapis/repo-automation-bots/blob/main/packages/owl-bot/cloud-build/force-update-lock-branch.yaml

SurferJeffAtGoogle:
https://github.com/googleapis/repo-automation-bots/issues/2709 was reported for pull requests triggered by new post processor images.  This PR pre-emptively fixes the same issue for pull requests triggered by copying code from googleapis/googleapis  with [a new code path](https://github.com/googleapis/repo-automation-bots/compare/main...SurferJeffAtGoogle:no-noop4) that runs the post processor *before* creating a pull request.

bcoe (approved):

PR #3222, opened 2022-02-25
fix: don't deleted copy branches that have changed in the meantime

SurferJeffAtGoogle:
Avoids one potential data race.


bcoe (commented):
Is a side effect of this, that in cases where we have contention on a copy branch, the branch might never be deleted?

I wonder if we should consider dropping the branch deletion logic entirely in the future, but introduce an automation that cleans up stale branches?

PR #3250, opened 2022-03-02
feat(owl-bot): combine pull requests for monorepos

SurferJeffAtGoogle:
For #2835.  Doesn't yet fix it because new behavior is disabled by flag.


packages/owl-bot/src/bin/commands/copy-code-and-create-pull-request.ts
```diff
@@ -12,12 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// To Run: node ./build/src/bin/owl-bot.js copy-code-and-create-pull-request <args>
-
```
tmatsuo, line 16: Why do you remove the comment above?

SurferJeffAtGoogle: That was a mistake.  I restored it.

packages/owl-bot/src/bin/commands/copy-code-into-pull-request.ts
```diff
@@ -47,7 +47,7 @@ export const copyCodeIntoPullRequestCommand: yargs.CommandModule<{}, Args> = {
       })
       .option('owl-bot-yaml-path', {
```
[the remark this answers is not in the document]
SurferJeffAtGoogle: Preserving the existing name makes the roll out and potential roll back easier.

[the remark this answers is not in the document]
tmatsuo: Understood

packages/owl-bot/src/copy-code.ts
```diff
@@ -329,187 +249,249 @@ export function branchNameForCopy(yamlPath: string): string {
...
+      } else {
+        // There's no open pull request; We'll need to create a pull request.
+        leftOvers.push(...yamlPathSet);
+      }
     }
-    return {kind: 'AppendedPullRequest', link: pulls.data[0].html_url};
-  } else {
-    return whatHappened;
+    yamlPathSets = [leftOvers];
+  } while (appendedPullCount > 0 && yamlPathSets[0].length > 0);
```
tmatsuo, line 416: I don't understand the condition for the `while` here. When is this condition met?
Is it worth adding a comment?

packages/owl-bot/src/copy-code.ts
```diff
@@ -594,45 +576,60 @@ export async function regeneratePullRequest(
   }
   const pull = pulls.data[0];
 
+  // If something goes wrong while examining the state of the open pull request,
+  // report the error on the log and in a comment on the pull request.
+  const reportError = async (error: string) => {
+    logger.error(`Error in ${pull.html_url}:\n${error}`);
+    await octokit.pulls.createReviewComment({
```
tmatsuo, line 583: I used createIssueComment for similar purpose. Do you prefer a review comment?

packages/owl-bot/src/copy-code.ts
```diff
@@ -649,11 +646,13 @@ export async function regeneratePullRequest(
 
   // Update the PR body with the full commit history.
   const {title, body} = resplit(commitMsg, WithRegenerateCheckbox.Yes);
+
+  const apiList = apiNames.length > 3 ? 'Many APIs' : apiNames.join(',');
```
tmatsuo, line 650: A small nit, but is it worth having a helper function for creating apiList?

packages/owl-bot/src/copy-code.ts
SurferJeffAtGoogle: In this case, it's an open review that we need to comment on, so `createReviewComment()` seemed like the natural choice.

packages/owl-bot/src/copy-code.ts
SurferJeffAtGoogle: I thought about it, because the same line of code appears twice.  If it appears a third time, I'll wrap it in a function.

SurferJeffAtGoogle:
Ready for another look.

tmatsuo (approved):

PR #3254, opened 2022-03-04
fix(owlbot): lock post-processor on SHA to prevent race conditions

### LINKED ISSUE #3225: bug(owl-bot): reduce number of commits on new owlbot PRs from 4 to 2
Recently I've noticed that new owlbot PRs have 4 commits instead of 2. Each commit takes a hit on our github quota for status checks so we should the number of commits back to 2. I have noticed more issues with kokoro quota recently and I think this may be related.

Recently new owl bot PRs have 4 commits:
https://github.com/googleapis/python-dataproc/pull/350/commits
image

About 2 weeks ago, new owl bot PRs had only 2 commits:
https://github.com/googleapis/python-dataproc/pull/342/commits
image

bcoe:
If a label and synchronize event happen in conjunction, it leads to two post-processor
runs kicking off. We now lock for 60s on SHA to prevent this from happening.

Fixes #3225

### DIFF
--- a/packages/owl-bot/package.json
+++ b/packages/owl-bot/package.json
@@ -63,6 +63,7 @@
     "uuid": "^8.3.2"
   },
   "dependencies": {
+    "@google-automations/datastore-lock": "^3.1.0",
     "@google-automations/label-utils": "^2.0.0",
     "@google-cloud/cloudbuild": "^2.2.6",
     "@octokit/rest": "^18.6.7",
--- a/packages/owl-bot/src/app.ts
+++ b/packages/owl-bot/src/app.ts
@@ -15,7 +15,7 @@
 // eslint-disable-next-line node/no-extraneous-import
 import {Probot} from 'probot';
 import {GCFBootstrapper} from 'gcf-utils';
-import {OwlBot} from './owl-bot';
+import {owlbot} from './owl-bot';
 const bootstrap = new GCFBootstrapper();
 // Unlike other probot apps, owl-bot-post-processor requires the ability
 // to generate its own auth token for Cloud Build, we use the helper in
@@ -23,7 +23,7 @@ const bootstrap = new GCFBootstrapper();
 module.exports.owl_bot = bootstrap.gcf(
   async (app: Probot) => {
     const config = await bootstrap.getProbotConfig(false);
-    OwlBot(config.privateKey, app);
+    owlbot.OwlBot(config.privateKey, app);
   },
   // owl-bot typically waits for a Cloud Build build to finish.
   // Cloud Build has a limit for number of concurrent builds, so
--- a/packages/owl-bot/src/cmd.ts
+++ b/packages/owl-bot/src/cmd.ts
@@ -25,7 +25,9 @@ export function newCmd(logger = console): Cmd {
     options?: proc.ExecSyncOptions | undefined
   ): Buffer => {
     logger.info(command);
-    return proc.execSync(command, options);
+    const output = proc.execSync(command, options);
+    if (typeof output === 'string') return Buffer.from(output);
+    else return output;
   };
   return cmd;
 }
--- a/packages/owl-bot/src/core.ts
+++ b/packages/owl-bot/src/core.ts
@@ -126,7 +126,8 @@ export async function triggerPostProcessBuild(
     );
     return null;
   }
-
+  if (!prData?.head?.repo?.full_name)
+    throw Error(`invalid response ${owner}/${repo} pr=${args.pr}`);
   const [prOwner, prRepo] = prData.head.repo.full_name.split('/');
   const cb = core.getCloudBuildInstance();
   const [resp] = await cb.runBuildTrigger({
@@ -371,6 +372,8 @@ export async function fetchOwlBotLock(
     repo,
     pull_number: pullNumber,
   });
+  if (!prData?.head?.repo?.full_name)
+    throw Error(`invalid response ${owner}/${repo} pr=${pullNumber}`);
   const [prOwner, prRepo] = prData.head.repo.full_name.split('/');
   const configString = await getFileContent(
     prOwner,
@@ -630,6 +633,8 @@ async function updatePullRequestAfterPostProcessor(
       pull_number: prNumber,
       state: 'closed',
     });
+    if (!pull?.head?.repo?.full_name)
+      throw Error(`invalid response ${owner}/${repo} pr=${prNumber}`);
     if (pull.head.repo.full_name === `${owner}/${repo}`) {
       logger.info(`Deleting branch ${pull.head.ref}`);
       await octokit.git.deleteRef({owner, repo, ref: `heads/${pull.head.ref}`});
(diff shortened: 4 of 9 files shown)


packages/owl-bot/package.json
```diff
@@ -63,6 +63,7 @@
     "uuid": "^8.3.2"
   },
   "dependencies": {
+    "@google-automations/datastore-lock": "^3.1.0",
```
bcoe, line 66: New version of datastore-lock with the `peek()` method.

packages/owl-bot/src/app.ts
```diff
@@ -15,15 +15,15 @@
...
+import {owlbot} from './owl-bot';
 const bootstrap = new GCFBootstrapper();
 // Unlike other probot apps, owl-bot-post-processor requires the ability
 // to generate its own auth token for Cloud Build, we use the helper in
 // GCFBootstrapper to load this from Secret Manager:
 module.exports.owl_bot = bootstrap.gcf(
   async (app: Probot) => {
     const config = await bootstrap.getProbotConfig(false);
-    OwlBot(config.privateKey, app);
+    owlbot.OwlBot(config.privateKey, app);
```
bcoe, line 26: Pulled exported methods from OwlBot service onto `owlbot` object, so that it's easier to mock.

packages/owl-bot/src/cmd.ts
```diff
@@ -25,7 +25,9 @@ export function newCmd(logger = console): Cmd {
     options?: proc.ExecSyncOptions | undefined
   ): Buffer => {
     logger.info(command);
-    return proc.execSync(command, options);
+    const output = proc.execSync(command, options);
+    if (typeof output === 'string') return Buffer.from(output);
```
bcoe, line 29: There was a breaking change in `@types/node` that bit us, this just corrects the type information.

packages/owl-bot/src/core.ts
```diff
@@ -126,7 +126,8 @@ export async function triggerPostProcessBuild(
     );
     return null;
   }
-
+  if (!prData?.head?.repo?.full_name)
```
bcoe, line 129: similarly, there was a breaking change in a patch for octokit, which this addresses.

packages/owl-bot/src/owl-bot.ts
```diff
@@ -49,11 +50,23 @@ interface PubSubContext {
   };
 }
 
-export function OwlBot(
-  privateKey: string | undefined,
-  app: Probot,
-  db?: Db
-): void {
+async function acquireLock(target: string): Promise<DatastoreLock | boolean> {
```
bcoe, line 53: The bulk of the logic is here, double check for me, goal:

1. if we a lock already exists, return false immediately.
2. otherwise acquire and return lock.

packages/owl-bot/src/owl-bot.ts
```diff
@@ -49,11 +50,23 @@ interface PubSubContext {
...
+  const lock = new DatastoreLock('owlbot', target, 25 * 1000);
+  if (await lock.peek()) {
+    return false;
+  }
+  const result = await lock.acquire();
```
bcoe, line 58: this is taken from library docs;

```js
const result = await lock.acquire();
if (!result) {
	// failure
}
```

I assume the lock returns `false` if it can't be acquired.

bcoe (commented):
Left notes about my update.

I had to update `package-lock.json` to address some compilation errors locally, I think there had been some breakages of types in patch versions of dependencies.

packages/owl-bot/src/owl-bot.ts
```diff
@@ -49,11 +50,23 @@ interface PubSubContext {
   };
 }
 
-export function OwlBot(
-  privateKey: string | undefined,
-  app: Probot,
-  db?: Db
-): void {
+async function acquireLock(target: string): Promise<DatastoreLock | boolean> {
+  const lock = new DatastoreLock('owlbot', target, 25 * 1000);
```
sofisl, line 54: Perhaps we should put the time into a constant and add also add it as an optional parameter to the function

packages/owl-bot/src/owl-bot.ts
```diff
@@ -49,11 +50,23 @@ interface PubSubContext {
...
+    return false;
+  }
+  const result = await lock.acquire();
+  if (!result) {
+    return false;
+  }
+  return lock;
+}
+
+function isLock(lock: DatastoreLock | boolean): lock is DatastoreLock {
```
sofisl, line 65: Instead of returning two different types in `acquireLock` and then adding this function, what if we just returned the lock or else threw an exception and handled it?

packages/owl-bot/src/owl-bot.ts
```diff
@@ -161,6 +182,14 @@ export function OwlBot(
         return;
       }
 
+      // Short-circuit if post-processor already running for this SHA:
+      const target = `${owner}_${repo}_${context.payload.pull_request.head.sha}`;
+      const lock = await owlbot.acquireLock(target);
+      if (!isLock(lock)) {
```
sofisl, line 188: If we just throw an exception, we won't have to add this clause either, just a try-catch block

packages/owl-bot/test/owl-bot.ts
```diff
@@ -1738,6 +1745,213 @@ describe('owlBot', () => {
...
+    });
+    await probot.load((app: Probot) => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      owlbot.OwlBot('abc123', app, sandbox.stub() as any);
+    });
+  });
+  afterEach(() => {
+    sandbox.restore();
+  });
+  it('returns immediately if lock cannot be acquired', async () => {
```
sofisl, line 1775: Maybe we should assert a test where we do create the lock, and we release?

packages/owl-bot/src/owl-bot.ts
bcoe: I think starting with a `const` is a good idea, makes it more readable. I don't think there's a good reason to configure it differently for different calls at this time, so I wouldn't bother making it an argument.

packages/owl-bot/src/owl-bot.ts
bcoe: that's a reasonable suggestion 👍

packages/owl-bot/test/owl-bot.ts
bcoe: I added tests ensuring `release` is called (which I think is valuable). I didn't bother testing the behavior of an event being triggered after we call `.release`, as I don't think this gives us any additional coverage (_since the acquireLock method is mocked_).

sofisl (approved):

tmatsuo (approved):

tmatsuo:
I'm not blocking, but this solution is still incomplete. It holds lock for 60 seconds and if a label event comes after 60 seconds, it will still kick off the postprocessor.

bcoe:
> I'm not blocking, but this solution is still incomplete. It holds lock for 60 seconds and if a label event comes after 60 seconds, it will still kick off the postprocessor.

I'm not concerned about this, the bug we're trying to address is two events being kicked off at the same time, due to a synchronize event from a push, and a label event from trusted contributor. If someone happens to manually label the PR while the post-processor is 30 seconds in to running, I think it's okay that this edge-case triggers a second post-processor run.

Issue #3242, opened 2022-03-01
chore: auto-label T-shirt size indicator should be assigned on every new pull request in all googleapis repositories


losalex:
Can we enable [Size Label Bot](https://github.com/googleapis/repo-automation-bots/tree/main/packages/auto-label#size-label) on all [googleapis](https://github.com/googleapis) repositories? Below some data points:

1. It runs already pretty flawless on [google-cloud-go](https://github.com/googleapis/google-cloud-go), [nodejs-logging](https://github.com/googleapis/nodejs-logging), [java-logging](https://github.com/googleapis/java-logging), [python-logging](https://github.com/googleapis/python-logging) and 6 more.
2. It runs only when PR is modified and does not runs on scheduler events, so the bot should not consume too much resources.
3. It will make sure that size indicator exists on every PR same as it is done in rest of company systems.

jskeet:
For google-cloud-dotnet, I'd be nervous of it generating even more emails per OwlBot PR - especially as the size will almost *always* change between the initial OwlBot generation and the subsequent post-processing. Maybe the label doesn't generate an email, which wouldn't make it too bad. There's still quite a lot happening on even the most trivial PRs though - [example](https://github.com/googleapis/google-cloud-dotnet/pull/8284). It can become slightly tricky to see the really important bits with lots of bots weighing in on every PR.

losalex:
As decided, Size-Label bot is going to be enabled on several googleapi repositories per language - below are associated PRs: [1384](https://github.com/googleapis/synthtool/pull/1384), [1383](https://github.com/googleapis/synthtool/pull/1383), [1382](https://github.com/googleapis/synthtool/pull/1382), [1381](https://github.com/googleapis/synthtool/pull/1381)

PR #3304, opened 2022-03-18
chore: auto-label T-shirt size indicator should be assigned on every new pull request in all googleapis repositories

losalex:
Enabling T-shirt size to be assigned on every PR submitted in Google repositories. Below some data points:

1. It runs already pretty flawless on [google-cloud-go](https://github.com/googleapis/google-cloud-go), [nodejs-logging](https://github.com/googleapis/nodejs-logging), [java-logging](https://github.com/googleapis/java-logging), [python-logging](https://github.com/googleapis/python-logging) and 6 more.
2. It runs only when PR is modified and does not runs on scheduler events, so the bot should not consume too much resources.
3. It will make sure that size indicator exists on every PR same as it is done in rest of company systems.
4. 
Fixes #[3242](https://github.com/googleapis/repo-automation-bots/issues/3242) 🦕


bcoe:
Added do not merge, pending communicating this change with users.

losalex:
> Historically changing the default behavior has been making some people uncomfortable.
> 
> In general, this type of behavior change should be handled in a more user friendly manner.
> 
> For example, the bot can automatically propose a change (PR) in the config file explaining the new feature, etc.

Makes sense. I wonder if I actually should ask users who wants to opt out from this feature and see if there will be a few who wants to opt out - this way I could make a change in their repo manually perhaps

tmatsuo:
I think we can close this. Feel free to reopen if you think differently.

Issue #3313, opened 2022-03-22
Convert bots to use cloud run front-end


chingor13:
See #2114

- [x] auto-approve
- [x] auto-label
- [x] bazel-bot
- [x] blunderbuss
- [x] canary-bot
- [x] cherry-pick-bot
- [x] conventional-commit-lint
- [x] do-not-merge
- [x] failurechecker
- [x] flakybot
- [x] generated-files-bot
- [x] header-checker-lint
- [x] merge-on-green
- [x] label-sync
- [x] owl-bot
- [x] policy
- [x] release-please
- [x] release-trigger
- [x] repo-metadata-lint
- [x] secret-rotator
- [x] snippet-bot
- [x] sync-repo-settings
- [x] trusted-contribution

tmatsuo:
Initially I thought it's better to have a shared frontend server. I think we can just deploy a dedicated frontend server for all the bots.

tmatsuo:
Deployed Cloud Run frontend for all the bots 🎉

PR #3315, opened 2022-03-22
build: deploy do-not-merge frontend

tmatsuo:
Towards #3313


chingor13 (changes requested):
I think this one is missing a `Dockerfile` too

PR #3317, opened 2022-03-23
build(owl-bot): refactor frontend deployment

tmatsuo:
so that we can deploy the backend cloud run with lower risk.


chingor13 (commented):
The frontend and backend can run the same docker image

tmatsuo:
> The frontend and backend can run the same docker image

Yeah I know, but separating allows us to deploy the backend cloud run with minimal risk.

tmatsuo:
Also, the owlbot bootstrap needs to fetch secrets which is a performance penalty for cold start.

chingor13 (approved):
Currently the docker image is serving the frontend so, this is just renaming the image in case we want different images between the frontend and backend?

tmatsuo:
@chingor13 Yes, I'll add backend docker image in the next PR.

Issue #3318, opened 2022-03-24
Migrate owlbot to Cloud Run


tmatsuo:
- In general, Cloud Run has some benefits including longer timeout, concurrency, etc
- Also the longer timeout will likely mitigate #3303 

The Pub/Sub handler in the serverless scheduler proxy is targetting Cloud function endpoint, so we'll likely need to keep the Cloud Function backend too unless we change the target in serverless sheculer proxy.

tmatsuo:
Done

PR #3320, opened 2022-03-25
build(owl-bot): deploy backend in cloud run

tmatsuo:
This should be safely deployed because there's no changes in the
frontend and cloud functions deployment.

After deploying the backend, we can change the task target in the
frontend to switch the backend.

Towards #3318

### DIFF
--- a/packages/owl-bot/Dockerfile.backend
+++ b/packages/owl-bot/Dockerfile.backend
@@ -1,4 +1,4 @@
-# Copyright 2021 Google LLC
+# Copyright 2022 Google LLC
 #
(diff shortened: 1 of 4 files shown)


packages/owl-bot/Dockerfile.backend
```diff
@@ -0,0 +1,49 @@
+# Copyright 2021 Google LLC
```
chingor13, line 1: Nit: 2022

tmatsuo:
@chingor13 Thanks, PTAL

chingor13 (approved):

PR #3365, opened 2022-03-31
chore(canary-bot): Use gcf-utils 13.3.0

tmatsuo:
Also route the canary-bot-cloud-run to the Cloud Run backend and narrow the backend access.


sofisl:
Are we fully deprecating the use of canary bots with functions? Perhaps it would be useful to test both while we still support both.

tmatsuo:
@sofisl This PR doesn't deprecate Functions version. The Functions version will still alive. We have two bots, one in Cloud Run, one in Functions.

sofisl (approved):

PR #3411, opened 2022-04-11
fix: remove docs from required checks for nodejs

SurferJeffAtGoogle:
Links more frequently break because the link target disappeared than because the code change introduce a new broken link.


chingor13 (approved):

tmatsuo:
It seems like the test depends on the current value. @SurferJeffAtGoogle Can you fix the test somehow?

Issue #3424, opened 2022-04-13
Use dedicated service accounts for bot deployments


tmatsuo:
The service accounts for the bot deployment are:
- web-frontend@repo-automation-bots.iam.gserviceaccount.com
  for Cloud Run frontend deployment 
- web-backend@repo-automation-bots.iam.gserviceaccount.com 
  for normal backend deployment
- policy-bot-backend@repo-automation-bots.iam.gserviceaccount.com
  If a bot needs additional permissions, it's advised that we create a dedicated service account. Policy bot needs BigQuery write permission.

-- 
- [x] auto-approve
- [x] auto-label
- [x] bazel-bot
- [x] blunderbuss
- [x] canary-bot
- [x] cherry-pick-bot
- [x] conventional-commit-lint
- [x] do-not-merge
- [x] failurechecker
- [x] flakybot
- [x] generated-files-bot
- [x] header-checker-lint
- [x] label-sync
- [x] merge-on-green
- [x] owl-bot
- [x] policy
- [x] release-please
- [x] release-trigger
- [x] repo-metadata-lint
- [x] secret-rotator
- [x] snippet-bot
- [x] sync-repo-settings
- [x] trusted-contribution

chingor13:
`release-trigger` and `repo-metadata-lint` already appear to use custom service accounts

tmatsuo:
owl-bot is still using the default account for the frontend deployment.

tmatsuo:
It's done now.

tmatsuo:
It turned out that the serverless-scheduler-proxy is using the default compute service account.

PR #3432, opened 2022-04-13
build(owl-bot): use dedicated service accounts

tmatsuo:
for cloud function service and the Cloud Run frontend

part of #3424


tmatsuo:
I think we need to allow `owl-bot-web-backend` to act as `task-caller`

sofisl (approved):

PR #3433, opened 2022-04-13
build: use web-backend service account for deployments

tmatsuo:
`web-backend` is a new service account dedicated for running our bots.

This PR will use `web-backend` service account for the following 4 bots:
- auto-approve
- auto-label
- blunderbuss
- cherry-pick-bot

Part of #3424


kurtisvg:
Wouldn't the best practice here to use a different service account for each bot?

tmatsuo:
@kurtisvg Likely most of the bots will be ok with the current set of permissions. When we need to add a permission to a bot, we'll create a new service account specific for that bot. For example, policy bot needs BigQuery editor, so it has a dedicated service account.

kurtisvg:
@tmatsuo I think from an auditability perspective, it's a best practice to make a different service account for each upfront, even if they permissions or the same/identical now.

kurtisvg (approved):
Approving for /blunderbuss

sofisl (approved):

PR #3434, opened 2022-04-14
feat: initial files for owlbot-bootstrapper

sofisl:
Opening this PR in lieu of #3356 since there were issues with the commit authorship. I've addressed comments from the last PR though.

Note: this bot is not fully wired up as a probot app since we do not have a solid indicator of whether an API wants this. Instead, it's set up as a CLI tool currently that kicks off a build file.

After we merge this PR, we can merge the terraform configs to create the containers, etc. After that, we can create the bot in Github, and we'll need to follow up with another PR to change those values (installation ID, etc.)


packages/owlbot-bootstrapper/common-container/split-repo.ts
```diff
@@ -0,0 +1,181 @@
...
+   */
+  public async _commitAndPushToMain(
+    repoName: string,
+    directoryPath: string,
+    githubToken?: string,
+    repoUrl?: string
+  ) {
+    if (githubToken) {
+      cmd(
+        `git add .; git commit -m "feat: adding initial files"; git branch -M main; git remote add origin https://x-access-token:${githubToken}@github.com/${ORG}/${repoName}; git push -u origin main`,
```
chingor13, line 120: Since this is shelling out to call `git`, we should ensure that our container has the necessary packages installed

chingor13 (commented):
This is looking like a good start!

Organizationally, I see this as 3 separate node modules and might separate the code that way:

* `packages/owlbot-bootstrapper/app`
* `packages/owlbot-bootstrapper/cli`
* `packages/owlbot-bootstrapper/common-container`
```
Each could be its own module tested independently as it doesn't look like they share code between them.

WDYT?

File: packages/owlbot-bootstrapper/cloudbuild-owlbot-bootstrapper.yaml [hunk 2]
```diff
@@ -0,0 +1,52 @@
...
+
+steps:
+  # Clones the target repository and opens a branch for mono-repos, for split-repos creates a repository
+  - name: ${_CONTAINER}
+    env:
+    - 'LANGUAGE=$_LANGUAGE'
+    - 'PROJECT_ID=$PROJECT_ID'
+    - 'API_ID=$_API_ID'
+    - 'REPO_TO_CLONE=$_REPO_TO_CLONE'
+    - 'IS_PRE_PROCESS=$_IS_PRE_PROCESS'
```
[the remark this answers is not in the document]
sofisl: The reason I chose to use an env variable is because we're triggering a cloud build script with the command once, and then the cloud build script just runs the containers in sequence. In order to determine whether it should be a 'pre' or 'post' process, the cloud build file itself changes the env variable [here](https://github.com/googleapis/repo-automation-bots/blob/bc193dc56ce052471f663b5ae3cb3430b70c39db/packages/owlbot-bootstrapper/cloudbuild-owlbot-bootstrapper.yaml#L50).

Since we're only using the command-line call once, how would we alternate between the pre- and post- steps using just the `command` argument?

packages/owlbot-bootstrapper/common-container/common-container.ts
```diff
@@ -0,0 +1,168 @@
...
+      process.env.BUILD_ID,
+      process.env.PROJECT_ID,
+      process.env.LANGUAGE,
+      (err as any).toString()
+    );
+    throw err;
+  }
+}
+
+main();
```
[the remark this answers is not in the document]
sofisl: So we are using yargs (now in cli/cli.ts), which then calls the build file, which calls the container that has this logic. I think we need to put those arguments as env variables in order to pass them into the build file. We need to use the build file to call the containers. Eventually we'll phase out calling the build file from the command line, and instead the bot will call it, but hope that helps answer your comment.

packages/owlbot-bootstrapper/common-container/mono-repo.ts
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Monorepo class
+ *
+ * @param language the Language of the particular monorepo
+ * @param repoToCloneUrl the url for the repo to clone without the preceeding https://, e.g., just github.com/googleapis/nodejs-kms.git
+ * @param repoName the name for the repo to clone, i.e., nodejs-kms
+ * @param githubToken a short-lived access Github access token
+ * @param octokit an instance of Octokit
+ */
+export class MonoRepo {
```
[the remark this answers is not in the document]
sofisl: They have slight differences, for example the actual pre- and post- steps. I don't totally envision us expanding into different objects, but definitely would consider doing it if we end up adding more processes.

SurferJeffAtGoogle:
@chingor13, since you've taken on this review, I'll switch the assignee to you.

packages/owlbot-bootstrapper/cloudbuild-owlbot-bootstrapper.yaml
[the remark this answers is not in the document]
sofisl: Talked it over, implemented!

packages/owlbot-bootstrapper/common-container/github-authenticator.ts
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * @param projectId project ID where owlbot-bootstrapper secret is stored
+ * @param appInstallationId installation ID for owlbot-bootstrapper on github org
+ * @param secretManagerClient a secret manager service client
+ */
+export class GithubAuthenticator {
+  projectId: string;
+  appInstallationId: string;
+  secretManagerClient: SecretManagerServiceClient;
+  OWLBOT_SECRET_NAME = 'owlbot-bootstrapper-app';
```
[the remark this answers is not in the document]
sofisl: Created #3474 to track, don't want this work to be bunched in with owlbot bootstapper.

sofisl:
@chingor13, thanks for the review, few changes/comments:

1. Now using arguments instead of env variables and using yargs to parse; 
2. This change makes it so that there's shared code between cli and common-container, so I can't modularize the three separately, 
3. Created a tracking ticket #3474 to make octokit authenticator its own class since I don't want to make this PR any larger

PTAL, thanks!

packages/owlbot-bootstrapper/app/run-trigger.ts
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file is kicked off by a manual CLI invocation, either by an event or by a human.
+// It kicks off a trigger in cloud build.
+
+import {CloudBuildClient} from '@google-cloud/cloudbuild';
+
+export async function runTrigger(argv: any, cb: CloudBuildClient) {
```
chingor13, line 20: Can we define this type? It looks like we do already have a type for it.

packages/owlbot-bootstrapper/cloudbuild-owlbot-bootstrapper.yaml
```diff
@@ -0,0 +1,38 @@
...
+# Run by a CLI command manually (by a human), or when googleapis/googleapis API Index changes
+# In order to run containers, so that languages can have their own isolated workflow to generate
+# their language-specific repositories, we need to run these containers in Cloud Build. This file
+# will trigger the pre- and post- processing steps of a language-specific container, as well as running
+# the language-specific container itself.
+
+steps:
+  # Clones the target repository and opens a branch for mono-repos, for split-repos creates a repository
+  - name: ${_CONTAINER}
+    args: ["pre-process", "--language", "$_LANGUAGE", "--projectId", "$_PROJECT_ID", "--apiId", "$_API_ID", "--repoToClone", "$_REPO_TO_CLONE", "--installationId", "$_INSTALLATION_ID'", "--buildId", "$BUILD_ID"]
```
chingor13, line 24: nit: can we make this multi-line so any diffs are easier to read

packages/owlbot-bootstrapper/common-container/cloudbuild-deploy-container.yaml
```diff
@@ -0,0 +1,36 @@
...
+# This build file will build and push the common-container in googleapis-bootstrapper
+# to container registry. There will be a trigger in owlbot-bootstrapper that
+# will build and push the image when any merges to main in repo-automation-bots occur.
+# This will ensure that the common-container we call will stay up-to-date.
+
+steps: 
+  - name: gcr.io/cloud-builders/docker
+    id: "build-docker"
+    waitFor: ["-"]
+    dir: packages/owlbot-bootstrapper
```
chingor13, line 24: Looks like this needs an extra directory
```suggestion
    dir: packages/owlbot-bootstrapper/common-container
```

packages/owlbot-bootstrapper/common-container/mono-repo.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  /**
+   * Clones a repo from github
+   *
+   * @param githubToken a short-lived access Github access token
+   * @param repoToCloneUrl from where to clone the repo
+   * @param directoryPath where to clone the repo to
+   */
+  public async _cloneRepo(
```
chingor13, line 62: nit: does this need to be public (the underscore-led name implies it doesn't)

chingor13 (changes requested):
Looks great! A couple tiny nits.

Also, I would expect there to be 3 separate package.json files (1 for app, cli, and common-container) and have the tests separated into each folder. This can be done in a follow.

packages/owlbot-bootstrapper/common-container/cloudbuild-deploy-container.yaml
sofisl: great catch, thank you!

packages/owlbot-bootstrapper/common-container/mono-repo.ts
sofisl: It doesn't, but it's making testing easier. Do you have any recs?

packages/owlbot-bootstrapper/package.json
```diff
@@ -18,13 +18,9 @@
...
-    "compile": "tsc -p .",
-    "start": "probot run ./build/src/googleapis-bootstrapper.js",
-    "start:local": "node ./build/common-container/common-container.js",
-    "pretest": "npm run compile",
-    "test": "cross-env LOG_LEVEL=fatal c8 mocha --exit build/test",
-    "fix": "gts fix",
-    "lint": "gts check"
+    "test": "cd cli; npm t; cd ../app; npm t; cd ../common-container; npm t",
+    "fix": "cd cli; npm run fix; cd ../app; npm run fix; cd ../common-container; npm run fix",
+    "lint": "cd cli; npm run lint; cd ../app; npm run lint; cd ../common-container; npm run lint"
```
chingor13, line 23: nit: this can be done in a follow up, but if you do run the test from this parent folder, the `;` will make it so each test runs regardless of if the previous command passed or failed. If the `cli` tests fail, then it will continue. If the `common-container` tests pass, then it will report success regardless of if `cli` or `app` tests failed

chingor13 (approved):

packages/owlbot-bootstrapper/common-container/mono-repo.ts
chingor13: This code is now an implementation detail of the common container so it's less of a big deal.

Visibility in typescript is a bit hard because anything exported from any module is technically public, but generally, we can assume that a developer should only `import {something} from 'my-package-name'`. Public methods on an exported class can be unintentionally exposed if the exported class is considered part of the public interface (again, not a big concern for this package). Since this method also requires no instance methods, you could move it to be an exported (from this module, but not the main package) function within this file so it can be tested.

chingor13: Another option is to test the public methods that use this code. Generally, testing simpler private methods is not advised as it's an implementation detail of the real interface and thus you're coupling your tests to the implementation.

PR #3437, opened 2022-04-14
fix(owlbot): stop crashing on DatastoreLockError

bcoe:
There are legitimate reasons this error might happen, e.g., lock times out and another post-processor build kicks off in the interim.

Still warn, but stop throwing.

### DIFF
--- a/packages/owl-bot/src/owl-bot.ts
+++ b/packages/owl-bot/src/owl-bot.ts
@@ -14,7 +14,10 @@
 
 // eslint-disable-next-line node/no-extraneous-import
 import admin from 'firebase-admin';
-import {DatastoreLock, DatastoreLockError} from '@google-automations/datastore-lock';
+import {
+  DatastoreLock,
+  DatastoreLockError,
+} from '@google-automations/datastore-lock';
 import {FirestoreConfigsStore, Db} from './database';
 // eslint-disable-next-line node/no-extraneous-import
 import {Probot, Logger} from 'probot';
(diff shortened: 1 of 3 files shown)


packages/owl-bot/src/owl-bot.ts
```diff
@@ -64,6 +64,19 @@ async function acquireLock(target: string): Promise<DatastoreLock> {
   return lock;
 }
 
+async function releaseLock(lock: DatastoreLock): Promise<void> {
+  try {
+    await lock.release();
+  } catch (_err: unknown) {
+    const err = _err as Error;
```
chingor13, line 71: When I've written these try/catch blocks in typescript, my IDE infers that the caught exception is an Error -- is this more save?

packages/owl-bot/src/owl-bot.ts
```diff
@@ -64,6 +64,19 @@ async function acquireLock(target: string): Promise<DatastoreLock> {
...
+    if (err.message.includes('DatastoreLockError')) {
```
chingor13, line 72: nit: would checking the error `instanceof` be more resilient than relying on the message string?

bcoe: Yes, but this would require exposing the Error from `DatastoreLock` and refactoring `DatastoreLock` to use a typed error.

I wasn't sure if it's worth the effort.

packages/owl-bot/src/owl-bot.ts
bcoe: At some point typescript changed the type of caught errors to `unknown` or `any`, which is kind of annoying.

chingor13 (approved):

PR #3438, opened 2022-04-14
feat: throw typed error

bcoe:
switch to throwing typed error.


chingor13 (approved):

chingor13 (changes requested):
nit: can we add the `@throws` to the jsdoc?

chingor13 (approved):

PR #3444, opened 2022-04-14
build: use dedicated service accounts

tmatsuo:
for the following 4 bots:
- generated-files-bot
- header-checker-lint
- merge-on-green
- policy

The policy-bot is using its own service account because it needs
BigQuery permission.

part of #3424


tmatsuo:
Needs an internal cl to get merged first

bcoe (approved):

PR #3459, opened 2022-04-18
feat: instantiate new child logger for each request

### LINKED ISSUE #3421: process(gcf-utils): review logger code for concurrent requests
Since Cloud Run accepts concurrent requests within a single instance, we should review gcf-utils' logger code in terms of thread(?) safety.

chingor13:
probot does not expose a way to provide a per-request logger to the event handler function. Instead, we inject our own into the payload and extract within the handler.

```ts
import {getContextLogger} from 'gcf-utils';
// ...

app.on('event', async context => {
  const logger = getContextLogger(context);
  logger.info('some message');  
});
```

You can still `import {logger} from 'gcf-utils'` which will get you the root logger, but it will not have any context values set.

Note: `context.log` can be set/configured (and [we are](https://github.com/googleapis/repo-automation-bots/pull/3459/files#diff-2d10ce1b796b0a4b6f0cbcfb880b8921d11a4710fe6af0f89c85c1abf8ce64fdR340)), but it cannot be modified per-request.

Fixes #3421

### DIFF
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -480,9 +480,14 @@ export class GCFBootstrapper {
         return;
       }
 
-      const requestLogger = logger.child(
-        buildTriggerInfo(triggerType, id, name, request.body)
-      );
+      /**
+       * Note: any logs written before resetting bindings may contain
+       * bindings from previous executions
+       */
+      const triggerInfo = buildTriggerInfo(triggerType, id, name, request.body);
+      logger.resetBindings();
+      logger.addBindings(triggerInfo);
+      const requestLogger = logger.child(triggerInfo);
       try {
         if (triggerType === TriggerType.UNKNOWN) {
           response.sendStatus(400);
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -534,6 +534,7 @@ export class GCFBootstrapper {
             return;
           }
 
+          setContextLogger(payload, requestLogger);
           // TODO: find out the best way to get this type, and whether we can
           // keep using a custom event name.
           await this.probot.receive({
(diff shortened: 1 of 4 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -482,12 +480,9 @@ export class GCFBootstrapper {
         return;
       }
 
-      /**
-       * Note: any logs written before resetting bindings may contain
-       * bindings from previous executions
-       */
-      logger.resetBindings();
```
tmatsuo, line 489: Can we somehow preserve this behavior and make these removals opt-in?

Then we can safely roll out this version without breaking anything.

chingor13: That's an option. I was envisioning releasing as a major version bump, then as you upgrade gcf-utils, you'd start grabbing the context logger in the dep upgrade PR.

tmatsuo: I think our deps updater script will blindly create a PR without knowing the change, so we should let the oncall know about this.

packages/gcf-utils/src/logging/trigger-info-builder.ts
```diff
@@ -67,18 +67,15 @@ export function buildTriggerInfo(
...
-      trigger_sender: requestBody.sender?.login || UNKNOWN,
-      payload_hash: getPayloadHash(requestBody),
-      github_event_type: getEventTypeDetails(
-        githubEventName,
-        requestBody.action
-      ),
-    };
-    triggerInfo.trigger = {...webhookProperties, ...triggerInfo.trigger};
-  }
+  // if (triggerType === TriggerType.GITHUB) {
```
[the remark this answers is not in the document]
chingor13: The payloads for http requests from Cloud Tasks can have the trigger data (source repo/event) - this is actually why the tests are failing (which can be fixed or we can make this a follow up PR).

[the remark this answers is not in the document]
chingor13: Restored the original behavior, but we can enhance in a follow-up

packages/gcf-utils/src/gcf-utils.ts
chingor13: Restoring the original (buggy) behavior, but keeps it from being a breaking behavior change. In the future, when we migrate loggers in bots, we can remove this reset logic.

tmatsuo:
As far as I read the code, I think we can correlate the logs with `triggerInfo.trigger.github_delivery_guid` in a Cloud Run backend with concurrent requests.

Is my understanding correct?

tmatsuo (approved):

chingor13:
Going to go with the weakmap version. It feels cleaner. We will update a single bot first and watch memory usage.

PR #3462, opened 2022-04-19
fix(flakybot): revert back to the default service account

### LINKED ISSUE #3461: flakybot is showing strange status code
Log:

Function execution took 9 ms. Finished with status: response error

tmatsuo:
hopefully fixes #3461


tmatsuo:
Hm, the error seemed to start happening before the last deployment

PR #3499, opened 2022-04-25
fix: use yargs default instead of coerce

### LINKED ISSUE #3498: release-branch is not loading github-token from environment
```typescript
      .option('github-token', {
        describe: 'GitHub access token',
        type: 'string',
        coerce: arg => {
          return arg || process.env.GITHUB_TOKEN || process.env.GH_TOKEN;
        },
        demand: true,
      })
```

`coerce` is not being run before `demand`.

chingor13:
We were relying on a bug from yargs - coerce is only run if the argument is provided. Instead, we use `default` which can accept a callback function (although the type definition is incorrect - we hack around it with some casting).

Fixes #3498


danielbankhead (approved):
LGTM - the `function -> unknown -> string` is a little annoying to see, but totally understand

bcoe (approved):
This is gross, I'll do my best to fix the upstream type ASAP.

PR #3505, opened 2022-04-25
fix: bump bazel-bot timeout to 3 hours

chingor13:
We are hitting an issue where we need to rebuild every API and it's taking over 2 hours


bcoe (approved):
looks like a bigger number to me.

PR #3509, opened 2022-04-26
feat: add timestamps to log statements

SurferJeffAtGoogle:
to see where scan-googleapis-gen-and-create-pull-requests is using its time.

### DIFF
--- a/packages/owl-bot/src/logger.ts
+++ b/packages/owl-bot/src/logger.ts
@@ -1,4 +1,4 @@
-// Copyright 2021 Google LLC
+// Copyright 2022 Google LLC
 //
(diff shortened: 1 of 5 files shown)


packages/owl-bot/src/logger.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright 2021 Google LLC
```
chingor13, line 1: To appease the header linter:

```suggestion
// Copyright 2022 Google LLC
```

tmatsuo: Well, this file is not auto generated. Changing copyring header seems wrong. I'd fix the bot, but is it difficult?

chingor13 (approved):

PR #3517, opened 2022-04-27
build: create GCB test that ensures the image builds

chingor13:
Future PRs touching the serverless-scheduler-proxy directory will trigger this build as a presubmit


chingor13:
This test can be extended in the future to test the built container further

tmatsuo (approved):

PR #3526, opened 2022-04-27
feat: inject trace logging attribute for request correlation

chingor13:


packages/gcf-utils/src/bot-request.ts
```diff
@@ -0,0 +1,120 @@
...
+  signature?: string;
+  taskName?: string;
+  taskRetryCount?: number;
+  traceId?: string;
+  triggerType: TriggerType;
+}
+
+export function parseBotRequest(request: express.Request): BotRequest {
+  const eventName =
+    request.get('x-github-event') || request.get('X-GitHub-Event') || '';
```
bcoe, line 41: Test, but this should be case-insensitive:

> Returns the specified HTTP request header field (case-insensitive match). The Referrer and Referer fields are interchangeable.

https://expressjs.com/en/5x/api.html

packages/gcf-utils/src/bot-request.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+/**
+ * Parse the trace id from the trace context header. The format of the header
+ * looks something like `<trace-id>/<span-id>;o=<options-flags>`.
+ * @param request incoming trigger request
+ */
+function parseTraceId(request: express.Request): string | undefined {
+  const traceContext =
+    request.get('X-Cloud-Trace-Context') ||
```
bcoe, line 108: Again, might as well make our logic simpler.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -571,6 +499,24 @@ export class GCFBootstrapper {
     };
   }
 
+  private buildLoggerBindings(
+    botRequest: BotRequest,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    requestBody: {[key: string]: any}
```
bcoe, line 505: Does express expose a better type for `body`?

packages/gcf-utils/test/bot-request.ts
```diff
@@ -0,0 +1,154 @@
...
+
+import {describe, it} from 'mocha';
+import assert from 'assert';
+import * as express from 'express';
+import {parseBotRequest, TriggerType} from '../src/bot-request';
+
+describe('parseBotRequest', () => {
+  describe('with all headers', () => {
+    const headers = {
+      'x-github-event': 'issues',
```
bcoe, line 23: I would randomize the case a bit in the headers, to confirm that request doesn't care.

bcoe (changes requested):
Left some suggestions, like the refactor.

tmatsuo (approved):

packages/gcf-utils/test/bot-request.ts
chingor13: Looks like the header getter is case insensitive, but the normalization happens sometime in the building of the request object.

packages/gcf-utils/src/gcf-utils.ts
chingor13: I don't believe so - it depends on the shape of the data and how it's decoded. In this case, we've JSON.parse'd the body string somehow with a filter.

packages/gcf-utils/test/bot-request.ts
chingor13: The header accessors are now referencing the upper camel-cased versions of the headers and the test is using all lowercase now.

bcoe (approved):

PR #3528, opened 2022-04-27
feat: require tagging titles with [autoapprove] for owlbot template changes

sofisl:
cc @parthea


packages/auto-approve/src/process-checks/owl-bot-template-changes.ts
```diff
@@ -63,10 +66,11 @@ export class OwlBotTemplateChanges extends Process implements LanguageRule {
...
 
-    const titleMatches = checkTitleOrBody(
-      this.incomingPR.title,
-      this.classRule.titleRegex
-    );
+    const titleMatches =
+      // We don't want it to include fix, feat, or !
+      !checkTitleOrBody(this.incomingPR.title, this.classRule.titleRegex[0]) &&
+      // We do want it to include [autoapprove] in title
+      checkTitleOrBody(this.incomingPR.title, this.classRule.titleRegex[1]);
```
chingor13, line 73: Rather than relying on the position of the regex for exclusion, can we add a new config for excludeTitleRegex or something like that?

chingor13 (changes requested):

packages/auto-approve/src/process-checks/owl-bot-template-changes.ts
```diff
@@ -52,7 +53,11 @@ export class OwlBotTemplateChanges extends Process implements LanguageRule {
     ),
       (this.classRule = {
         author: 'gcf-owl-bot[bot]',
-        titleRegex: /(fix|feat|!)/,
+        // For this particular rule, we want to check a pattern and an antipattern;
+        // we want fix/feat/! to not be in the title, and we do want [autoapprove] to
+        // be in the title
+        titleRegex: /\[autoapprove\]/,
+        titleRegexExclude: /(fix|feat|!)/,
```
chingor13, line 60: Does this mean that all configurations of this rule require bot a match and a negative match?

sofisl: Basically it means that any template changes in synthtool can't be breaking or fixes/feats, and have to be marked in the title with [autoapprove]. In other words, we're just adding that the PRs should be opt-in, i.e., you need to specify [autoapprove] in it.

chingor13 (approved):

Issue #3524, opened 2022-04-27
update gcf-utils and migrate to context logger


chingor13:
- [x] auto-approve
- [x] auto-label
- [x] blunderbuss
- [x] canary-bot
- [x] cherry-pick-bot
- [x] conventional-commit-lint
- [x] do-not-merge
- [x] failurechecker
- [x] flakybot
- [x] generated-files-bot
- [x] header-checker-lint
- [x] label-sync
- [x] merge-on-green
- [x] owl-bot
- [x] policy
- [x] release-please
- [x] release-trigger
- [x] repo-metadata-lint
- [x] snippet-bot
- [x] sync-repo-settings
- [x] trusted-contribution

PR #3550, opened 2022-04-29
fix(release-trigger): update gcf-utils and use context-logger

chingor13:
Towards #3524


tmatsuo (approved):

chingor13:
> Should `packages-release-trigger-test` be required?

We can probably make all the cloud build checks required assuming it allows for skipped tests

PR #3590, opened 2022-05-02
deps: update gcf-utils to 13.5.2

chingor13:
Updates previously updated bots from gcf-utils 13.5.1 to 13.5.2 to fix trigger information for Cloud Task requests


tmatsuo (approved):

chingor13:
Another case of #3985

PR #3641, opened 2022-05-09
build(release-trigger): add minimal docker frontend image

### LINKED ISSUE #3352: release-trigger front-end still running out of memory
2022-03-29 around 11am, Java made a bunch of releases. release-trigger front-end ran out of memory and crashed.

From the logs:
```
Memory limit of 953M exceeded with 966M used. Consider increasing the memory limit, see https://cloud.google.com/run/docs/configuring/memory-limits
```

```
"The request failed because either the HTTP response was malformed or connection to the instance had an error. Additional troubleshooting documentation can be found at: https://cloud.google.com/run/docs/troubleshooting#timeout-503
While handling this request, the container instance was found to be using too much memory and was terminated. This is likely to cause a new container instance to be used for the next request to this revision. If you see this message frequently, you may have a memory leak in your code or may need more memory. Consider creating a new revision with more memory."
```

The memory graph does not appear to have a memory leak:

AVdYgnUEvdpN7aF

chingor13:
Release-trigger already has a split frontend/backend that shared the same docker image. We are occasionally seeing some memory issues on the frontend, so we will switch the frontend to use a smaller docker image.

Additionally, we switch the frontend to use the web-frontend service account which has less permissions than the backend requires.

Maybe fixes #3352


packages/release-trigger/cloudbuild.yaml
```diff
@@ -45,15 +47,31 @@ steps:
...
+      - "-f"
+      - "Dockerfile.frontend"
+      - "-t"
+      - "gcr.io/$PROJECT_ID/release-trigger-frontend"
+      - "."
+
+  - name: gcr.io/cloud-builders/docker
+    id: "push-docker-frontend"
+    waitFor: ["build-docker-frontend"]
+    args: ["push", "gcr.io/$PROJECT_ID/release-trigger"]
```
tmatsuo, line 65: It needs to be `release-trigger-frontend`

packages/release-trigger/cloudbuild.yaml
```diff
@@ -45,15 +47,31 @@ steps:
...
+
   - name: gcr.io/cloud-builders/gcloud
-    id: "deploy-backend"
-    waitFor: ["push-docker"]
+    id: "deploy-frontend"
+    waitFor: ["push-docker-frontend"]
     entrypoint: bash
     env:
-      - "SERVICE_ACCOUNT=autorelease@repo-automation-bots.iam.gserviceaccount.com"
+      - "IMAGE_NAME=gcr.io/$PROJECT_ID/owl-bot-frontend"
```
tmatsuo, line 72: Actually it will likely work, but it should be `release-trigger-frontend`.

tmatsuo (changes requested):

packages/release-trigger/cloudbuild.yaml
chingor13: Thanks, good catch

packages/release-trigger/cloudbuild.yaml
chingor13: Thanks, good catch.

packages/release-trigger/cloudbuild-test.yaml
```diff
@@ -33,3 +33,15 @@ steps:
...
+  - name: gcr.io/cloud-builders/docker
+    id: "build-docker-frontend"
+    waitFor: ["-"]
+    dir: packages/release-trigger
+    args:
+      - "build"
+      - "-f"
+      - "Dockerfile.frontend"
+      - "-t"
+      - "gcr.io/$PROJECT_ID/release-trigger"
```
tmatsuo, line 46: Not a big deal, but maybe `release-trigger-frontend`?

tmatsuo (approved):

PR #3669, opened 2022-05-12
fix: force label now attempts releases

chingor13:
Currently, the release-please bot only attempts releases on a push to main. There's no way to attempt a retry. After this PR, adding the `release-please: force-run` label will also attempt to tag a release.


packages/release-please/src/release-please.ts
```diff
@@ -225,6 +225,64 @@ async function buildManifest(
...
+          repoLanguage,
+          branchConfiguration,
+          logger
+        );
+      }
+    } catch (e) {
+      if (e instanceof Errors.DuplicateReleaseError) {
+        // In the future, this could raise an issue against the
+        // installed repository
+        logger.warn('Release tag already exists, skipping...', e);
```
losalex, line 263: Perhaps I am missing something, but should you return here?

losalex (approved):
Left a comment - not blocking on it since I am not sure regarding correct behaviour

packages/release-please/src/release-please.ts
chingor13: No, this has 2 steps, attempting to tag releases (if any), then maybe opening a new release pull request. We are explicitly ignoring duplicate tag errors from tagging and will continue on to try opening new pull requests

bcoe (approved):

PR #3713, opened 2022-05-17
build(policy,snippet-bot): add test builds

tmatsuo:
I think sync-trigger will create two triggers and these tests will start running for upcoming PRs.


packages/policy/cloudbuild-test.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright 2021 Google LLC
```
chingor13, line 1: nit
```suggestion
# Copyright 2022 Google LLC
```

packages/snippet-bot/cloudbuild-test.yaml
```diff
@@ -0,0 +1,37 @@
+# Copyright 2020 Google LLC
```
chingor13, line 1:
```suggestion
# Copyright 2022 Google LLC
```

chingor13 (changes requested):

tmatsuo:
@chingor13 Done

chingor13 (approved):

PR #3770, opened 2022-05-23
fix: handle uncaught 404 exception as permission errors

### LINKED ISSUE #3643: auto-approve: Error 404 not found in get-pr-info.js getChangedFiles
This is triggering on https://github.com/GoogleCloudPlatform/bq-utilization-alerts/pull/37

It seems like the bot shouldn't even be running on this repo as it's not configured.

sofisl:
Fixes #3643


packages/auto-approve/src/process-checks/owl-bot-template-changes.ts
```diff
@@ -96,13 +96,13 @@ export class OwlBotTemplateChanges extends Process implements LanguageRule {
 
     reportIndividualChecks(
       ['authorshipMatches', 'titleMatches', 'bodyMatches', 'isGAPIC'],
-      [authorshipMatches, !titleMatches, !bodyMatches, isGAPIC],
+      [authorshipMatches, titleMatches, !bodyMatches, isGAPIC],
```
sofisl, line 99: Technically this is an unrelated change, just a little bug I noticed with logging. We were reporting the inverse of what titleMatches should show

tmatsuo: I suggest you separate it into 2 PRs.

sofisl: `reportIndividualChecks` only relates to logging, so it won't break anything to make this change. I would prefer to bundle it given that it would only be a single character change, if that's alright.

tmatsuo: I still think you should separate.

It's hard to revert just one part.

packages/auto-approve/test/auto-approve.test.ts
```diff
@@ -68,6 +68,17 @@ function getReviewsCompleted(owner: string, repo: string, response: Reviews[]) {
     .reply(200, response);
 }
 
+function listFiles(
+  owner: string,
+  repo: string,
+  code: number,
+  response: File[] | undefined
```
chingor13, line 75: I don't see where the `File` type is defined.

packages/auto-approve/src/process-checks/owl-bot-template-changes.ts
chingor13: I agree on this point too.

packages/auto-approve/src/get-pr-info.ts
```diff
@@ -53,6 +53,7 @@ export async function getChangedFiles(
       logger.error(
         `Not found error, ${err.status}, ${err.message} for ${owner}/${repo}/${prNumber}`
       );
+      return undefined;
```
chingor13, line 56: Is this a setup issue for the bot (not requesting enough permissions)? Shouldn't it have permissions to read the code in the PR?

chingor13 (changes requested):

packages/auto-approve/src/get-pr-info.ts
sofisl: I think some repos in the org have more restrictions than others, the current set of permissions seem to be enough for the repos that use auto-approve.

chingor13 (approved):

PR #3771, opened 2022-05-23
fix: throw instance of Error instead of class

### LINKED ISSUE #3645: auto-approve: AutoApproveNotConfigured error thrown
We should probably be catching this error and handling it

https://console.cloud.google.com/errors/detail/CPa-oOb15orfzwE?project=repo-automation-bots

sofisl:
Fixes #3645

### DIFF
--- a/packages/auto-approve/src/check-config.ts
+++ b/packages/auto-approve/src/check-config.ts
@@ -135,12 +135,12 @@ export async function checkAutoApproveConfig(
       } else {
         // This branch means auto-approve is not on this repo, so we're
         // throwing an error (essentially, skipping the check)
-        throw AutoApproveNotConfigured;
+        throw new AutoApproveNotConfigured();
       }
     } catch (err) {
       // This branch means auto-approve is not on this repo, so we're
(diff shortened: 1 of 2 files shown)


packages/auto-approve/src/check-config.ts
```diff
@@ -135,12 +135,12 @@ export async function checkAutoApproveConfig(
...
         // This branch means auto-approve is not on this repo, so we're
         // throwing an error (essentially, skipping the check)
-        throw AutoApproveNotConfigured;
+        throw new AutoApproveNotConfigured();
       }
     } catch (err) {
       // This branch means auto-approve is not on this repo, so we're
       // throwing an error (essentially, skipping the check)
-      throw AutoApproveNotConfigured;
+      throw new AutoApproveNotConfigured();
```
chingor13, line 143: Can we add a test to make sure this is being appropriately caught?

sofisl: added!

chingor13 (approved):

Issue #3734, opened 2022-05-20
Release-trigger bot does not trigger releases properly


losalex:
As you can see from https://github.com/googleapis/nodejs-logging/issues/1281, the release was failing to trigger and only manual release re-triggering helps. Seems like a bug - please take a look

bcoe:
Another example where I had to kick off the job manually:

https://github.com/googleapis/nodejs-run/pull/2

chingor13:
From looking at the logs, I don't see any failures. I do see the `release.created` events and it doesn't say it even attempted a release (there's no output of other failure cases either).

Adding some logging to output the number of found, pending release PRs. Also, adding a cron to periodically scan for the missing pull requests as a fallback. We will be able to dig in further later.

chingor13:
I think this is a race condition between release-please creating the release (which triggers the release-trigger bot) and it labeling the PR with `autorelease: tagged`. This bot looks for merged PRs with the `autorelease: tagged` label.

#3774 should add additional logging - a `release.published` event log expects to see a log message that it found a release to trigger.

chingor13:
This is a race condition - the `autorelease: tagged` label is not yet applied to pull request by the time the `release.created` event triggers.

PR #3774, opened 2022-05-25
fix: run release-trigger on a schedule to catch failed webhooks

chingor13:
chore: add logging for the number of found release pull requests

Towards #3734


chingor13:
I forgot that the scheduler can't target Cloud Run services yet

chingor13:
Updated the `LAUNCH_DATE` constant to avoid picking up old merged PRs that might now be picked up by the cron.

bcoe (approved):

PR #3783, opened 2022-05-26
feat: allow targeting Cloud Run URLs with crons

### LINKED ISSUE #2143: serverless-scheduler-proxy: should be able to proxy to a Cloud Run instance
See #1817

Cron payloads look something like:
```json
{
  "Name":"canary_bot",
  "Type":"function",
  "Location":"us-central1"
}
```

chingor13:
Cron payloads have always had a `Type` field which was hard-coded to `function`. To set the cron payload, you set the `TARGET_TYPE=run` env variable when running the cron_deploy Cloud Build step.

After deploying this and releasing a new cron-utils package, we will test with canary bot's crons.

Went with option 1b from #2143 - the service identifier is consistent within our project and we don't need to target other projects.

Unblocks #3774 

Fixes #2143


serverless-scheduler-proxy/cloudbuild.yaml
```diff
@@ -35,6 +35,8 @@ steps:
     id: "serverless-scheduler-proxy-deploy"
     waitFor: ["serverless-scheduler-proxy-push"]
     entrypoint: bash
+    env:
+      - "SERVICE_IDENTIFIER=c7mcaca6sa"
```
tmatsuo, line 39: It seems like the same identifier is shared between Cloud Run deployments. Is it guaranteed?

chingor13: In some docs it's referred to as the "project hash". From all the examples I see online, it's static for your project, but I don't see any official declaration on the product docs.

tmatsuo (approved):

serverless-scheduler-proxy/cloudbuild.yaml
tmatsuo: Good for now. Maybe give feedback to the product team?

PR #3786, opened 2022-05-27
build(canary-bot): target cloud run with cron

chingor13:
See #2143
Unblock #3774 

fix: cloud run urls have `-uc` in them.
build: allow overriding the service name in the cron-deploy script

### DIFF
--- a/packages/canary-bot/cloudbuild.yaml
+++ b/packages/canary-bot/cloudbuild.yaml
@@ -43,6 +43,8 @@ steps:
     id: "cron-deploy"
     waitFor: ["-"]
     entrypoint: bash
+    env:
+      - "TARGET_TYPE=run"
     args:
       - "-e"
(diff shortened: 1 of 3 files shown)


chingor13:
This isn't quite working yet. Manually updated the trigger to test it and it was proxying to `canary_bot-<project-hash>.a.run.app`. We need to be able to configure the Cloud Run service name too (not just infer from the directory name).

chingor13:
This one should be ready now

tmatsuo (approved):

chingor13:
> What does uc stand for?

It's actually unclear if the `-uc` is part of the project hash or not. This current configuration of the environment variable and this template string will generate the correct URL for our Cloud Run services. We could adjust the template string and environment variable later if the `-uc` is actually part of the project hash.

PR #3791, opened 2022-05-27
fix: add Datastore lock

### LINKED ISSUE #3602: bug(auto-approve): auto approve bot didn't add the automerge label
In PR https://github.com/googleapis/python-apigee-connect/pull/111, auto-approve bot approved the PR but did not add the `automerge: exact` label.  The label was added successfully in https://github.com/googleapis/python-access-approval/pull/198.

sofisl:
Fixes #3602 
@tmatsuo, I'm attempting to solve an apparent race condition when adding labels between auto-approve and auto-label. I'm adding Datastore lock to auto-approve; should I also add it to auto-label?


SurferJeffAtGoogle (commented):
Strategy check:

It appears possible to update a pull request atomically to eliminate the race condition.  This strategy would avoid using datastore, and avoid some inconveniences of datastore:
- separate credentials required to talk to datastore
- another point of failure
- a lock could be held for a long time if something (crash, network issues, etc) prevented the holder from unlocking it.

To update the PR atomically:
1.  Get the PR and observe its `etag` header.
2. Specify an `If-Match` header when updating the pull request.
3. If the PR has changed between getting it and updating it, then the update will fail, and the caller can try again.

Relevant docs:
https://tools.ietf.org/id/draft-dusseault-http-patch-13.html#patch
https://stackoverflow.com/questions/28060116/which-is-more-reliable-for-github-api-conditional-requests-etag-or-last-modifie

tmatsuo:
I agree with @SurferJeffAtGoogle

sofisl:
@SurferJeffAtGoogle, thanks for the suggestion!  Implemented in #4028

PR #3881, opened 2022-06-09
feat(gcf-utils): add flow control for scheduling jobs

### LINKED ISSUE #3880: gcf-utils: flow control for scheduler jobs
For "repository" type cron jobs, gcf-utils fans out Cloud Task jobs very quickly. This results in a burst of instance count in the backend service and also those services are facing the secondary rate limit from Github.

I think we can add some flow control for mitigating this.

tmatsuo:
fixes #3880


tmatsuo:
> Implementation-wise this LGTM, however, should we make it opt-in and/or configurable? We do already have hundreds of installed repositories for a few bots and if we're not careful syncing up the delay with the cron schedule, we could end up with overlapping runs of tasks (maybe it's okay though).

Currently all the bots are just queuing tasks at the same time. This PR will improve things.

re: opt-in and configuration
I don't see the benefit of making it configurable because I don't see a problem when those batches overwrap among bots. Our bots have their own backend service and they should be fine.

chingor13 (approved):

PR #3933, opened 2022-06-13
fix: when logging errors, log all errors contained in an AggregateError

### LINKED ISSUE #3819: Bot errors appear to be grouped incorrectly in Cloud Error Reporting
We get grouped errors like:

9cfyjZocRLK75JY

But when you click through on the error group, you see other error types included.

7H4vfvJifoPoPQb

My hunch is that Error Reporting is confused because these errors are actually `AggregateError` types from probot. The `jsonPayload` for there errors actually looks like:

BHFSRqRg6yjcx4Q

and:

55uv6CzkhfkjdKz

Both have the same `name` and first line of the `stack`.

chingor13:
Probot can run multiple handlers for a single event. It wraps any thrown errors in an `AggregateError`. When handling an `AggregateError`, log each contained error so Error Reporting will pick up each one individually.

This should be a huge quality of life fix for our on-call to see and investigate bot errors.

Fixes #3819


packages/gcf-utils/src/logging/gcf-logger.ts
```diff
@@ -232,3 +232,18 @@ function pinoLevelToCloudLoggingSeverity(
   const UNKNOWN_SEVERITY = 'DEFAULT';
   return {severity: severityMap[level] || UNKNOWN_SEVERITY, level: level};
 }
+
```
SurferJeffAtGoogle, line 235: Could you add a comment to the GCFLogger class definition explaining why `console.log()` is insufficient?

SurferJeffAtGoogle (changes requested):

packages/gcf-utils/src/logging/gcf-logger.ts
chingor13: Done

SurferJeffAtGoogle (approved):

PR #3944, opened 2022-06-14
fix: small tweaks in production to get owlbot-bootstrapper running, pass GH token as env variable, redact GH token from logging

sofisl:
This PR includes:
- small changes in production to get the bootstrapper running (e.g., secret name, regexes)
- passing in the short-lived GH access token to an environment variable
- redacting any GH tokens from error messages


tmatsuo:
It seems like there are some implementation changes too.

I suggest you either
- separate the PR
- include the other changes in the PR description

SurferJeffAtGoogle (approved):

PR #4028, opened 2022-06-23
fix: add etag retry logic

### LINKED ISSUE #3602: bug(auto-approve): auto approve bot didn't add the automerge label

sofisl:
Fixes #3602


packages/auto-approve/src/auto-approve.ts
```diff
@@ -56,6 +56,56 @@ export async function authenticateWithSecret(
...
+ * @param repo string, the name of the repo of the incoming PR
+ * @param prNumber number, the number of the PR
+ * @param octokit the octokit instance
+ */
+export async function retryAddLabel(
+  retryCount: number,
+  owner: string,
+  repo: string,
+  prNumber: number,
+  octokit: Octokit
```
chingor13, line 73: I would recommend injecting the GCFLogger instance into this function as an optional param that defaults to the global logger.
```suggestion
  octokit: Octokit,
  logger: GCFLogger = defaultLogger
```

Where `defaultLogger` is imported from `gcf-utils` like:
```ts
import {getContextLogger, logger as defaultLogger} from 'gcf-utils';
```

This will allow these log messages to keep the request context for easier filtering in the console.

chingor13 (changes requested):

packages/auto-approve/src/auto-approve.ts
```diff
@@ -56,6 +56,57 @@ export async function authenticateWithSecret(
...
+      labels: ['automerge: exact'],
+      // The comparison with the stored ETag for if-none-match uses the
+      // weak comparison algorithm, meaning two files are considered identical
+      // if the content is equivalent, not identical.
+      // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match
+      headers: {'if-none-match': `'${etag}'`},
+    });
+  } catch (err) {
+    if ((err as any).status === 412 && retryCount < 3) {
+      setTimeout(() => 2000);
```
bcoe, line 98: This should be `awaited`, however I don't know that you should even need to worry about adding a sleep (I would think the update that caused the collision updating would already be complete).

bcoe (changes requested):

packages/auto-approve/src/auto-approve.ts
```diff
@@ -56,6 +56,56 @@ export async function authenticateWithSecret(
...
+ * Grabs an etag for a PR, and retries adding a label if etag is different
+ *
+ * @param retryCount the amount of retries for adding a label
+ * @param owner string, of the repo of the incoming PR
+ * @param repo string, the name of the repo of the incoming PR
+ * @param prNumber number, the number of the PR
+ * @param octokit the octokit instance
+ */
+export async function retryAddLabel(
+  retryCount: number,
```
chingor13, line 69: A more common pattern is to provide the number of retries remaining, rather than the retry attempt number. A private helper might keep the current attempt count as an argument. It's a bit weird for the caller to always provide the 0 argument on the initial call and have no control over the number of attempts.

chingor13 (changes requested):

bcoe (approved):

chingor13 (approved):

PR #4033, opened 2022-06-29
feat(owlbot-bootstrapper): add PR copy-tag info, and API ID info in PR title

sofisl:
See: https://github.com/googleapis/google-cloud-node/pull/3206

This PR allows owlbot-bootstrapper to fill in the Copy-Tag header for the PRs. The Copy-Tag header comprises of the .OwlBot.yaml path, and the last commit of googleapis-gen. Since the container doesn't know where the .OwlBot.yaml lives, it must receive it as an argument from the language-specific container. The language-specific container will write the .OwlBot.yaml path to the interContainerVars.json file, where the common-container can grab it from.


packages/owlbot-bootstrapper/common-container/test/split-repo.test.ts
```diff
@@ -149,9 +155,14 @@ describe('SplitRepo class', async () => {
...
-    const scope = nock('https://api.github.com')
-      .post('/repos/googleapis/fakeRepo/pulls')
-      .reply(201);
+    const scopes = [
+      nock('https://api.github.com')
+        .get('/repos/googleapis/googleapis-gen/commits')
+        .reply(201, {sha: '6dcb09b5b57875f334f61aebed695e2e4193db5e'}),
+      nock('https://api.github.com')
+        .post('/repos/googleapis/fakeRepo/pulls')
+        .reply(201),
```
chingor13, line 164: you can chain these together if they use the same host, rather than creating an array yourself.

chingor13 (approved):

packages/owlbot-bootstrapper/common-container/utils.ts
```diff
@@ -18,13 +18,15 @@ import {uuid} from 'uuidv4';
 import {Octokit} from '@octokit/rest';
 import * as fs from 'fs';
 import {Language} from './interfaces';
+import {dir} from 'console';
 
-const BRANCH_NAME_FILE = 'branchName.md';
+export const BRANCH_NAME_FILE = 'branchName.md';
```
chingor13, line 23: Do we foresee more variables being passed between these containers? Perhaps creating a defined JSON or textproto formatted metadata file would be more robust than several individual files.

chingor13 (changes requested):
Can you document somewhere the communication between containers? This forms the contract/interface of what the language containers need to provide in order to work. Ideally, this is forward compatible so that we don't have to change N language containers every time we add a feature to the common container.

packages/owlbot-bootstrapper/common-container/utils.ts
```diff
@@ -131,21 +172,24 @@ export async function openAnIssue(
...
-      fs
-        .readFileSync(`${directoryPath}/${BRANCH_NAME_FILE}`)
-        .toString()
-        // Need to remove whitespace from branch name
-        .replace(/\s+/g, '')
+    console.log(
+      fs.readFileSync(`${directoryPath}/${fileNameJSON}`).toString('utf8')
+    );
+    return JSON.parse(
+      fs.readFileSync(`${directoryPath}/${fileNameJSON}`).toString()
```
chingor13, line 192: Might be worth considering setting default values in case the file is not present (not necessary in this PR)

chingor13 (approved):

PR #4042, opened 2022-07-06
fix: do not check for intercontainervars before it exists

sofisl:


chingor13 (changes requested):
Can you add a test and/or describe what bug this is fixing?

Issue #3869, opened 2022-06-06
sync-repo-settings: Getting "YAMLException: bad indentation of a mapping entry"


tmatsuo:
Log:

```
 YAMLException: bad indentation of a mapping entry (35:8)
     32 |   isAdminEnforced: false
     33 |   requiresStrictStatusChecks: false
     34 |   requiredStatusCheckContexts:
     35 |     - 'cla/google'
    -------------^
     36 |     - 'snippet-bot check'
     37 |     - 'header-check'
        at generateError (/workspace/node_modules/js-yaml/lib/loader.js:183:10)
        at throwError (/workspace/node_modules/js-yaml/lib/loader.js:187:9)
        at readBlockMapping (/workspace/node_modules/js-yaml/lib/loader.js:1182:7)
        at composeNode (/workspace/node_modules/js-yaml/lib/loader.js:1441:12)
        at readBlockSequence (/workspace/node_modules/js-yaml/lib/loader.js:1008:5)
        at composeNode (/workspace/node_modules/js-yaml/lib/loader.js:1472:45)
        at readBlockMapping (/workspace/node_modules/js-yaml/lib/loader.js:1164:11)
        at composeNode (/workspace/node_modules/js-yaml/lib/loader.js:1441:12)
        at readDocument (/workspace/node_modules/js-yaml/lib/loader.js:1625:3)
        at loadDocuments (/workspace/node_modules/js-yaml/lib/loader.js:1688:5) 
```

PR #4050, opened 2022-07-07
feat: add new issue-utils package

chingor13:
Extracts `addOrUpdateIssue` into a new helper utility library. This was extracted from `release-please` which is using this to notify installed repositories of configuration errors.

Example PR: https://github.com/googleapis/google-cloud-java/issues/7948

Towards #3869

### DIFF
--- a/packages/issue-utils/src/issue-utils.ts
+++ b/packages/issue-utils/src/issue-utils.ts
@@ -1,0 +1,148 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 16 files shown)


packages/issue-utils/src/issue-utils.ts
```diff
@@ -0,0 +1,148 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// eslint-disable-next-line node/no-extraneous-import
+import {Octokit} from '@octokit/rest';
+import {GCFLogger, logger as defaultLogger} from 'gcf-utils';
+
+export interface Issue {
```
tmatsuo, line 19: Not a big deal, but is there a reason to have our own interface instead of the type defined in one of the Octokit suites?
Maybe add a comment about it?

tmatsuo (approved):

packages/issue-utils/src/issue-utils.ts
chingor13: I'll add a comment in a follow up. The reason is because relying on the third-party times can be prone to fragility. Additionally their types can be weird (e.g. `labels: (string | {name?: string})[];`) which forces you to write more defensive code. We coerce the types to make it easier to work with.

PR #4074, opened 2022-07-19
feat: implement owlbot:copy-code label

### LINKED ISSUE #4072: FR(owlbot): make sure owlbot can regenerate a PR that is creating a new .OwlBot.yaml file
When attempting to regenerate this PR: https://github.com/googleapis/google-cloud-node/pull/3229, owlbot fails with the message: ```Error: ENOENT: no such file or directory, open '/tmp/tmp-1-vhWQfLkPW6aq/dest/packages/google-cloud-kms/.github/.OwlBot.yaml'``` because that file does not yet exist on `main`, it only exists on the branch.  See build: https://pantheon.corp.google.com/cloud-build/builds;region=global/3b276573-48ed-4fab-9168-c516897302b8;step=5?project=repo-automation-bots

We should make sure owlbot can regenerate a PR even if .OwlBot.yaml is not on main yet.

- [ ] @SurferJeffAtGoogle to refactor owlbot
- [ ] @sofisl to refactor owlbot-bootstrapper (adding a label owlbot:copy-code)
- [ ] @sofisl to add label to labelsync

SurferJeffAtGoogle:
Owl Bot Bootstrapper labels a pull request with 'owlbot:copy-code' to request Owl Bot to copy the source code from googleapis-gen.  The pull request must have a copy tag.

While adding this feature, I removed the obsolete command line options source-repo-commit-hash and owl-bot-yaml-path.  Owl bot already inspects the copy-tag to learn these two details.  So they were obsolete before this pull request.  It was easier to remove them than to update them for the new code path.

Fixes #4072

### DIFF
--- a/packages/owl-bot/src/copy-code.ts
+++ b/packages/owl-bot/src/copy-code.ts
@@ -541,24 +541,26 @@ command in a local clone of this repo:
   }
 }
 
+export type CopyCodeIntoPullRequestAction =
+  | 'regenerate' // Discard all current commits and regenerate.
+  | 'append'; // Copy code and into a new commit and append it to the PR.
+
 /**
- * Regenerates a pull request.
- * Uses `git push -f` to completely replace the existing contents of the branch.
+ * Appends or regenerates a pull request.
  *
- * This is quite complicated because the PR may have multiple open commits.
- * Regenerating the pull request will result in a single commit.  We need
- * to preserve the full history of all the commits.
- *
- * In time, this will completely replace copyCodeIntoPullRequest().
+ * Regenerating is quite complicated because the PR may have multiple open
+ * commits.  Regenerating the pull request will result in a single commit,
+ * but we need to preserve the full history of all the commits.
  *
  * @param sourceRepo: the source repository, either a local path or googleapis/googleapis-gen
  * @param destRepo: the destination repository, either a local path or a github path like googleapis/nodejs-vision.
  */
-export async function regeneratePullRequest(
+export async function copyCodeIntoPullRequest(
   sourceRepo: string,
   destRepo: GithubRepo,
   destBranch: string,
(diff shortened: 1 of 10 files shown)


packages/owl-bot/src/copy-code.ts
```diff
@@ -663,7 +664,20 @@ export async function regeneratePullRequest(
 
   const sourceRepoCommitHash = copyTags[0].h;
 
-  cmd(`git checkout -b ${destBranch}`, {cwd: destDir});
+  if (action === 'regenerate') {
```
sofisl, line 667: Should both line 667 and 732 be action === regenerate? If so, why is there a type for 'append' (what does its code path look like)? More generally, what determines whether the action is regenerate or append, since there's only one label (I'm assuming)?

SurferJeffAtGoogle: > Should both line 667 and 732 be action === regenerate?

Yes.

> If so, why is there a type for 'append' (what does its code path look like)?  More generally, what determines whether the action is regenerate or append, since there's only one label (I'm assuming)?

There are two code paths through `copyCodeIntoPullRequest()`.

The first path is the path that existed before this PR, when the user checks the "Regenerate this pull request checkbox."  `copyCodeIntoPullRequest(action==="regenerate")` should behave exactly as `regeneratePullRequest()` did before this PR.  The tests were very useful to confirm this.

The new code path is for when Bootstrapper applies the label `owlbot:copy-code`.  In that case,  Owl Bot invokes `copyCodeIntoPullRequest(action==="append")`.  There are two important differences in the path:
1.  Instead copying the code into a copy of the main branch, Owl Bot copies the code into the PR's branch.
2.  Instead of rewriting the PR's title and body, it leaves them unchanged.

sofisl (approved):

PR #4085, opened 2022-07-25
docs(sync-repo-settings): describe all supported options

### LINKED ISSUE #4081: [sync-repo-settings]: support "requiresLinearHistory" in branchProtectionRule
Branch protection rules allow maintainers to require a linear commit history, or not. Requiring it disallows the use of Merge Commits. While Merge Commits are generally not encouraged in `googleapis` repos, they are idiomatic means of keeping a branch in sync with another. So if we are working a feature branch that we want to stay in sync with `main`, we must use a Merge commit with the changes from `main` into a feature branch. By default `require linear history` is enabled, which disallows merge commits on the branch. I'd like to be able to disable it on a per-branch basis.
image

chingor13:
These options were already in use and described in https://github.com/googleapis/repo-automation-bots/blob/main/packages/sync-repo-settings/src/schema.json, but were missing in the README

Fixes #4081

### DIFF
--- a/packages/sync-repo-settings/README.md
+++ b/packages/sync-repo-settings/README.md
@@ -27,6 +27,9 @@ branchProtectionRules:
 # Identifies the protection rule pattern. Name of the branch to be protected.
 # Defaults to `master`
 - pattern: master
+  # Will new commits pushed to matching branches dissmiss pull request review approvals.
+  # Defaults to `false`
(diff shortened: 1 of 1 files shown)


chingor13:
cc @noahdietz

noahdietz:
Thank you @chingor13 :)

tmatsuo (approved):

PR #4092, opened 2022-07-27
refactor: reimplement failurechecker with failure messages

### LINKED ISSUE #4067: [failurechecker] An informative message upon release failure
Feedback from: @ansh0l 

Recent spanner release failed: https://github.com/googleapis/java-spanner/issues/1948

- It didn't give Kokoro job link to troubleshoot. A message to check "prod:cloud-devrel/client-libraries/java/java-spanner/release/stage" would be nice.
- It should be marked as P1 or higher to handle the release failure quickly

chingor13:
This is a major refactor of the code to make it easier to follow and to be able to return the reason for a failure. Since each publication job is different, we can only report on the general expected state.

We also switch to use `@google-automations/issue-utils` to help manage the issue

Fixes #4067


packages/failurechecker/src/failurechecker.ts
```diff
@@ -51,148 +60,208 @@ export const TimeMethods = {
...
+        } labels = ${releasePullRequest.labels.join(',')}`
+      );
+      failedReleases.push({
+        number: releasePullRequest.number,
+        reason: 'The release job failed -- check the build log.',
+      });
+    }
+
+    // Look for in-progress releases - these usually failed to start the next
+    // step in the pipeline, or failed to report status back
```
tmatsuo, line 139: Question: How do we make sure failures are really failures, not a real in-progress?

chingor13: There's a check for the minimum time (3 hours). If the PR is newer than 3 hours, we ignore.

tmatsuo: Ah ok, i would add that in the comment

chingor13: Added to the comment.

tmatsuo (approved):

Issue #4096, opened 2022-08-01
mono-repo-publish should be able to publish documentation


chingor13:
Currently, mono-repo-publish can only publish to npm. We would also like to be able to conditionally publish docs.

PR #4097, opened 2022-08-01
refactor: convert mono-repo-publish to typescript

chingor13:
Converts mono-repo-publish to typescript for maintainability

Towards #4096


packages/mono-repo-publish/src/main.ts
```diff
@@ -12,74 +12,107 @@
...
 }
 
-function getOctokitInstance () {
-  const appId = readFileSync(process.env.APP_ID_PATH, 'utf8');
-  const privateKey = readFileSync(process.env.GITHUB_PRIVATE_KEY_PATH, 'utf8');
-  const installationId = readFileSync(process.env.INSTALLATION_ID_PATH, 'utf8');
+export function getOctokitInstance(
+  appIdPath: string,
+  privateKeyPath: string,
+  installationIdPath: string
```
chingor13, line 48: change: made these explicit - read the env vars at the edge of the code (the bin script)

packages/mono-repo-publish/src/bin/mono-repo-publish.ts
```diff
@@ -0,0 +1,80 @@
...
+        demand: true,
+      })
+      .option('dry-run', {
+        describe: 'whether or not to publish in dry run',
+        type: 'boolean',
+        default: false,
+      });
+  },
+  async handler(argv) {
+    const appIdPath = process.env.APP_ID_PATH;
```
sofisl, line 43: What if, instead of using these as arguments, we grab them from a secret in repo-automaiton-bots, like the rest of our bots?

chingor13: Let's follow up on that feature outside of this PR. I did this part because of typing and flow (didn't want to mess with the error handling in `getOctokitInstance()`)

sofisl (approved):

PR #4115, opened 2022-08-08
feat(conventional-commit-lint): support configuration file

bcoe:
Add support for configuration file, which allows automation to be turned off.

### DIFF
--- a/packages/conventional-commit-lint/src/conventional-commit-lint.ts
+++ b/packages/conventional-commit-lint/src/conventional-commit-lint.ts
@@ -85,7 +85,7 @@ export = (app: Probot) => {
           owner,
           repo,
           CONFIGURATION_FILE,
-          {},
+          {}
         );
       } catch (e) {
         const err = e as Error;
--- a/packages/conventional-commit-lint/src/conventional-commit-lint.ts
+++ b/packages/conventional-commit-lint/src/conventional-commit-lint.ts
@@ -86,7 +86,6 @@ export = (app: Probot) => {
           repo,
           CONFIGURATION_FILE,
           {},
-          {schema: schema}
         );
       } catch (e) {
         const err = e as Error;
--- a/packages/conventional-commit-lint/src/conventional-commit-lint.ts
+++ b/packages/conventional-commit-lint/src/conventional-commit-lint.ts
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {getConfigWithDefault} from '@google-automations/bot-config-utils';
+import {getConfigWithDefault, ConfigChecker} from '@google-automations/bot-config-utils';
+
 // eslint-disable-next-line node/no-extraneous-import
 import {Probot} from 'probot';
 import {PullRequest} from '@octokit/webhooks-types';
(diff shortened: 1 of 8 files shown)


bcoe:
I'm going to update this PR with config validation, before we merge.

packages/conventional-commit-lint/src/conventional-commit-lint.ts
```diff
@@ -38,6 +46,31 @@ export = (app: Probot) => {
...
+      // opt in. For this reason config is loa
+      let config: Configuration | undefined = undefined;
+      try {
+        config = await getConfigWithDefault<Configuration>(
+          context.octokit,
+          owner,
+          repo,
+          CONFIGURATION_FILE,
+          {},
+          {schema: schema}
```
chingor13, line 60: I'm of the opinion that we should not throw an error here if the schema doesn't match. Let the presubmit check that you're going to add surface the error. Alternatively, use `@google-automations/issue-utils` to open an issue on the repository that the config is invalid (otherwise, to the repo owner, you don't know there's an error).

Example: if the user erroneously adds a config file with extra entries, then this whole event handler does not run.

bcoe: > I'm of the opinion that we should not throw an error here if the schema doesn't match. Let the presubmit check that you're going to add surface the error.

I was intending to let the failed check surface config format errors, the `try`/`catch` is to catch other types of errors (missing permissions in bot).

tmatsuo: I still think `log.error` is not right because as a bot, it is the correct behavior.

bcoe: > I still think log.error is not right because as a bot, it is the correct behavior.

I agree, I've switched to `logger.warn`.

tmatsuo: Update: Ah sorry I somehow think the exception is thrown when the schema is wrong.

I re-read bot-config-utils and realized the exception is thrown only when something bad happens.

tmatsuo: Ok, `warn` also works for me

packages/conventional-commit-lint/src/conventional-commit-lint.ts
```diff
@@ -38,6 +50,56 @@ export = (app: Probot) => {
...
+      // opt in. For this reason config is loaded with default values.
+      let config: Configuration | undefined = undefined;
+      try {
+        config = await getConfigWithDefault<Configuration>(
+          context.octokit,
+          owner,
+          repo,
+          CONFIGURATION_FILE,
+          {},
+          {schema: schema}
```
chingor13, line 89: I think if I had a config like:

```yaml
enabled: false
foo: bar
```

Then `getConfigWithDefault` will throw an error and `config` would not be set. In this case, the bot would still run the linter. Is this the desired outcome?

tmatsuo: Oh @chingor13 is right

FYI: The error is [`InvalidConfigurationFormat`](https://github.com/googleapis/repo-automation-bots/blob/4ffc71b49c286ea0322fc76945574d1d82b5fa0d/packages/bot-config-utils/src/bot-config-utils.ts#L438)

bcoe: I dropped the `{schema: schema}`, I think it's sufficient that we validate the configuration as they're added to repos.

chingor13 (approved):

PR #4117, opened 2022-08-08
feat(bot-config-utils)!: validateConfigChanges now resolve boolean for valid config

bcoe:
The `Checker` class had been updated, but not the public interface.

The method signature of the public interface changes, so calling this a breaking change.


tmatsuo (approved):
I don't think changing Promise<void> to Promise<boolean> is a breaking change though.

bcoe:
> Update: Ah ok, some tests will fail.

@tmatsuo it's easy to fix the tests, but I figured it would create confusion when tests simply start failing on main post release.

PR #4125, opened 2022-08-09
feat(conventional-commit-lint)!: enforce subject case

bcoe:
Technical writers have requested that CHANGELOG notes begin each line with a
capital letter.

---

Introduces title case rule (which can be turned off) based on technical writer feedback. We should not merge this until we've announced these additional requirements to stakeholders (I have an email I will be sending).


packages/conventional-commit-lint/README.md
```diff
@@ -17,9 +17,10 @@ npm start
 The following configuration settings can be set by creating a
 `.github/conventional-commit-lint.yaml` file:
 
-| Name         | Description                        | Default              |
-|--------------|------------------------------------|----------------------|
-| enabled      | Should commit linting be enabled?  | `boolean` = `true`   |
+| Name         | Description                                     | Default              |
+|--------------|-------------------------------------------------|----------------------|
+| enabled      | Should commit linting be enabled?               | `boolean` = `true`   |
+| titleCase    | Should commit description begin with a capital? | `boolean` = `true`   |
```
chingor13, line 23: nit: I'd probably name this as an action like `enforceTitleCase`

chingor13 (changes requested):
Is there a way to acknowledge without changing case? Consider a commit like `fix: gRPC connection can configure keep-alive timeout`. Perhaps some kind of annotation that signifies that I know what I'm doing with my casing.

PR #4132, opened 2022-08-10
fix(gcf-utils)!: update @octokit/rest to 19

tmatsuo:
I think this is a breaking change, also dropping node 12 support.

I introduced a new helper function `getAuthenticatedOctokit` for providing the newest Octokit from `@octokit/rest`.


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -151,6 +151,71 @@ export interface CronPayload {
...
+  installationId: number | null
+): Promise<Octokit> {
+  const botSecrets = await getBotSecrets();
+  if (installationId === null) {
+    return new Octokit({
+      authStragety: createAppAuth,
+      auth: {
+        appId: botSecrets.appId,
+        privateKey: botSecrets.privateKey,
+        clientSecret: botSecrets.secret,
```
chingor13, line 205: I think the stored secret is actually the webhook shared secret, not an oauth client secret

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -151,6 +151,71 @@ export interface CronPayload {
...
+    privateKey: secrets.privateKey,
+    appId: secrets.appId,
+    secret: secrets.secret,
+  };
+}
+
+/**
+ * A helper for getting an Octokit instance authenticated as an App.
+ */
+export async function getAuthenticatedOctokit(
```
chingor13, line 195: Should the Bootstrapper class call this one instead of it's instance method?

tmatsuo: Not sure

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: Yeah I don't know how to authenticate it as an app. Now it's throwing an error.

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: I could not make it work without the installation id.

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: I figured out how to authenticate it as an app, but that version doesn't take care of token expiration

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: Update: I finally make it work without the installation id

tmatsuo: But it doesn't automatically renew the token upon expiration and the expiration seems to be 10 minutes. It might be long enough for our use case.

I'll add some tests in the canary bot and if it works, I'll change Bootstrapper to use this version.

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: I finally made it all right, I think.

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: Update: I hope the current version will handle token expiration

chingor13 (commented):
This seems ok, but a little weird. If a bot developer uses, `context.octokit`, they'll get a v18 octokit instance (from probot), but if they use `getAuthenticatedOctokit()`, they get a v19 octokit instance?

tmatsuo:
Yes it's weird, but this will make the `@octokit/rest` version controllable within our software stack (apps and libraries), without making a major change (like replacing probot).

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -151,6 +151,78 @@ export interface CronPayload {
   cron_org: string;
 }
 
+export interface BotSecrets {
+  privateKey: string;
+  appId: string;
+  secret: string;
```
chingor13, line 157: nit: let's rename this `webhookSecret` for clarity

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -151,6 +151,78 @@ export interface CronPayload {
...
+}
+
+/**
+ * A helper for fetch secret from SecretManager.
+ */
+async function getBotSecrets(): Promise<BotSecrets> {
```
chingor13, line 163: Do we want this method and `BotSecrets` exported publicly? This file is the main file for this library so someone can do:
```ts
 import {getBotSecrets} from `gcf-utils`;
```

If it's not intended to be publicly exported (exported for testing), let's move into a separate file.

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -151,6 +151,78 @@ export interface CronPayload {
...
+    delete secrets.cert;
+  }
+  if (Object.prototype.hasOwnProperty.call(secrets, 'id')) {
+    secrets.appId = secrets.id;
+    delete secrets.id;
+  }
+  return {
+    privateKey: secrets.privateKey,
+    appId: secrets.appId,
+    secret: secrets.secret,
```
chingor13, line 188: Along with my requested renaming above, we could check for `secrets.webhookSecret ?? secrets.secret` for backwards and forwards compatibility (allow putting the the value in either `secret` or `webhookSecret` in the stored JSON contents).

chingor13 (changes requested):

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: Exporting sounds good

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: Done

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: Done

tmatsuo:
@chingor13 Thanks, PTAL

chingor13 (approved):

PR #4138, opened 2022-08-12
feat(owlbot-bootstrapper): add default cli args

sofisl:


packages/owlbot-bootstrapper/cli/helper.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export class Helper {
+  args: CliArgs;
+  languageContainers: {
+    language: string;
+    languageContainer: string;
+    repoToClone: string;
+  }[];
+
+  constructor(args: CliArgs) {
```
chingor13, line 26: I would not import this type from the CLI (it couples the 2 pieces of code), instead declare the type within this module. If the types match, then you can just pass the parsed args object to this constructor.

packages/owlbot-bootstrapper/cli/helper.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    // If there was never a match, and there was no language container specified in the CLI, we should throw
+    if (!this.args['languageContainer']) {
+      throw new Error('No language-specific container specified.');
+    }
+
+    return undefined;
+  }
+
+  public async getFullLatestContainerName(container: string) {
```
chingor13, line 56: this method does not need to be an instance method -- it doesn't require any of the object's properties. Consider making it a static method or a separate exported helper.

packages/owlbot-bootstrapper/cli/helper.ts
```diff
@@ -0,0 +1,67 @@
...
+    const latestSha = Buffer.from(
+      execSync(
+        `gcloud container images describe ${container} --format="value(image_summary.digest)"`
```
chingor13, line 59: Do we need to shell out to get this data? It'd be preferable to do this in code rather than requiring the runtime to have the `gcloud` CLI installed. Consider using the container registry API client.

chingor13 (changes requested):

packages/owlbot-bootstrapper/cli/helper.ts
sofisl: This helper is in the cli though, these would have to be coupled regardless (it would be the same type). What if we instead created an interfaces file that exported the type to the various files that needed it in cli/?

packages/owlbot-bootstrapper/cli/helper.ts
sofisl: I wrote it this way so that it was easier to test, or else you can't mock this function while testing. I did a fair bit of experimentation with the testing, so let me know if you see another way to test this additional functionality without making it a part of a helper class.

packages/owlbot-bootstrapper/cli/helper.ts
sofisl: I didn't see that as [an option](https://cloud.google.com/container-registry/docs/managing). I think it's because container registry is being [deprecated](https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr), even though it's the default for pushing up docker images to gcp. I agree that it would be better to not have the gcloud CLI installed but I am not sure there's another way, feel free to point to other documentation if there is any!

packages/owlbot-bootstrapper/cli/helper.ts
chingor13: Sinon can mock static methods (`sandbox.stub(Helper, 'getFullLatestContainerName').resolves(something)` and other exported functions (`import * as helperModule from 'helper'; sandbox.stub(helperModule, 'someExportedFunction').resolves(something);`)

packages/owlbot-bootstrapper/cli/helper.ts
chingor13: The logic class should define the interface and its caller should use that interface. The caller (CLI) could also declare its own type and if it matches the logic class' interface then that's fine. Or alternatively, if the shared interface is truly shared (something that makes sense like something that actually is a concept in the domain), then it could be declared separately and shared.

chingor13: In this case, the only things this class needs are the `language` and `languageContainer` fields. So we should declare only those arguments. You can still pass the entire `CliArgs` object to this constructor.

packages/owlbot-bootstrapper/cli/helper.ts
chingor13: Ew, I didn't realize that Container Registry only provides a docker API (it would be possible to use a generic docker API client library with added GCP credential headers, but ugh). I have a strong aversion to node libraries shelling out to command line tools and parsing outputs unless absolutely necessary.

sofisl:
@chingor13, refactored the code due to moving to artifact registry, cloud build actually doesn't need a sha if the docker image is in AR, eliminating the call to it. Simplifies the code quite a bit

chingor13 (approved):

Issue #1885, opened 2021-05-25
Migrate bots to Node 14


sofisl:
We should update bots to Node 14 to update dependencies. But, we need to be careful when upgrading bots that it does not break when deploying to GCF.

tmatsuo:
Note: gcf-utils has node 10 as the engine param. Maybe it doesn't matter, but maybe we want to change it too?

Also, can we make sure with canary bot first?

bcoe:
I noticed on Friday this behavior:

1. I tried to switch owlbot to Node 14 through the UI.
2. the deployment failed, so the Cloud Function was not actually upgraded.

Later that day I tried editing code to test a bug, and noticed the exact same issue, **without upgrading to Node 14.**.

This makes me think that the bug I'm bumping into is with deploying using the UI, not the upgrade to Node 14.

## Alternate way to upgrade

Perhaps we can instead pass in the Node version as a parameter to the `gcloud` command we use to deploy? With it defaulting to Node 10.

> gcf-utils has node 10 as the engine param

Once we switch everything over to 14, I think we can update the engines' field post-hoc, Node 14 should be backwards compatible to Node 10.

tmatsuo:
I'd like to still start with canary bot.
"It should be fine" and "It's proven to be working" is a different thing.

bcoe:
> It should be fine" and "It's proven to be working" is a different thing.

I agree, I would rather we update at least a few of the bots one at a time, to make sure the upgrade is smooth, before we make the swap widely -- as the runtime is a significant change.

tmatsuo:
canary bot works with Cloud Function node14 runtime.

PR #4149, opened 2022-08-15
deps: use gcf-utils 14

tmatsuo:
for auto-approve and auto-label

part of #1885

### DIFF
--- a/packages/auto-approve/src/auto-approve.ts
+++ b/packages/auto-approve/src/auto-approve.ts
@@ -16,7 +16,12 @@
 
 // eslint-disable-next-line node/no-extraneous-import
 import {Probot, Context} from 'probot';
-import {GCFLogger, getContextLogger, logger as defaultLogger} from 'gcf-utils';
+import {
+  GCFLogger,
+  getAuthenticatedOctokit,
+  getContextLogger,
+  logger as defaultLogger
+} from 'gcf-utils';
 import {checkPRAgainstConfig} from './check-pr';
 import {checkPRAgainstConfigV2} from './check-pr-v2';
 import {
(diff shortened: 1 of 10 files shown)


tmatsuo:
@chingor13 Does the cron payload always have `installation.id`?

sofisl (approved):

packages/auto-approve/src/auto-approve.ts
```diff
@@ -222,8 +229,12 @@ export function handler(app: Probot) {
         return;
       }
 
+      const octokit = await getAuthenticatedOctokit(
+        context.payload.installation!.id
```
sofisl, line 233: I think this will fail when installation ID doesn't exist, which we don't want to have [happen](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator).  I think we want to attempt to authenticate only if the installation ID exists.

tmatsuo: It's required for compilation because the schema (or interface) claims it's an optional field, but there should always be the installation id in webhooks that the bot receives.

Without attempting to authenticate, we cannot compile `getChangedFiles` because `context.octokit` is the `Octokit` from v18, but we need `Octokit` from v19.

What do you suggest otherwise?

sofisl: We should probably add a branch like:

```
let octokit;
if (installation.id) {
      octokit = await getAuthenticatedOctokit(
        context.payload.installation!.id
       )
} else {
    throw new Error('Installation ID not provided in ${context.payload.event}; cannot authenticate for getting files`)
}
```

sofisl (changes requested):
Let's handle the error if we do not get an installation ID

packages/auto-approve/src/auto-approve.ts
tmatsuo: It will likely be dead code, but somehow installation id is missing it will give a better error message.

tmatsuo: Done!

tmatsuo:
@sofisl Thanks for the review, PTAL

sofisl (approved):

sofisl:
NB, it would be nice to add a test for the case when an installation ID is not provided.

PR #4156, opened 2022-08-16
fix(auto-label): throws an Error when installation id is not provided

tmatsuo:
This is a followup on #4149


sofisl (approved):
Thank you!

Issue #4157, opened 2022-08-17
Release please not proposing a release PR for google-cloud-java monorepo 


ddixit14:
Trying to implement release-please for [google-cloud-java ](https://github.com/googleapis/google-cloud-java) but release please is not proposing a PR after "releasable units" are committed to the default branch. CLI states "timeout".

suztomo:
@ddixit14 Add your observations. Screenshots and suspicious log messages.

ddixit14:
@chingor13 (Adding observations that might be of some use)

**Timeout Error logs:**
https://cloudlogging.app.goo.gl/c4XxyKBy6eSZQHpV9
https://cloudlogging.app.goo.gl/edLJ3F2pYArBUAW4A

**Logs for one invocation** - https://drive.google.com/file/d/1khFoyzzOOfwkDa3Yd7Oqkiy9cNAoLvBI/view?usp=sharing&resourcekey=0-VWI9rSoX2qXZXWGGmXDi8Q

After looking at the logs, only 83/126 [modules](https://github.com/googleapis/google-cloud-java/blob/monorepo_script_output/.release-please-manifest.json#L2-L83) are getting processed before the timeout (approx 9 minutes). Doing an approximation, if the timeout is increased to 15 mins, it should be able to process all the modules.

tmatsuo:
Here is an excerpt from the log:

> message: "Unknown error: {"data":null,"errors":[{"message":"Something went wrong while executing your query. This may be the result of a timeout, or it could be a GitHub bug. Please include `F64D:8CDF:650E49:133FA82:62FC3CC6` when reporting this issue."}]}: Unknown error: {"data":null,"errors":[{"message":"Something went wrong while executing your query. This may be the result of a timeout, or it could be a GitHub bug. Please include `F64D:8CDF:650E49:133FA82:62FC3CC6` when reporting this issue."}]}"
name: "GitHubAPIError"
stack: "HttpError: Unknown error: {"data":null,"errors":[{"message":"Something went wrong while executing your query. This may be the result of a timeout, or it could be a GitHub bug. Please include `F64D:8CDF:650E49:133FA82:62FC3CC6` when reporting this issue."}]}
    at /workspace/node_modules/@octokit/request/dist-node/index.js:86:21
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async sendRequestWithRetries (/workspace/node_modules/@octokit/auth-app/dist-node/index.js:376:12)
[...]

PR #4160, opened 2022-08-17
build: run release-please app on Cloud Run

chingor13:
Towards #4157

### DIFF
--- a/packages/release-please/src/server.ts
+++ b/packages/release-please/src/server.ts
@@ -12,22 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// eslint-disable-next-line node/no-extraneous-import
-import {Probot} from 'probot';
 import {GCFBootstrapper} from 'gcf-utils';
+import {api} from './release-please';
 
 const bootstrap = new GCFBootstrapper({
-  taskTargetEnvironment: 'functions',
+  taskTargetEnvironment: 'run',
 });
-
-// We only need to deploy gcf-utils in the frontend server because
-// only thing it does is to enqueue a task for the release-please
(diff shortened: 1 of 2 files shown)


packages/release-please/src/server.ts
```diff
@@ -12,22 +12,13 @@
...
 
 const bootstrap = new GCFBootstrapper({
-  taskTargetEnvironment: 'functions',
+  taskTargetEnvironment: 'run',
 });
-
-// We only need to deploy gcf-utils in the frontend server because
-// only thing it does is to enqueue a task for the release-please
-// backend. Thus we deploy a dummy empty bot code.
-
```
tmatsuo, line 26: So are you planning to use this file for both frontend and backend?

tmatsuo (approved):

tmatsuo (changes requested):
I think you also need to modify the cloudbuild.yaml.

packages/release-please/src/server.ts
chingor13: Yes, we are currently running a frontend, but we will have a single service that handles both.

chingor13:
> I think you also need to modify the cloudbuild.yaml.

We are already deploying it to GCF and Cloud Run. If this works as expected, I'll submit a follow up PR that stops building/deploying to GCF

tmatsuo:
I still recommend we have a Cloud Run frontend service **and** another Cloud Run backend service because those services have different scaling requirements.

packages/release-please/src/server.ts
chingor13: Now we are running separate frontend/backend.

PR #4164, opened 2022-08-18
deps(flakybot): update gcf-utils to v14

tmatsuo:
* build: drop node12 support

part of #1885

### DIFF
--- a/packages/flakybot/src/flakybot.ts
+++ b/packages/flakybot/src/flakybot.ts
@@ -29,7 +29,7 @@ import {DatastoreLock} from '@google-automations/datastore-lock';
 // eslint-disable-next-line node/no-extraneous-import
 import {Octokit} from '@octokit/rest';
 import {components} from '@octokit/openapi-types';
-import {getContextLogger, GCFLogger} from 'gcf-utils';
+import {getAuthenticatedOctokit, getContextLogger, GCFLogger} from 'gcf-utils';
 import {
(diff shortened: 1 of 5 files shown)


packages/flakybot/src/flakybot.ts
```diff
@@ -150,7 +150,16 @@ export function flakybot(app: Probot) {
   app.on('schedule.repository' as any, async context => {
     const owner = context.payload.organization.login;
     const repo = context.payload.repository.name;
-    await syncLabels(context.octokit, owner, repo, FLAKYBOT_LABELS);
+    let octokit: Octokit;
+    if (context.payload.installation && context.payload.installation.id) {
```
chingor13, line 154: Option for conciseness:
```suggestion
    if (context.payload.installation?.id) {
```

packages/flakybot/src/flakybot.ts
```diff
@@ -162,8 +171,19 @@ export function flakybot(app: Probot) {
     async context => {
       const configChecker = new ConfigChecker<Config>(schema, CONFIG_FILENAME);
       const {owner, repo} = context.repo();
+      let octokit: Octokit;
+      if (context.payload.installation && context.payload.installation.id) {
```
chingor13, line 175:
```suggestion
      if (context.payload.installation?.id) {
```

chingor13 (approved):

PR #4165, opened 2022-08-18
feat: add findFilesByFilename and findFilesByExtension

chingor13:
In release-please, we need the ability to find files by name and extension. We now add these 2 methods to the file cache so that we can share the git tree cache between the search and get contents APIs.

The general logic is:
1. fetch to root tree (recursive=true) to attempt to get all files in a single request
2. if we have all files (truncated=false), then run a find/filter on the list
3. if we do not have all files (truncated=true), then refetch the tree (recursive=false), and repeat this algorithm on each subdirectory (attempting recursive=true first, then doing our own recursion if that subdirectory has too many files).

This cache allows us to re-use our git tree API results and drastically reduce the number of API calls for large monorepos.

Implementation notes:
* we now cache an internal `CachedTree` interface which is the `TreeEntry[]` list and whether the cached result was from a `recursive=true` response. This signals to the caller whether or not we need to search deeper recursively ourselves
* we also implement an async generator `treeEntryIterator` which abstracts the cache usage and lets you iterate over files in the repository from a given base ref/SHA

Towards #4157


packages/git-file-utils/test/find-files.ts
```diff
@@ -0,0 +1,620 @@
...
+              'github-data-api/data-api-trees-successful-response-subdir-pkg-b'
+            ))
+          );
+        const files = await cache.findFilesByFilename('foo.json', 'pkg');
+        expect(files).lengthOf(2);
+        expect(files).to.eql(['a/foo.json', 'b/foo.json']);
+        req.done();
+      });
+
+      it('finds multiple files and strips prefix', async () => {
```
tmatsuo, line 265: nit: it actually finds just one file

tmatsuo (approved):

PR #4171, opened 2022-08-19
deps: update release-please to 14.1.0

chingor13:
Fixes #4157

PR #4173, opened 2022-08-19
feat: bot-config-utils can validate JSONSchema formats

chingor13:
JSON Schema supports `format` options for `string` types. For example, you can enforce that a string value looks like a URI.

We use the `ajv-formats` extension for `ajv` to also validate these types.

See https://github.com/googleapis/repo-automation-bots/pull/4171#issuecomment-1220998235


packages/bot-config-utils/test/fixtures/formats-wrong.json
```diff
@@ -0,0 +1,3 @@
+{
```
tmatsuo, line 1: Is this file used?

chingor13: No, I can remove. We had a JSON version of all the other fixtures too.

tmatsuo (approved):

PR #4306, opened 2022-09-01
fix(snippet-bot): use git-file-utils for fetching files

### LINKED ISSUE #2703: snippet-bot: use Git Data API for fetching files
If a PR has a large file, it gets an error like this:
```
HttpError: This API returns blobs up to 1 MB in size. The requested blob is too large to fetch via the API, but you can use the Git Data API to request blobs up to 100 MB in size.: {"resource":"Blob","field":"data","code":"too_large"}"
```

In this case, the bot ignores that file. This is probably ok because a file larger than 1MB likely doesn't contain any region tags.

However, it's still theoretically possible so it's better to migrate to Git Data API for fetching files.

Pragmatically I don't think we'll really need to fix this, so marking as p3. If we ever see such a file, I'll bump the priority to P2.

tmatsuo:
fixes #2703

### DIFF
--- a/packages/snippet-bot/src/snippet-bot.ts
+++ b/packages/snippet-bot/src/snippet-bot.ts
@@ -35,7 +35,6 @@ import {
   formatRegionTag,
   formatViolations,
   formatMatchingViolation,
-  isFile,
   downloadFile,
 } from './utils';
 import {invalidateCache} from './snippets';
@@ -50,7 +49,7 @@ import schema from './config-schema.json';
 import {ConfigChecker, getConfig} from '@google-automations/bot-config-utils';
 import {
   FileNotFoundError,
-  RepositoryFileCache
+  RepositoryFileCache,
 } from '@google-automations/git-file-utils';
 import {syncLabels} from '@google-automations/label-utils';
 import {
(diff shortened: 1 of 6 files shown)


packages/snippet-bot/src/snippet-bot.ts
```diff
@@ -301,14 +299,10 @@ async function scanPullRequest(
         tagsFound = true;
       }
     } catch (e) {
-      const err = e as RequestError & Error;
-      // Ignoring 403/404 errors.
-      if (err.status === 403 || err.status === 404) {
-        logger.info(
-          `ignoring 403/404 errors upon fetching ${file}: ${err.message}`
-        );
+      if (e instanceof FileNotFoundError) {
```
bcoe, line 302: 403s will now be fatal? Do we know why were ignoring 403s previously?

bcoe (approved):
This seems like a really nice cleanup (along with addressing the bug).

packages/snippet-bot/src/snippet-bot.ts
tmatsuo: I don't remember, but probably a mistake :)

PR #4353, opened 2022-09-06
chore: update renovate config

tmatsuo:
There are the following changes in this PR

* Separate PRs for each bots by `additionalBranchPrefix`
* Combine PRs for minor and patch updates into one
* Combine PRs for all the `@octokit` updates into one
* Enable dashboard
* Add stabilityDays = 30

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -19,9 +19,22 @@
     {
       "packageNames": ["google.golang.org/genproto"],
       "schedule": "after 7pm on monday"
+    },
+    {
+      "matchPackagePatterns": [
+        "*"
+      ],
+      "matchUpdateTypes": [
+        "minor",
+        "patch"
+      ],
+      "groupName": "all non-major dependencies",
+      "groupSlug": "all-minor-patch"
     }
   ],
   "lockFileMaintenance": {
     "enabled": true
   }
+  "additionalBranchPrefix": "{{parentDir}}-"
+  
 }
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -19,9 +19,22 @@
...
+    },
+    {
+      "matchPackagePatterns": [
+        "*"
+      ],
+      "matchUpdateTypes": [
+        "minor",
+        "patch"
+      ],
+      "groupName": "all non-major dependencies",
```
alexander-fenster, line 31: Renovate creates a branch in its fork and makes all updates for the given package group from that branch.
I tried it once in gapic-generator-typescript and after about a year of successful update, the branch eventually somehow got corrupted and I found no way in Renovate to force it recreate the branch. So if this happens here, we should be ready to rename the group name / slug to force it to switch to a new branch.

tmatsuo: Thanks for the info! We'll keep in mind.

renovate.json
```diff
@@ -2,9 +2,9 @@
   "extends": [
     "config:base",
```
SurferJeffAtGoogle, line 3: By keeping `config:base` here, won't we continue setting `group:monorepos`?  Is that intended?

tmatsuo: I have no intention actually, but I also think it won't cause any problems (not 100% sure though).

SurferJeffAtGoogle (approved):

PR #4382, opened 2022-09-12
feat(gcf-utils): switch to @google-cloud/run

chingor13:
Switch to the newly published Gapic Cloud Run Admin API client


packages/gcf-utils/test/gcf-bootstrapper.ts
```diff
@@ -1837,8 +1826,11 @@ describe('GCFBootstrapper', () => {
           },
         },
       });
-      runScope.done();
-      sinon.assert.calledOnce(adcStub);
+      // https://github.com/DefinitelyTyped/DefinitelyTyped/issues/36436
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      sinon.assert.calledOnceWithExactly(getServiceStub as any, {
```
tmatsuo, line 1831: I never experienced that I need to use `any` for this type of asserts.

tmatsuo (approved):

packages/gcf-utils/test/gcf-bootstrapper.ts
chingor13: There's a bug with sinon types and multiple type definitions

PR #4414, opened 2022-09-16
fix(owlbot-bootstrapper): drop support for split repo

parthea:


sofisl (approved):

chingor13 (changes requested):
@parthea The tests are not passing. Perhaps the tests were originally written so that python was explicitly not considered a monorepo?

parthea:
@sofisl Could we remove the split repo functionality?

chingor13 (approved):
I think this is ok. Confirmed with the Java team that we will want to generate the new libraries into their monorepo. PHP is the other repo that has extra repos, but those just need to be created and this was a separate path anyways.

Will wait for @sofisl to review as well.

sofisl (approved):

PR #4426, opened 2022-09-19
feat: add NESTED_COMMIT markers when combining multiple source commits

### LINKED ISSUE #3342: Owlbot multiple commit descriptions can miss release notes
This commit https://github.com/googleapis/nodejs-spanner/commit/d1a9d7c3da7c378b433dcdab459b0c46279a4fb2 has "fix: annotating some fields as REQUIRED" in the description. Apparently this commit has multiple "PiperOrigin-RevId:" associated with it and the first entry is "chore: update v2.14.1 gapic-generator-typescript".

Screen Shot 2022-03-23 at 1 00 22 PM

The commit was released as part of [v5.17.0](https://github.com/googleapis/nodejs-spanner/releases/tag/v5.17.0)

Screen Shot 2022-03-23 at 1 01 00 PM

However, the v5.17.0 release note https://github.com/googleapis/nodejs-spanner/releases/tag/v5.17.0 does not have the entry "fix: annotating some fields as REQUIRED".

Screen Shot 2022-03-23 at 1 03 39 PM

Would you troubleshoot and fix the problem?

CC: @bcoe

chingor13:
In https://github.com/googleapis/release-please/pull/1566, we added the ability to explicitly denote multiple commit messages by using `BEGIN_NESTED_COMMIT`/`END_NESTED_COMMIT` markers.

This PR now augments owl-bot to include those markers when combining multiple copy-code runs in a single PR. We add a new `prependCommitMessage` helper that encapsulates the logic of recombining a previous owl-bot copy PR, wrapping the previous message in a `BEGIN/END_NESTED_COMMIT` block and pre-pending the new commit message.

Note that any existing PR bodies will be grouped together into a single nested commit when the next copy code commit is added.

Fixes #3342


chingor13:
Blocked until #4427 is deployed

packages/owl-bot/__snapshots__/create-pr.js
```diff
@@ -0,0 +1,73 @@
+exports[
```
SurferJeffAtGoogle, line 1: I've never seen a __snapshots__ directory before.  What is it for?

packages/owl-bot/src/create-pr.ts
```diff
@@ -53,6 +58,56 @@ export function resplit(
   return {title, body: body.substring(0, MAX_BODY_LENGTH)};
 }
 
+const NESTED_COMMIT_SEPARATOR = 'BEGIN_NESTED_COMMIT';
+
+/**
+ * Given pull request content and a new commit message. Rewrite the pull request
+ * title and body. If the initial pull request title was truncated with ellipses,
```
SurferJeffAtGoogle, line 65: Rewrite it how?  Please add sample input and output to the comment.

packages/owl-bot/src/create-pr.ts
```diff
@@ -53,6 +58,56 @@ export function resplit(
...
+      : // there are already existing nested commit tags, content before the first
+        // one is wrapped in a new nested commit tag
+        `${NESTED_COMMIT_SEPARATOR}\n${
+          bodyParts[0]
+        }\nEND_NESTED_COMMIT\n${bodyParts
+          .slice(1)
+          .join(NESTED_COMMIT_SEPARATOR)}`;
+
+  // prepend the new commit message and use original title truncation logic
+  return resplit(
```
SurferJeffAtGoogle, line 105: Why not call prependCommitMessage before calling resplit?  Then resplit gets called only once and the unsplitting logic in this function can be removed.

packages/owl-bot/test/create-pr.ts
```diff
@@ -97,6 +99,40 @@ describe('resplit', () => {
...
+        const pullContent = resplit(
+          'feat: some feature\n\nadditional context',
+          withCheckbox
+        );
+        const prependedContent = prependCommitMessage(
+          'fix: some new feature\n\nmore additional context',
+          pullContent,
+          withCheckbox
+        );
+        snapshot(prependedContent.title);
```
SurferJeffAtGoogle, line 118: snapshot is hard to follow.  I'm much rather see two strings compared in line.

SurferJeffAtGoogle (commented):
Please add a flag to enable this new behavior.

Do the delimiters get added when a PR contains only a single commit?

packages/owl-bot/test/create-pr.ts
chingor13: The snapshots are easier to construct code-wise at the expense of the asserted value living in a different file. It is useful for large text bodies that are harder to read inline. I can switch this back to inline assertions though.

chingor13:
@SurferJeffAtGoogle I'm unsure how we want to opt-in this functionality. In code, I added an enum that is surfaced as a CLI argument, but I'm unsure how we want to configure an opt-in for a repository.

packages/owl-bot/test/create-pr.ts
SurferJeffAtGoogle: Thank you for offering to switch back to inline assertions.  Please do.

packages/owl-bot/src/create-pr.ts
SurferJeffAtGoogle: > We do want to rejoin that previous message to be nicer for the release notes.

Indeed, so why not call this function on the original message (before resplit)?

SurferJeffAtGoogle:
> I'm unsure how we want to opt-in this functionality. In code, I added an enum that is surfaced as a CLI argument, but I'm unsure how we want to configure an opt-in for a repository.

The CLI argument looks great!

SurferJeffAtGoogle:
Do the delimiters get added when a PR contains only a single commit?

chingor13:
> Do the delimiters get added when a PR contains only a single commit?

No, they are only added when we append to an existing PR

packages/owl-bot/src/copy-code.ts
```diff
@@ -372,14 +373,11 @@ async function findAndAppendPullRequest(
...
-    `${commitBody}\n\n` +
-      `${pull.title}\n` +
-      pull.body
-        ?.replace(REGENERATE_CHECKBOX_TEXT, '')
-        .replace(EMPTY_REGENERATE_CHECKBOX_TEXT, '')
-        .trim() ?? '',
-    WithRegenerateCheckbox.Yes
+  const {title, body} = prependCommitMessage(
+    commitBody,
+    {title: pull.title, body: pull.body || ''},
```
SurferJeffAtGoogle, line 378: Nit: the typescript idiom is `pull.body ?? ''`.

SurferJeffAtGoogle (approved):

meltsufin:
@SurferJeffAtGoogle @chingor13 This still doesn't seem to be working on the Java monorepo. Do we need to add any config?
Example: https://github.com/googleapis/google-cloud-java/pull/8614

PR #4431, opened 2022-09-19
build: fix auto-label config

chingor13:
For some reason we had 2 config files (the `.yml` one is not the correct file extension). This PR merges the configs into the correct file and fixes a few entries to match the labels that currently exist.


chingor13:
After this is merged, we should be able to delete all the `api: X` labels from this repository as they do not apply (and should not have been created).

tmatsuo (approved):

Issue #4412, opened 2022-09-16
[failurechecker] Can failurechecker close the issue when a new release succeeds?


suztomo:
This week is my release rotation for J-team. Many Release Please pull requests have been merged and I listed all issues with "recent release failed" in their title, to identify which repositories' releases failed.

<img width="1302" alt="Screen Shot 2022-09-15 at 9 48 18 PM" src="https://user-images.githubusercontent.com/28604/190538942-e4fd56a1-9d3a-4924-8f37-3f0a307addaf.png">

One by one, I check each repository to see whether a recent release succeeded or not. Mostly the issues were generated for the releases in past. I'm only interested in the releases that failed today or yesterday.

One by one, I close the issues after confirming recent releases in the repositories.

<img width="930" alt="Screen Shot 2022-09-15 at 9 43 19 PM" src="https://user-images.githubusercontent.com/28604/190538704-7d6c147d-584e-4f93-b5b8-edc06e6f8f69.png">

(It was too many. I gave up.)

I wish failurechecker closes the issues once a recent release succeeds.

Related: https://github.com/googleapis/repo-automation-bots/issues/3188

PR #4432, opened 2022-09-19
feat: add closeIssue helper

chingor13:
This helper uses the same heuristics to find an existing issue as the `addOrUpdateIssue` helper.

Towards #4412


packages/issue-utils/test/issue-utils.ts
```diff
@@ -122,3 +122,63 @@ describe('addOrUpdateIssue', () => {
...
+      )
+      .reply(200, [
+        {
+          number: 123,
+          title: 'my issue title',
+          body: 'my issue body',
+          labels: [],
+        },
+      ])
+      .patch('/repos/test-owner/test-repo/issues/123')
```
tmatsuo, line 147: Does it make sense to assert the patch request's body?

chingor13: Done.

tmatsuo (approved):

PR #4438, opened 2022-09-20
feat(gcf-utils): Add @type property for unhandled errors.

tmatsuo:
This should make Cloud Error Reporting great again.

### DIFF
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -1315,7 +1315,7 @@ function parseRateLimitError(e: Error): RateLimits | undefined {
  */
 export function logErrors(logger: GCFLogger, e: Error) {
   // Add "@type" bindings so that Cloud Error Reporting will capture these logs.
-  const bindings = logger.getBindings;
+  const bindings = logger.getBindings();
   if (bindings['@type'] !== ERROR_REPORTING_TYPE_NAME) {
     logger = logger.child({
       '@type': ERROR_REPORTING_TYPE_NAME,
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -1315,7 +1315,7 @@ function parseRateLimitError(e: Error): RateLimits | undefined {
  */
 export function logErrors(logger: GCFLogger, e: Error) {
   // Add "@type" bindings so that Cloud Error Reporting will capture these logs.
-  logger.addBindings({"@type": ERROR_REPORTING_TYPE_NAME});
+  logger.addBindings({'@type': ERROR_REPORTING_TYPE_NAME});
   if (e instanceof AggregateError) {
     for (const inner of e) {
       // AggregateError should not contain an AggregateError, but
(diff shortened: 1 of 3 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -1310,7 +1313,9 @@ function parseRateLimitError(e: Error): RateLimits | undefined {
  * @param {GCFLogger} logger The logger to log to
  * @param {Error} e The error to log
  */
-function logErrors(logger: GCFLogger, e: Error) {
+export function logErrors(logger: GCFLogger, e: Error) {
+  // Add "@type" bindings so that Cloud Error Reporting will capture these logs.
+  logger.addBindings({'@type': ERROR_REPORTING_TYPE_NAME});
```
chingor13, line 1318: This is modifying the logger instance and could cause unexpected behavior elsewhere. `logger.child(additionalBindings)` will give you a new logger instance with extra bindings (we may want to skip the recursion on line 1323 if we do that)

chingor13 (changes requested):

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: Now it's using `logger.child`

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -1310,7 +1313,15 @@ function parseRateLimitError(e: Error): RateLimits | undefined {
  * @param {GCFLogger} logger The logger to log to
  * @param {Error} e The error to log
  */
-function logErrors(logger: GCFLogger, e: Error) {
+export function logErrors(logger: GCFLogger, e: Error) {
+  // Add "@type" bindings so that Cloud Error Reporting will capture these logs.
+  const bindings = logger.getBindings;
```
chingor13, line 1318:
```suggestion
  const bindings = logger.getBindings();
```

packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -1310,7 +1313,15 @@ function parseRateLimitError(e: Error): RateLimits | undefined {
...
+  if (bindings['@type'] !== ERROR_REPORTING_TYPE_NAME) {
+    logger = logger.child({
+      '@type': ERROR_REPORTING_TYPE_NAME,
+      ...bindings,
```
chingor13, line 1322: Hmm, for some reason I thought it would should have maintained the existing bindings for you and merged in new entries, but reading the source, it doesn't appear that way.

chingor13 (changes requested):

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: Yeah I think logger.child has a bug (forgetting to merge parent's bindings)

chingor13 (approved):

packages/gcf-utils/src/gcf-utils.ts
tmatsuo: Filed #4439

PR #4441, opened 2022-09-21
chore: testing the new gcf-utils with canary

tmatsuo:
Now it throws intentionally for an issue titled `canary-bot error` .


chingor13 (approved):
Another option is to have an alternative logic branch. Perhaps something like if the title matches `canary-bot error`, it will throw an error.

tmatsuo:
Ah that might be a better idea

tmatsuo:
@chingor13 I added a new code block. PTAL

packages/canary-bot/test/test.ts
```diff
@@ -198,6 +198,16 @@ describe('canary-bot', () => {
       await probot.receive({name: 'issues', payload, id: 'abc123'});
       sinon.assert.notCalled(addOrUpdateIssueCommentStub);
     });
+    it('thorws an error for canary-bot error', async () => {
```
chingor13, line 201:
```suggestion
    it('throws an error for canary-bot error', async () => {
```

chingor13 (approved):

PR #4443, opened 2022-09-21
feat: introduce todo-bot

### LINKED ISSUE #1301: New Bot: TODO bot
Scan for instances of TODOs/FIXMEs in code and maintain issues on the repository.

Potentially assign to a person if tagged. Example: `TODO(chingor13): do something`

chingor13:
This is a proof of concept for scanning a repository for annotations (like `TODO` or `FIXME`). The bot does not do anything useful with the results of the scan.

Currently implemented is an `AnnotationScanner` class which encapsulates the logic of shallow cloning a repo and scanning for annotations and a CLI to test it out locally.

Fixes #1301


packages/todo-bot/src/scanner.ts
```diff
@@ -0,0 +1,132 @@
...
+  // TODO: example todo
+  // TODO(context): another example annotation
+  async findAnnotations(annotations: string[]): Promise<Annotation[]> {
+    const foundAnnotations: Annotation[] = [];
+    const workdir = await this.getWorkdir();
+    for (const annotation of annotations) {
+      const contextMatcher = new RegExp(
+        `${annotation}(?:\\((?<context>.*)\\))?:`
+      );
+      const {stdout} = await exec(`grep -rn -E '${annotation}(\\(.*\\))?:'`, {
```
tmatsuo, line 79: Is it possible not to use exec?

packages/todo-bot/src/scanner.ts
```diff
@@ -0,0 +1,132 @@
...
+        `git clone --depth=1 --branch=${this.branch} https://github.com/${this.owner}/${this.repo}.git ${this.workdir}`,
+        {logger: this.logger}
+      );
+    }
+    return this.workdir;
+  }
+
+  // TODO: example todo
+  // TODO(context): another example annotation
+  async findAnnotations(annotations: string[]): Promise<Annotation[]> {
```
tmatsuo, line 72: How about to make the interface receiving a regex and returning the match object?

That way it's more generic

tmatsuo (commented):
A generic scanner is useful if the interface is generic enough to support other use cases.
In that case, I think it's worth having it as a library.

packages/todo-bot/src/scanner.ts
chingor13: I also assume grep will be way more efficient in terms of speed and memory usage than we can possibly do in node.

PR #4445, opened 2022-09-21
refactor: use REST transport for GAPIC clients

alexander-fenster:
Warning ⚠️  - this PR might break things. Or not. Are the tests reliable enough? Let's see :)

Let's use REGAPIC (HTTP transport instead of gRPC) in all GAPIC clients (except owl-bot which is handled separately in #4446).

I personally verified Cloud Run works with REGAPIC. Other APIs should just work, but we'll likely be the first ones to dogfood REGAPIC with them.


chingor13 (changes requested):
Can we separate the Firestore client refactor(s) into separate PRs from the ones that just update the fallback? The rewrite ones seem more complicated.

alexander-fenster:
Sure, I'll move the owlbot changes

> Can we separate the Firestore client refactor(s) into separate PRs from the ones that just update the fallback? The rewrite ones seem more complicated.

Done! #4446

chingor13 (approved):

chingor13:
@sofisl Can you check the secret-rotator change?

tmatsuo:
Consider waiting for #4447 and #4448 because with those PRs, Cloud Error Reporting will capture unhandled errors.

alexander-fenster:
SG! I'll wait for those PRs, and it will give me some time to revert the secret rotator changes (they are needed because `@googleapis/iam` is outdated, and it will be released soon).

tmatsuo:
@alexander-fenster Those PRs are merged JFYI

sofisl (approved):

PR #4490, opened 2022-09-22
deps(release-trigger): update gcf-utils to V14.2.1

tmatsuo:
With this version, gcf-utils will cache Cloud Run URL.


tmatsuo:
gcf-utils 14.2.1 works for canary bot

chingor13 (approved):

PR #4496, opened 2022-09-23
feat(gcf-utils): introduce ServiceUnavailable

### LINKED ISSUE #4495: gcf-utils should provide a way for bots to return HTTP 503
Background:
We use Cloud Task push queue and if the handler returns HTTP 503, Cloud Task will add extra back off for next attempt.

Our bot should return HTTP 503 whenever it encounters its performance limitation to prevent catastrophic failure.

A natural implementation would be to introduce an Error type like `ServiceNotAvailable` and catch it then return HTTP 503.

tmatsuo:
A bot can throw ServiceUnavailable when it experiences resource limitation. The next task attempt will be delayed with some backoff.

fixes #4495

### DIFF
--- a/packages/gcf-utils/src/gcf-utils.ts
+++ b/packages/gcf-utils/src/gcf-utils.ts
@@ -613,7 +613,8 @@ export class GCFBootstrapper {
               let isServiceUnavailable = e instanceof ServiceUnavailable;
               let serviceUnavailableMessage =
                 e instanceof ServiceUnavailable ? e.message : '';
-              let serviceUnavailableStack = e instanceof ServiceUnavailable ? e.originalError.stack : '';
+              let serviceUnavailableStack =
+                e instanceof ServiceUnavailable ? e.originalError.stack : '';
               if (e instanceof AggregateError) {
                 for (const inner of e) {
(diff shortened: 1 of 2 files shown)


packages/gcf-utils/src/gcf-utils.ts
```diff
@@ -91,6 +91,13 @@ interface EnqueueTaskParams {
   name: string;
 }
 
+/**
+ * Bot can throw this error to indicate it experienced some form of
+ * resource limitation.  GCFBootstrapper will catch this and return
+ * HTTP 503 to suggest Cloud Task adds backoff for the next attempt.
+ */
+export class ServiceUnavailable extends Error {}
```
chingor13, line 99: Should we wrap the underlying cause (stack) in this exception? We get the base error message from the constructor, but the stack could be useful.

tmatsuo: Now you can preserve the original stack trace and it will be in the log.

tmatsuo:
@chingor13 Thanks, PTAL

chingor13 (approved):

PR #4505, opened 2022-09-28
feat(owlbot-bootstrapper): get apis.json file from drift and save it to a well-known location

sofisl:
Per the conversation with @suztomo  and @parthea, owlbot bootstrapper will supply the apis.json file that contains DRIFT data for a given API. This file will later be replaced with the service_config.yaml once Cloud Drop is implemented @SurferJeffAtGoogle 

Once this is merged, I'll update owlbot-bootstrapper instructions accordingly


chingor13 (changes requested):
I have concerns with just dropping in the full `apis.json`:

1. We can probably be smarter and provide only the relevant API data rather than forcing 7 implementations to extract the one entry from `apis.json`
2. When we do switch to `service_config.yaml`, all implementations will need to switch.

IMO, we should define a consistent interface of exactly what values we will provide the per-language bootstrapping containers and provide that by whatever implementation is available at the time (via apis.json search now, via service_config.yaml later).

sofisl:
@chingor13, I added a new class to grab API-related information, which we should be able to switch out once we get the new service_config.yaml

suztomo:
What does "a consistent interface of exactly what values we will provide the per-language bootstrapping containers" look like now? Can I see in this pull request (code or comment)?

sofisl:
> @sofisl What does "a consistent interface of exactly what values we will provide the per-language bootstrapping containers" look like now? Can I see in this pull request (code or comment)?

This is the interface, line 70 in [packages/owlbot-bootstrapper/common-container/interfaces.ts] under ApiFields(https://github.com/googleapis/repo-automation-bots/pull/4505/files#diff-a84c8be49ae7f16aba06ed204fadfa3915db75aced5b6eb64efe3a2d00c577e2R70)

packages/owlbot-bootstrapper/common-container/interfaces.ts
```diff
@@ -49,4 +49,59 @@ export interface CliArgs {
...
+  apiIssueTracker?: string;
+  // API Release level or launch stage (e.g., Beta)
+  apiReleaseLevel?: string;
+  // API ID (as described by 'name' in googleapis/google/*/*/*_v*.yaml)
+  // This is confusing notation as we also have apiId which refers to the unique identifier;
+  // But this is how most .repo-metadata.json files have their configuration.
+  apiId?: string;
+}
+
+export interface ApiFields {
```
suztomo, line 70: Can you add the path to proto file? For Java, we consider "google/cloud/dataform" (no version at the end) as path to proto.

You may want to add the version (example value "v1beta") for client libraries that create different artifact per version.

packages/owlbot-bootstrapper/common-container/interfaces.ts
```diff
@@ -49,4 +49,59 @@ export interface CliArgs {
...
+  apiShortName: string;
+  // API pretty or full name (e.g., Access Approval API)
+  apiPrettyName?: string;
+  // Product documentation (e.g., https://cloud.google.com/cloud-provider-access-management/access-approval/docs)
+  apiProductDocumentation?: string;
+  // Client documentation (e.g., https://cloud.google.com/nodejs/docs/reference/access-approval/latest)
+  apiClientDocumentation?: string;
+  // API Issue tracker (e.g., https://issuetracker.google.com/issues/new?component=190865&template=1161103)
+  apiIssueTracker?: string;
+  // API Release level or launch stage (e.g., Beta)
```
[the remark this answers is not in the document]
chingor13: +1, this could be an enum type (or simply and inline `'stable' | 'preview'` type)

packages/owlbot-bootstrapper/common-container/api-field-fetcher.ts
```diff
@@ -0,0 +1,197 @@
...
+    this.storageClient = storageClient;
+  }
+
+  /**
+   * Function that gets api information from a public DRIFT bucket, and saves it to a well-known location on disk
+   * @param apiId the unique API ID
+   * @param storageClient an instance of Google Cloud Storage
+   * @returns DriftApi info
+   */
+  public async _getDriftMetadata(
```
chingor13, line 45: nit: if this is public, embrace it

```suggestion
  public async getDriftMetadata(
```

packages/owlbot-bootstrapper/common-container/api-field-fetcher.ts
```diff
@@ -0,0 +1,197 @@
...
+    return this._extractApiInfoFromJson(apis, apiId);
+  }
+  /**
+   * Function that extracts information for a given API from apis.json
+   *
+   * @param apis the apis.json file downloaded from drift as a JSON object
+   * @param apiId the unique identifier of the API
+   * @returns DriftAPI the fields as described in Drift
+   */
+  public _extractApiInfoFromJson(apis: DriftApi[], apiId: string): DriftApi {
```
chingor13, line 66: Same on this (public method with `_` name).
```suggestion
  public extractApiInfoFromJson(apis: DriftApi[], apiId: string): DriftApi {
```

packages/owlbot-bootstrapper/common-container/api-field-fetcher.ts
```diff
@@ -0,0 +1,197 @@
...
+import {getWellKnownFileContents, INTER_CONTAINER_VARS_FILE} from './utils';
+import * as fs from 'fs';
+/**
+ * Fetches API-related information
+ *
+ * @param apiId the unique API ID
+ * @param storageClient a new instance of Storage
+ * @param octokit a new instance of Octokit
+ */
+export class ApiFieldFetcher {
```
chingor13, line 28: If you're worried about having these extra public methods, you can use `interface` definitions to "hide" them.

For example, you could make `ApiFieldFetcher` be a general interface:

```ts
export interface ApiFieldFetcher {
  async getAndSaveApiInformation(directoryPath: string) => Promise<void>;
}

export class DriftApiFieldFetcher implements ApiFieldFetcher {
  // implementation you have
}
```

Then callers will only interact with a `ApiFieldFetcher` implementation (they don't care what the implementation is) and can only "see" the desired exposed API.

Also, by writing out this interface, we can see that it seems weird that the "fetcher"'s only method involves writing a file. I would expect a fetcher to return a value to be processed later. In that way, I might write the interface to be:

```ts
export interface ApiFieldFetcher {
  loadApiFields(): Promise<ApiFields>;
}
```

The code to write the file to disk would be handled elsewhere (separate the responsibilities into smaller chunks).

chingor13: Some folks might view this approach as overkill and more Java-ish which it is, but declaring an interface is pretty inexpensive in my opinion.

chingor13 (changes requested):
In the PR description, can you describe at a high level what the changes to the interface are?

My interpretation from reading the code is that we are adding all important fields into the intercontainer vars file and that file structure is well defined. Adding context to the description of the PR will help us as reviewers know what to expect and folks reading the changes later can understand what the intent is.

packages/owlbot-bootstrapper/common-container/interfaces.ts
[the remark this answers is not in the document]
sofisl: Hi @suztomo, I'm not sure what these values will be, since I don't think these have an enum from where I will be getting them from (first DRIFT, then service_config.proto - see here cl/476957576)

packages/owlbot-bootstrapper/common-container/interfaces.ts
sofisl: HI @suztomo, the path to the proto file is the same as the API_ID (as described in the environment variable, i.e., the unique identifier). For example, for google cloud it is google.cloud.kms.v1. Given that different languages may want to do different things to that path, I'll leave it as is.

packages/owlbot-bootstrapper/common-container/interfaces.ts
```diff
@@ -49,4 +49,59 @@ export interface CliArgs {
...
+  apiReleaseLevel?: string;
+  // API ID (as described by 'name' in googleapis/google/*/*/*_v*.yaml)
```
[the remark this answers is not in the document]
sofisl: I don't know of any particular reason for it being called apiId, I'm merely taking it from the survey done in this internal doc called `PRD: .repo-metadata.json data integrity` (shared with you).

packages/owlbot-bootstrapper/common-container/api-field-fetcher.ts
sofisl: Spoke about this earlier; if we decide to add in a second class, we can implement this.

packages/owlbot-bootstrapper/common-container/interfaces.ts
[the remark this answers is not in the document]
suztomo: My comment above is about how to write the document.

Not suggesting renaming apiId field.

packages/owlbot-bootstrapper/common-container/interfaces.ts
suztomo: By any chance, are the environment variable API_ID and ApiField's apiId different?

If they are different, I feel it's confusing. How might we avoid introducing the confusion?

suztomo:
@sofisl In my opinion, the ApiFields design is worth a Google Docs to solicit feedback. You can efficiently work on the implementation after the design is set.

chingor13:
We have a new design using the service_config.yaml. Let's close this PR and you can open a fresh one with the new design.

PR #4561, opened 2022-10-11
feat: create, require and read env variables for MONO_REPO_PATH, SERVICE_CONFIG_PATH, INTER_CONTAINER_VARS_PATH

sofisl:
This feature corresponds to the work in the internal doc,`Design Doc: Providing API-specific information in owlbot-bootstrapper`. We've agreed to store the paths in env variables as opposed to constants. 

There's a small refactor in this PR as well that corresponds to the env variables. I separated out a `writeToWellKnownFile` function so that it corresponds to the environment variable that you want to write to, instead of being hidden in an `openABranch` function. This makes the env variable more obvious as to where it's being used.

### DIFF
--- a/packages/owlbot-bootstrapper/cli/cloudbuild-owlbot-bootstrapper.yaml
+++ b/packages/owlbot-bootstrapper/cli/cloudbuild-owlbot-bootstrapper.yaml
@@ -35,6 +35,12 @@ steps:
     - "$_INSTALLATION_ID'"
     - "--buildId"
     - "$BUILD_ID"
+    - "--monoRepoPath"
+    - "$_MONO_REPO_PATH"
+    - "--serviceConfigPath"
+    - "$_SERVICE_CONFIG_PATH"
+    - "--interContainerVarsPath"
+    - "$_INTER_CONTAINER_VARS_PATH"
   # Generates the basic files necessary for a minimal implementation of a library
   # Current working directory contains the branch of the target repo in branchName.md
   # Keeping this as env variables so as to not force parsing on language-specific containers;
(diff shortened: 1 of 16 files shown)


chingor13 (changes requested):
This PR mostly looks fine, but it doesn't match the PR title/description. I don't see handling for env variables (I see CLI arguments being read/required).

sofisl:
@chingor13, added in the cli work (forgot to add it to the commit), PTAL

packages/owlbot-bootstrapper/cli/cloudbuild-owlbot-bootstrapper.yaml
```diff
@@ -47,6 +53,12 @@ steps:
     - 'APP_INSTALLATION_ID=$_INSTALLATION_ID'
     - 'BUILD_ID=$BUILD_ID'
     - 'GITHUB_TOKEN=$_GITHUB_TOKEN'
+    - "--monoRepoPath"
+    - "$_MONO_REPO_PATH"
+    - "--serviceConfigPath"
+    - "$_SERVICE_CONFIG_PATH"
+    - "--interContainerVarsPath"
+    - "$_INTER_CONTAINER_VARS_PATH"
```
chingor13, line 61: These should be env vars, not CLI arguments.

Shouldn't the env var that the language container receive for SERVICE_CONFIG_PATH and INTER_CONTAINER_VARS_PATH be paths including the filename? I would not expect each container to have to know/specify the file name within the directory.

chingor13 (changes requested):

PR #4567, opened 2022-10-12
feat: fetch api information from DRIFT and service_config.yaml

sofisl:
Implements the design agreed upon in `Design Doc: Providing API-specific information in owlbot-bootstrapper`

Grabs the service_config.yaml file, and, temporarily, fills in some pieces of information from DRIFT.


packages/owlbot-bootstrapper/package.json
```diff
@@ -23,6 +23,7 @@
     "lint": "cd cli && npm run lint && cd ../app && npm run lint && cd ../common-container && npm run lint"
   },
   "dependencies": {
+    "@google-automations/git-file-utils": "^1.2.1",
```
chingor13, line 26: Is this root package.json even necessary? app, cli, and common container all have their own separate package.json files.

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
```diff
@@ -0,0 +1,149 @@
...
+ * Function that extracts information for a given API from apis.json
+ *
+ * @param apis the apis.json file downloaded from drift as a JSON object
+ * @param apiId the unique identifier of the API
+ * @returns DriftAPI the fields as described in Drift
+ */
+function extractApiInfoFromJson(apis: DriftApi[], apiId: string): DriftApi {
+  const shortName = apiId.split('.')[apiId.split('.').length - 2];
+  for (const api of apis) {
+    if (api.api_shortname?.includes(shortName)) {
```
chingor13, line 53: This seems like it'll match false positives if an api short name contains another one. e.g maybe a short name is `bigquery` and this will match `bigquery-storage`.

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
```diff
@@ -0,0 +1,149 @@
...
+      return api;
+    }
+  }
+
+  return {
+    api_shortname: shortName,
+    display_name: '',
+    docs_root_url: '',
+    launch_stage: 'LAUNCH_STAGE_UNSPECIFIED' as ReleaseLevel,
```
chingor13, line 62: Does `ReleaseLevel.LAUNCH_STAGE_UNSPECIFIED` work?

chingor13, line 62: I'm unclear why we need to remap and overwrite fields if our input is `DriftApi[]` and we're finding the matching one based on shortname.

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
```diff
@@ -0,0 +1,149 @@
...
+ * Function that gets information from the service_config.yaml file in googleapis/googleapis
+ *
+ * @param apiId unique identifier fo API ID
+ * @param repositoryFileCache wrapper for getting repository contents from github (allows globbing)
+ * @returns a yaml object
+ */
+async function getApiProtoInformation(
+  apiId: string,
+  repositoryFileCache: RepositoryFileCache
+): Promise<ServiceConfigYaml | any> {
```
chingor13, line 77: We should avoid returning `any` types - we should be confident on our return types so other parts of the code do not need to be defensive

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
```diff
@@ -0,0 +1,149 @@
...
+    serviceConfig.github_label = driftData.github_label;
+  }
+
+  if (!serviceConfig?.documentation_uri) {
+    serviceConfig.documentation_uri = driftData.docs_root_url;
+  }
+
+  if (!serviceConfig?.launch_stage) {
+    serviceConfig.launch_stage = driftData.launch_stage;
+  }
```
chingor13, line 125: Consider adding logging so that we can inspect which fields are being taken from drift

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
```diff
@@ -0,0 +1,149 @@
...
+ * @param apiId the API ID of the API
+ * @param storageClient for Google CLoud Storage
+ * @param repositoryFileCache wrapper for getting repository contents from github
+ * @returns ServiceConfigYaml obejct
+ */
+export async function loadApiFields(
+  apiId: string,
+  storageClient: Storage,
+  repositoryFileCache: RepositoryFileCache
+) {
```
chingor13, line 141: Please add an explicit return type for readability

chingor13 (changes requested):

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
sofisl: Use partials.

packages/owlbot-bootstrapper/package.json
sofisl: Not necessary.

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
sofisl: Add a comment about if there's no match.

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
sofisl: Debug logging.

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
sofisl: Confirm why we wouldn't use exact matching.

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
```diff
@@ -0,0 +1,149 @@
...
+ * @param apiId the unique API ID
+ * @param storageClient an instance of Google Cloud Storage
+ * @returns DriftApi info
+ */
+async function getDriftMetadata(
+  apiId: string,
+  storageClient: Storage
+): Promise<DriftApi> {
+  const bucket = 'devrel-prod-settings';
+  const file = 'apis.json';
```
chingor13, line 34: consider turning these into global constants like `DEFAULT_BUCKET` and parameterizing this function to accept a bucket/file input (with defaults set to those constants).

sofisl:
Updated cl/481195940 as well.

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
```diff
@@ -0,0 +1,176 @@
...
+  repositoryFileCache: RepositoryFileCache
+): Promise<Partial<ServiceConfigYaml>> {
+  const path = apiId.toString().replace(/\./g, '/');
+
+  let yamlFile;
+  try {
+    yamlFile = await repositoryFileCache.getFileContents(
+      `${path}/.*?_v.*?.yaml$`,
+      'main'
+    );
```
chingor13, line 95: This method is for fetching file contents by an exact file by name. There is a `findFilesByGlob` that can return 0-n matching file names which can be used to discover the file to fetch.

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * @param apiId unique identifier fo API ID
+ * @param repositoryFileCache wrapper for getting repository contents from github (allows globbing)
+ * @returns a yaml object
+ */
+async function getApiProtoInformation(
+  apiId: string,
+  repositoryFileCache: RepositoryFileCache
+): Promise<Partial<ServiceConfigYaml>> {
+  const path = apiId.toString().replace(/\./g, '/');
```
chingor13, line 88: I'm not sure this makes sense to be here. The input should probably be the proto directory, not the api ID. If the directory can be derived from the API ID, that's fine, but it doesn't belong in this function.

chingor13 (changes requested):

chingor13 (approved):

packages/owlbot-bootstrapper/common-container/fetch-api-info.ts
```diff
@@ -0,0 +1,211 @@
...
+  driftData: DriftApi
+): ServiceConfigYaml {
+  if (!serviceConfig?.api_short_name) {
+    logger.info(
+      `Service config did not contain api_short_name; replacing with DRIFT ${driftData.api_shortname}`
+    );
+    serviceConfig.api_short_name = driftData.api_shortname;
+  }
+
+  if (!serviceConfig?.github_label) {
```
tmatsuo, line 147: In my understanding, `github_label`, `api_shortname`, `documentation_uri`, `doc_tag_prefix` are under `publishing`, not at the top level.

sofisl: Updated accordingly; thank you!

PR #4577, opened 2022-10-15
Get full mono repo path

sofisl:
This PR adds in an argument to the owlbot-bootstrapper, the full mono-repo path.

Since the CLI should only receive the directory in which to clone the mono repo, and the ssh address of the mono repo to clone, this PR does the following things:

1) Parses the repoToClone ssh address to get the mono repo name; then concatenates it to the monoRepoDir, which is the actual cli input. Then, throughout the rest of the process, we pass through the monoRepoDir, and the monoRepoPath. Language containers will only see the monoRepoPath.

2) Since we're already parsing the repoToClone ssh address, I'm removing where we do that in the common container, and instead passing through the parsed variables (monoRepoName and monoRepoOrg). This standardizes the regex matching we're doing as well.


packages/owlbot-bootstrapper/cli/run-trigger-command.ts
```diff
@@ -47,6 +47,18 @@ export function getLanguageSpecificValues(language: string) {
   throw new Error('No language-specific container specified');
 }
 
+export function getMonoRepoNameAndOrg(repoToClone: string | undefined) {
```
chingor13, line 50: Please define the return type explicitly

packages/owlbot-bootstrapper/cli/run-trigger.ts
```diff
@@ -21,13 +21,14 @@ import {CliArgs} from './run-trigger-command';
...
 const DIRECTORY_PATH = '/workspace';
-const MONO_REPO_PATH = DIRECTORY_PATH;
+const MONO_REPO_DIR = DIRECTORY_PATH;
 const SERVICE_CONFIG_PATH = `${DIRECTORY_PATH}/serviceConfig.yaml`;
 const INTER_CONTAINER_VARS_PATH = `${DIRECTORY_PATH}/interContainerVars.json`;
 
 export async function runTrigger(
   argv: CliArgs,
   cb: CloudBuildClient,
+  monoRepoNameAndOrg: {monoRepoName: string; monoRepoOrg: string},
```
chingor13, line 31: This type could be re-used. Additionally, the `monoRepo` part could be redundant.

Consider a type like:

```ts
interface GitHubRepository {
  owner: string;
  name: string;
}
```

packages/owlbot-bootstrapper/cli/test/run-trigger.test.ts
```diff
@@ -154,4 +162,31 @@ describe('tests running build trigger', () => {
...
+      runTriggerCommand.getMonoRepoNameAndOrg(
+        'git@github.com:googleapis/nodejs-scheduler.git'
+      ),
+      {monoRepoOrg: 'googleapis', monoRepoName: 'nodejs-scheduler'}
+    );
+
+    assert.throws(() => {
+      runTriggerCommand.getMonoRepoNameAndOrg(undefined);
+    }, new Error("Repo to clone arg is malformed; should be in form of ssh address,' git@github.com:googleapis/google-cloud-node.git'"));
+  });
```
chingor13, line 191: nit: put these in a separate `describe('getMonoRepoNameAndOrg')` block

packages/owlbot-bootstrapper/cli/run-trigger-command.ts
chingor13: Also, this is probably better named as `parseRepoNameAndOrg` or `parseGitHubRepository` as the fact that it's a mono-repo is irrelevant.

chingor13 (changes requested):

packages/owlbot-bootstrapper/cli/test/run-trigger.test.ts
```diff
@@ -154,4 +162,33 @@ describe('tests running build trigger', () => {
       }
     );
   });
+
+  describe('get mono repo name and org from ssh address', () => {
```
chingor13, line 166: nit: I would just name this describe block the exported function name being tested

This lets you easily run the subset of tests:

```
npm t -- -g 'parseRepoNameAndOrg'
```

chingor13 (approved):

Issue #4617, opened 2022-10-24
Add support for bypassing branch protection to sync-repo-settings bot


sararob:
The googlapis/python-aiplatform repo is using [Copybara](https://github.com/google/copybara) as part of our contribution workflow. In order to merge pull requests, the Copybara app needs to be able to bypass branch protections. I think this is a new setting on GitHub.

Could the `sync-repo-settings` bot add a [config option here](https://github.com/googleapis/repo-automation-bots/blob/main/packages/sync-repo-settings/src/schema.json) to enable unchecking the box for "Do not allow bypassing the above settings" in GitHub's branch protection rules? This will allow only the Copybara app and repo admins to bypass these settings.

chingor13:
I see an API call for this: https://docs.github.com/en/rest/branches/branch-protection#set-app-access-restrictions, but when I query the endpoint, I don't see existing entries (it's possibly hidden by a feature flag header for beta features that I don't see documented anywhere).

sararob:
Thanks for looking into this! Can you clarify what you mean by existing entries? For our repo, we'd want to set this as:

```
"bypass_pull_request_allowances":{"apps":["copybara-service"]}
```

chingor13:
Yes, we can possibly set it, but the getter API does not seem to return what the documentation states. For example:

```
curl -H "Accept: application/vnd.github+json" -H "Authorization: Bearer ${GITHUB_TOKEN}" https://api.github.com/repos/googleapis/python-aiplatform/branches/main/protection/restrictions/apps
{
  "message": "Push restrictions not enabled",
  "documentation_url": "https://docs.github.com/rest/reference/repos#list-apps-with-access-to-the-protected-branch"
}
```

I would expect the API to return the current allowlisted app in that response. I have little confidence that this API is actually working.

sararob:
Hmm, what does a GET request to `https://api.github.com/repos/googleapis/python-aiplatform/branches/main/protection` return for our repo?

If setting `bypass_pull_request_allowances` isn't possible right now, is there a way to disable the `sync-repo-settings` bot (and not use the default settings) until we can set this?

chingor13:
> what does a GET request to `https://api.github.com/repos/googleapis/python-aiplatform/branches/main/protection` return for our repo?

When I ran the GET request it returned that `Push restrictions not enabled` response which was unexpected because I see it in the UI.

Changing this to a feature request as #4626 implemented a workaround to skip stomping branch protection

[...]

PR #4626, opened 2022-10-25
feat: add enabled flag to sync-repo-settings

sararob:
Adds an enabled flag to turn off the sync-repo-settings bot.

Workaround for #4617

### DIFF
--- a/packages/sync-repo-settings/src/sync-repo-settings.ts
+++ b/packages/sync-repo-settings/src/sync-repo-settings.ts
@@ -48,6 +48,7 @@ function deepFreeze(object: any) {
 const languageConfig: LanguageConfig = deepFreeze(checks);
 
 const repoConfigDefaults: RepoConfig = deepFreeze({
+  enabled: true,
   mergeCommitAllowed: false,
   squashMergeAllowed: true,
   rebaseMergeAllowed: true,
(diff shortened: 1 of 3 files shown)


packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -82,6 +83,10 @@ export class SyncRepoSettings {
     const logger = this.logger;
     const repo = options.repo;
     const [owner, name] = repo.split('/');
+    if (!config?.enabled) {
+      logger.info(`config is not enabled for repository ${repo}`);
+      return false;
```
chingor13, line 88: Looks like this method returns a Promise<void>

chingor13: Yeah

packages/sync-repo-settings/src/sync-repo-settings.ts
```diff
@@ -82,6 +83,10 @@ export class SyncRepoSettings {
     const logger = this.logger;
     const repo = options.repo;
     const [owner, name] = repo.split('/');
+    if (!config?.enabled) {
```
chingor13, line 86: Looks like this should probably be
```suggestion
    if (config?.enabled === false) {
```

Then if the enabled field is not set, we default to running the sync.

chingor13 (changes requested):

chingor13 (approved):

PR #4646, opened 2022-11-02
fix: catch Datastore timeouts and throw ServiceUnavailable error

### LINKED ISSUE #3652: merge-on-green: occasional timeout to datastore
```
Error: Total timeout of API google.datastore.v1.Datastore exceeded 60000 milliseconds before any response was received.
```

sofisl:
Fixes #3652


packages/merge-on-green/src/merge-on-green.ts
```diff
@@ -101,7 +102,25 @@ handler.getDatastore = async function getDatastore(installationId?: number) {
...
-  const [prs] = await datastore.runQuery(query);
+  let prs;
+  try {
+    [prs] = await datastore.runQuery(query);
+  } catch (err) {
+    if (
+      (err as Error)
+        .toString()
+        .includes(
+          'Total timeout of API google.datastore.v1.Datastore exceeded 60000 milliseconds'
```
chingor13, line 113: Checking error types based on the message is error prone. Can we look at the request status/code first before looking at the message?

chingor13 (changes requested):
If we can't find any recent instances of this error (to discover what the full error looks like), let's close and fix when we get another instance

packages/merge-on-green/src/merge-on-green.ts
sofisl: I haven't seen any recent instances, so I wasn't able to look into the error.

PR #4659, opened 2022-11-04
feat: add option to ignore year in license header bot

kweinmeister:
There are scenarios in which it is helpful to ignore the license year check, such as migrating code across repos. This PR introduces a new option `ignoreLicenseYear` to `header-checker-lint`. The default setting is `false`, so there are no changes to the current behavior.

- [X] Ensure the tests and linter pass
- [X] Appropriate docs were updated (if necessary)


chingor13 (commented):
LGTM, but need to fix the lint with `npm run fix`

bcoe (approved):
LGTM.

PR #4663, opened 2022-11-04
fix: wait for all PRs to attempt to merge before throwing error

chingor13:
This PR modifies the main merge-on-green loop to run the main logic for all the PRs being considered. If any of the individual work items throws an error, catch it and throw an `AggregateError` containing all the sub-errors. Our error reporting framework should be able to unwrap nested `AggregateError`s and report each to error reporting.

Additionally, we add PR context to the logger for the main logic so it's easier to filter logs for a specific pull request.

feat: add PR context to the logger for each PR merge attempt
refactor: throw AggregateError for any errors thrown while handling PRs


sofisl (changes requested):
Ideally, we'd separate out the refactoring into parallelization from the new AggregateError work. However, it seems like the AggregateError is actually the less important part in this PR, vs. the parallelization work. If that's the case, then I would edit the PR description.

chingor13:
> separate out the refactoring into parallelization from the new AggregateError work

This is not really feasible - to refactor into parallel work, you need to be able to throw an aggregation of one or more thrown error from a work item.

sofisl (approved):

PR #4676, opened 2022-11-14
fix(owlbot-bootstrapper): get correct default branch for bootstrapper

### LINKED ISSUE #4675: Owlbot-bootstrapper is not loading service_config.yaml properly
The bootstrapper is not generating any of the files for a given library: https://github.com/googleapis/google-cloud-node/pull/3516

This is the log: https://pantheon.corp.google.com/cloud-build/builds;region=global/db42a7f0-ecdd-4ed2-a564-ebb54b3450dc;step=0?project=owlbot-bootstrap-prod

sofisl:
Fixes #4675 

We were looking for the service_config.yaml file in 'main', but it should have been 'master'.

I also added a function to get the default branch for when we're creating PRs, since we don't know what each library has as their default branch. I didn't do this for requests against googleapis/googleapis though, since it will always be a constant, and it would be wasteful to grab that default branch each time.

### DIFF
--- a/packages/owlbot-bootstrapper/common-container/__snapshots__/utils.test.js
+++ b/packages/owlbot-bootstrapper/common-container/__snapshots__/utils.test.js
@@ -1,4 +1,4 @@
-exports['common utils tests opens a PR against the main branch 1'] = {
+exports['common utils tests opens a PR against the default branch 1'] = {
   head: 'specialName',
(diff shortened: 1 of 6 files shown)


bcoe (approved):
LGTM (we should try to make all our branch names consistent, but not a blocker for this).

PR #4680, opened 2022-11-15
docs: deprecate most configuration options

chingor13:
As release-please's configuration options have gotten more and more complex, we only want to support them if you are using the manifest (json) configuration. This PR updates the documentation to recommend using a manifest config and to point to further documentation about the available configuration options.


bcoe (approved):
I think it's a really good decision to move people towards the manifest for configuration 👍

PR #4736, opened 2022-12-06
chore(snippet-bot): update lock file

tmatsuo:
For security patch


tmatsuo:
I found out `npm audit fix` fixes our issue, so I'll change

PR #4766, opened 2022-12-09
feat(ccl): add config to only check PR title

### LINKED ISSUE #4769: feat(ccl): introduce a config for always check against PR title
Github repository can be configured for only using squash commit and use the PR title for the final commit message. In this case, you may want to always check the PR title for conventional commit spec.

Related:
https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1304#issuecomment-1341636666

cc @apeabody

apeabody:

Fixes #4769 🦕
Fixes https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1304

### DIFF
--- a/packages/conventional-commit-lint/src/utils.ts
+++ b/packages/conventional-commit-lint/src/utils.ts
@@ -55,7 +55,8 @@ export async function scanPullRequest(
   context: Context<'pull_request'> | Context<'issue_comment'>,
   pull_request: PullRequest,
   logger: GCFLogger,
-  octokit: Octokit
+  octokit: Octokit,
+  onlyCheckPrTitle: boolean
 ) {
   // Fetch last 100 commits stored on a specific PR.
   const commitParams = context.repo({
@@ -108,7 +109,10 @@ export async function scanPullRequest(
   // This is done becaues GitHub uses the commit title, rather than the
   // issue title, if there is only one commit:
   let usingCommitMessage = false;
-  if (commits.length === 1 && !hasAutomergeLabel && !hasAutoMergeEnabled) {
+  if (
+    commits.length === 1 &&
+    ((!hasAutomergeLabel && !hasAutoMergeEnabled) || onlyCheckPrTitle)
+  ) {
     message = commits[0].commit.message;
     usingCommitMessage = true;
   }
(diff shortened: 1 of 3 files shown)


packages/conventional-commit-lint/src/utils.ts
```diff
@@ -182,12 +185,16 @@ export async function scanPullRequest(
   if (lintError) {
     let summary =
       'Some of your commit messages failed linting.\n\nVisit [conventionalcommits.org](https://conventionalcommits.org) to learn our conventions.\n\n';
+    if (commitMessageDiffer === true) {
+      summary +=
+        'The PR title and the commit message differ, and there is only one commit.\n';
```
tmatsuo, line 190: This message tells you about the current situation, but doesn't tell you what you should do. Can we tell users what to do next?

tmatsuo:
I think you'll need to update the test snapshot. IIRC, you can run `npm run test:snap` or something to update the snapshot. Then you may need to run `npm fix` for formatting the snapshot file.

packages/conventional-commit-lint/src/utils.ts
apeabody: Thanks @tmatsuo! - If I understand the current code correct, this will only occur when the commits.length === 1 AND the commit message initially fails lint, so the subsequent "Run `git commit --amend` and edit your message to match Conventional Commit guidelines." will also be added to the summary.  Let me know if I've misunderstood, or there is a different action we would like to communicate to users.

My goal was to limit the change in behavior, the other idea was to add a new configuration parameter alwaysUsePrTitle = T/F.  Useful for repos which only use merge commits with the PR title, and so don't worry about single commit messages.

tmatsuo: Here is the intention of the current code.

Background: This bot is to enforce all the commit messages conforms to the conventional commit spec. When someone opens a PR on the Github UI for a simple change, people tend to choose a PR title incompatible with conventional commit. In that case, they often fix the PR title to please this bot.

However, the bot still complains because Github will use the commit message (not the PR title) when you merge this PR without squashing.

So the bot adds this message in the issue commment.

```
you can use Github's automerge feature with squashing, or use `automerge` label.
```

If you use Github's automerge feature with squashing, it will pick the PR title for the squashed merge commit.
If you add `automerge` label (which will trigger one of our other bots "merge-on-green"), the final commit message will be picked from the PR title.

So this bot will stop complaining for those cases when the PR title conforms to conventional commit.

apeabody: Based on feedback, retooling for https://github.com/googleapis/repo-automation-bots/issues/4769

apeabody:
Close in favor of #4770

PR #4800, opened 2022-12-20
fix: lock the version of typescript to 4.8.4

SurferJeffAtGoogle:
See https://github.com/googleapis/repo-automation-bots/pull/4799

4.9 chokes on a dependency.

```
node_modules/@types/node/globals.d.ts:72:13 - error TS2403: Subsequent variable declarations must have the same type.  Variable 'AbortSignal' must be of type '{ new (): AbortSignal; prototype: AbortSignal; abort(reason?: any): AbortSignal; timeout(milliseconds: number): AbortSignal; }', but here has type '{ new (): AbortSignal; prototype: AbortSignal; }'.

72 declare var AbortSignal: {
               ~~~~~~~~~~~

  node_modules/typescript/lib/lib.dom.d.ts:2071:13
    2071 declare var AbortSignal: {
                     ~~~~~~~~~~~
    'AbortSignal' was also declared here.

Found 1 error in node_modules/@types/node/globals.d.ts:72
```

### DIFF
--- a/packages/gcf-utils/package.json
+++ b/packages/gcf-utils/package.json
@@ -70,7 +70,7 @@
     "snap-shot-it": "^7.9.6",
     "sonic-boom": "^3.2.0",
     "stream-mock": "^2.0.5",
-    "typescript": "~4.8.2"
+    "typescript": "4.8.2"
   },
   "engines": {
     "node": ">= 14"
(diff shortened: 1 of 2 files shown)


packages/gcf-utils/package.json
```diff
@@ -70,7 +70,7 @@
     "snap-shot-it": "^7.9.6",
     "sonic-boom": "^3.2.0",
     "stream-mock": "^2.0.5",
-    "typescript": "~4.8.2"
+    "typescript": "4.8.2"
```
tmatsuo, line 73: 4.8.4 was released 3 months ago. Maybe we can use it instead?

SurferJeffAtGoogle: Done.

tmatsuo (approved):

PR #4820, opened 2022-12-22
fix: upgrade jsonwebtoken to 9.0.0

SurferJeffAtGoogle:
To fix lots of dependabot alerts.

PR #4830, opened 2022-12-27
build(owlbot-bootstrapper): refactor tests to not directly test jsonwebtoken

sofisl:
Tests are breaking the bootstrapper here: https://github.com/googleapis/repo-automation-bots/pull/4820

However, after performing some integration tests locally, the functionality of the bot seems unaffected. I think this is a situation where the tests are too restrictive, so I've refactored to do more mocking.


packages/owlbot-bootstrapper/common-container/test/github-authenticator.test.ts
```diff
@@ -74,37 +75,61 @@ describe('behavior of Github Authenticator Class', async () => {
...
+      async () => {
+        const resp = await request({
+          url: 'https://api.github.com/app/installations/2345567/access_tokens',
+          method: 'POST',
+          headers: {
+            Authorization: 'Bearer 1234',
+            Accept: 'application/vnd.github.v3+json',
+          },
+        });
+        return (resp as any).data.token as string;
```
chingor13, line 92: I don't think this refactor makes sense to do -- it's testing its own fake. This replaces the `getGitHubShortLivedAccessToken` with a fake that makes and API call that's faked.

chingor13 (changes requested):

packages/owlbot-bootstrapper/common-container/test/github-authenticator.test.ts
sofisl: You're right. I couldn't think of a better solution but the break helped me come up with something better. Thanks!

packages/owlbot-bootstrapper/common-container/test/github-authenticator.test.ts
```diff
@@ -40,12 +41,15 @@ secretManagerClientStub.accessSecretVersion.resolves([
...
+const signStub = sinon.stub(jwt, 'sign');
+
 describe('behavior of Github Authenticator Class', async () => {
   it('should create the right type of object', async () => {
     const githubAuthenticator = new GithubAuthenticator(
       'projectId',
       '2345567',
-      secretManagerClientStub
+      secretManagerClientStub,
+      signStub as unknown as typeof jwt.sign
```
chingor13, line 52: Generally we don't pass method stubs around as parameters (you would pass a fake or a mock).

If using stubs, you would normally do:

```
const signStub = sinon.stub(jwt, 'sign').returns(something); // or .resolves(something);
```

This would stub out the response of jwt's sign method when it's called. As this is written, we're passing this sinon stub method into this method to be called which is somewhat odd.

chingor13 (changes requested):
I highly recommend splitting this class into it's own tested library

packages/owlbot-bootstrapper/common-container/test/github-authenticator.test.ts
sofisl: Hm, I don't see why this would be a problem, since it's not changing the tests' functionality. In the original test, we're passing a fake private key, and creating a jwt and then making a github call with that jwt. Ultimately though, this test was just making sure the github call was made; it was not testing that the jwt was created correctly. This test just makes it so that we're not testing the jwt dependency directly, but we don't need to be testing the output of what the sign function does.

packages/owlbot-bootstrapper/common-container/github-authenticator.ts
```diff
@@ -30,16 +30,19 @@ export class GithubAuthenticator {
...
   secretManagerClient: SecretManagerServiceClient;
+  signingClient: typeof sign;
   OWLBOT_SECRET_NAME = 'owlbot-bootstrapper';
 
   constructor(
     projectId: string,
     appInstallationId: string,
-    secretManagerClient: SecretManagerServiceClient
+    secretManagerClient: SecretManagerServiceClient,
+    signingClient: typeof sign
```
chingor13, line 40: I would declare your own type for this rather than having it dynamically typed based on a third-party library. You could define this type as `(payload: JwtPayload, privateKey: string, options: SomeOptionInterface) => string` which jwt's `sign` would match. If the third-party interface breaks, then your interface breaks. You are also exposing more options than necessary to the public interface.

Another option is to declare this as a simpler type like `(payload: object, privateKey: string) => string` and have your default implementation:

```ts
function jwtSigner(payload: object, privateKey: string): string {
  return sign(payload, privateKey, {algorithm: 'RS256'});
}
```

chingor13 (changes requested):
Ok, I see what this is trying to do now, but I think this is mixing testing paradigms.

If you are choosing to dependency inject, then don't pass a stubbed instance -- instead, provide a concrete test implementation.

e.g.

```ts
function testSigner(payload: object, privateKey: string): string {
  return 'foo';
}
```

If you are choosing to mock, then mock the entire jwt `sign` method and don't worry about dependency injecting into your authenticator class.

packages/owlbot-bootstrapper/common-container/pre-process.ts
```diff
@@ -27,6 +27,7 @@ import {CliArgs, Language} from './interfaces';
 import {loadApiFields} from './fetch-api-info';
 import {Storage} from '@google-cloud/storage';
 import {RepositoryFileCache} from '@google-automations/git-file-utils';
+import {sign} from 'jsonwebtoken';
```
chingor13, line 30: This import doesn't look used anymore.

chingor13 (approved):

bcoe:
@chingor13 @sofisl, just a thought on this one to reduce the amount of code change needed. I think the reason that tests break is because a 2048 bit key is now enforced, and our fake key used for testing is less bits than this.

You could generate a fake key that's 2048 bits and get tests working again.

Issue #4474, opened 2022-09-22
Error: Git Repository is empty.


chingor13:
This error is being thrown by the repo-metadata-lint cron job.

Stack trace:
```
HttpError: Git Repository is empty.
    at /workspace/node_modules/@octokit/request/dist-node/index.js:88:21
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async sendRequestWithRetries (/workspace/node_modules/@octokit/auth-app/dist-node/index.js:376:12)
    at async FileIterator.getFileListing (/workspace/build/src/file-iterator.js:61:26)
    at async FileIterator.repoMetadata (/workspace/build/src/file-iterator.js:40:23)
    at async /workspace/build/src/repo-metadata-lint.js:68:41
    at async Promise.all (index 0)
    at async /workspace/node_modules/gcf-utils/build/src/gcf-utils.js:389:25
```

Example repository: https://github.com/googleapis/google-cloud-php-api-keys (which is currently empty)

Opening as a P3 because this is an uncommon occurrence.

SurferJeffAtGoogle:
@bcoe appears to be the author of this bot.

PR #4863, opened 2023-01-05
fix: handle empty repository as if it has no files

chingor13:
When a repository is uninitialized, GitHub throws a 409 error on file operations. In this library, we want to silently handle this scenario as if the repository has no files.

Towards #4474


bcoe (approved):
:shipit:

PR #4927, opened 2023-01-10
fix: auto-label should recursively search through configuration to find the right label

### LINKED ISSUE #4926: auto-label is not labeling PRs appropriately
See on nodejs-docs-samples:

https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/2966

With corresponding auto-label config: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/.github/auto-label.yaml

No labels are added.

sofisl:
Fixes #4926 

The previous logic did not recursively search through nested objects, for example, the README example did not work:

```
# In .github/auto-label.yaml
path:
  pullrequest: true
  paths:
    .: 'root'
    # i.e., label all PRs of files in root directory and down as "root"
    test:
      .: 'test'
      # i.e. label all files in ./test/ directory and down as "test"
      fixtures: 'testfixtures'
      # i.e. label all files in ./test/fixtures and down as "testfixtures"
```
Auto-label did not label things as `testfixtures`, because it didn't go into the object recursively.


packages/auto-label/src/helper.ts
```diff
@@ -278,32 +278,25 @@ export function autoDetectLabel(
...
-      }
-    } else {
-      break; // break as this is the end of user defined path
+    if (typeof config[dir] === 'string') {
+      label = config[dir] as string;
+      exactMatch = true;
+    } else if (typeof config[dir] === 'object' || Array.isArray(config[dir])) {
+      label = getLabelFromPathConfig(filename, config[dir] as PathConfig);
+    } else if (config['.'] && !exactMatch) {
+      label = config['.'] as string;
```
chingor13, line 295: This is kinda hard to follow, can you add comments for these 3 scenarios?

chingor13 (changes requested):

packages/auto-label/src/helper.ts
```diff
@@ -278,32 +278,33 @@ export function autoDetectLabel(
...
+  // There are 3 possiblities for configuration file:
+  // 1. There is an exact match of filename and configuration path.
+  // In this case, set the label to that configuration. However, keep
+  // searching through the rest of the configuration file to find if
+  // there are any deeper matches.
+  // 2. There is a nested array or object that may contain a deeper match.
+  // If so, search recursively through that object.
+  // 3. If an exact deeper match was never found, set the label as whatever is
+  // listed for the whole directory, '.'
   for (const dir of dirs) {
```
chingor13, line 296: This is the for-loop that doesn't make sense to me.

chingor13 (approved):
After digging into the original code, I have doubts whether to for loop is the right behavior. It will match nested paths even though the config allows for nested paths.

Example:
file: `run/vision/compute/foo.ts`
config: has paths for `run`, `vision`, and `compute`

This would pick compute as the label. I would assume that the config is from the file root so this would match `run`. If you wanted to target further, you would nest the config:

```json
{
  "run": {
    "vision": {
      "compute": "run-vision-compute",
      ".": "run-vision"
    },
    ".": "run"
  },
  "vision": "vision",
  "compute": "compute"
}
```

Approving this to fix the bug, but this should probably be cleaned up.

packages/auto-label/src/helper.ts
sofisl: I see what you're saying. However, this is copied from the original function/author, so this preserves past functionality. Changing this might break expectations for others.

Issue #4946, opened 2023-01-24
Post processor reports failure when it was actually a success 


codyoss:
I have an OwlBot PR that is reporting an "unkown build failure" but if I drill into the build everything passed. Have seen this a couple times now:

- PR: https://github.com/googleapis/google-cloud-go/pull/7290

From a side chat @chingor13 mentioned it might be related to handling of a lock.

chingor13:
From the logs: `triggerPostProcessBuild: timed out waiting for build 51f32ea1-1593-435f-bd96-34bba4285316`

https://github.com/googleapis/repo-automation-bots/blob/424f3e348ff8f40953919a1f1750134b0800f007/packages/owl-bot/src/core.ts#L155-L157

chingor13:
Looks like we wait ~20 minutes (20000ms * 60 attempts) for the post-processor to run. The GCB job has a higher timeout so it succeeds, but the request that is watching the state of the GCB job gives up.

codyoss:
Got it, sounds good. We have an PR in the works to limit the amount of work done by the post processor for most change. This should in term reduce the time well under the 20 minute time it is today. Although I suspect the tail end build time to still be long when all client get touched right now. We have a seperate project underway that is moving snippet generation into the generator and out of the post processor. Once this is complete this should not be an issue at all.

PR #4958, opened 2023-01-25
fix: improve timeout error message

chingor13:
If the owlbot event handler times out waiting for the status of the post-processor GCB job, then return a better error status.

Towards #4946

### DIFF
--- a/packages/owl-bot/src/core.ts
+++ b/packages/owl-bot/src/core.ts
@@ -155,6 +155,14 @@ export async function triggerPostProcessBuild(
     logger.error(`triggerPostProcessBuild: ${err.message}`, {
       stack: err.stack,
     });
+    if (err instanceof TimeoutError) {
+      return {
+        conclusion: 'failure',
+        summary: 'timed out waiting for build',
+        text: 'timed out waiting for build',
+        detailsURL,
+      };
+    }
     return buildFailureFrom(detailsURL);
   }
 }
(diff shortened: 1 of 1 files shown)


packages/owl-bot/src/core.ts
```diff
@@ -200,6 +208,13 @@ function detailsUrlFrom(buildID: string, project: string): string {
   return `https://console.cloud.google.com/cloud-build/builds;region=global/${buildID}?project=${project}`;
 }
 
+class TimeoutError extends Error {
```
SurferJeffAtGoogle, line 211:
```
class TimeoutError extends Error { name = "TimeoutError" }
```

Is briefer and accomplishes the same thing.

SurferJeffAtGoogle (changes requested):

packages/owl-bot/src/core.ts
```diff
@@ -160,6 +160,14 @@ export async function triggerPostProcessBuild(
     logger.error(`triggerPostProcessBuild: ${err.message}`, {
       stack: err.stack,
     });
+    if (err instanceof TimeoutError) {
```
SurferJeffAtGoogle, line 163: I'd rather see `buildFailureFrom()` get smarter:

```ts
function buildFailureFrom(error: any, detailsUrl: string): BuildResponse {
  if (typeof error.name === 'string' && typeof error.message === 'string) {
    return {
      conclusion: 'failure',
      summary: error.name,
      text: error.message,
      detailsUrl
    };
 } else {
...
}
```

That way, we can see something useful for any kind of exception.

P.S.  I intended to propose this change in my first review ,but I must have clicked the wrong button or something and it got dropped.

chingor13: Done

SurferJeffAtGoogle (approved):

PR #4961, opened 2023-01-26
build: increase timeout for post processor check

### LINKED ISSUE #4946: Post processor reports failure when it was actually a success

sofisl:
Allows the post-processor check to wait as long as the build before it returns to GH

Example: https://github.com/googleapis/google-cloud-node/pull/3821/checks?check_run_id=10890026001

Fixes #4946

### DIFF
--- a/packages/owl-bot/src/core.ts
+++ b/packages/owl-bot/src/core.ts
@@ -205,7 +205,9 @@ async function waitForBuild(
   id: string,
   client: CloudBuildClient
 ): Promise<google.devtools.cloudbuild.v1.IBuild> {
-  for (let i = 0; i < 60; i++) {
+  // This loop is set to equal a total of 3 hours, which should
+  // match the timeout in cloud-build/update-pr.yaml's timeout
+  for (let i = 0; i < 540; i++) {
     const [build] = await client.getBuild({projectId, id});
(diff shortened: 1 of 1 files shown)


packages/owl-bot/src/core.ts
```diff
@@ -205,7 +205,9 @@ async function waitForBuild(
   id: string,
   client: CloudBuildClient
 ): Promise<google.devtools.cloudbuild.v1.IBuild> {
-  for (let i = 0; i < 60; i++) {
+  // This loop is set to equal a total of 3 hours, which should
+  // match the timeout in cloud-build/update-pr.yaml's timeout
+  for (let i = 0; i < 540; i++) {
```
bcoe, line 210: nit: I would pull the 540 and 20000 out into variables, so that the duration of the wait speaks for itself:

const THREE_HOURS = 3 * 60 * 60 * 1000;
const delay = 20000;
const iterations = parseInt(threeHours / 20000);

bcoe (changes requested):

chingor13 (approved):

Issue #4698, opened 2022-11-30
secret-rotator timeout without retry


chingor13:
It looks like one of our secret rotator jobs failed with a timeout (trace=projects/secret-rotator-prod/traces/e405d31910521e2757dab3c701d39aac). There's no retry mechanism so that secret ended up expiring (key expires after 3 months, rotation set to every 2 months so a single failure would let the key expire).

- [x] we should be notified on these errors
- [x] we should have some kind of retry mechanism for these transient failures
- [x] we should extend the timeout - it was probably a service cold start
- [x] we could spread the cron jobs out so they do not all run immediately at midnight on the same day after 0 traffic for a month

sofisl:
Fixed with cl/506436972

PR #4970, opened 2023-02-02
build: add logging to secret rotator

### LINKED ISSUE #4966: Secret Rotator fails without notification
The secret rotator's job scheduler failed in executing a rotation for google-cloud-java's service account, but we didn't know until the credentials stopped working. Manually running the job resolved our credentials issue.

There are several other jobs that have status 'Failed', and as @chingor13 mentioned in chat:
> It defeats the whole point if the rotation fails and we aren't proactively notified. Especially when it only runs like every other month

Please contact me in chat for any links or further details, if needed.

sofisl:
Along with cl/506436972, should fix #4966 and #4698

### DIFF
--- a/packages/secret-rotator/src/secret-rotator.ts
+++ b/packages/secret-rotator/src/secret-rotator.ts
@@ -34,13 +34,20 @@ export class SecretRotator {
     serviceAccountEmail: string
   ) {
     const name = `projects/${serviceAccountProjectId}/serviceAccounts/${serviceAccountEmail}`;
-    const key = await client.projects.serviceAccounts.keys.create({
-      name,
-      requestBody: {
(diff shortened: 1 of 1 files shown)


chingor13 (changes requested):
I don't think this is going to add much. Gcf-utils, assuming the version is up to date, will catch any exception and report it to error reporting. I'dake sure gcf-utils is at the latest version to get error reports from errors thrown from our bots

sofisl:
It should be: https://github.com/googleapis/repo-automation-bots/blob/ad632b9bfd3a7d5cbc30a27947407d5f8beb36ae/packages/secret-rotator/package.json#L50

I can't see the failed logs because it's been past 30 days (I tried looking earlier this week). Is it better to just close this PR then?

sofisl:
I'm going to close this for now and the bugs; seems like it's not needed.

PR #4990, opened 2023-03-03
chore: migrate title to regex in auto-approve v1 schema

kweinmeister:
This update updates the title field of the auto-approve v1 schema to a pattern rather than a fixed string. This allows for exceptions to certain packages.

We will explore defining a new process with the v2 schema, and this will help unblock us for now.


sofisl (changes requested):
The pattern you're suggesting is actually a bit more permissible than the constant that's currently in place, it is only matching `^(fix`. The JSON-escaped pattern that would most match the current const is: 

```
"\\^\\(fix\\|chore\\)\\\\\\\\\\(deps\\\\\\\\\\):"
```
However, we're left with a problem of the quotations, since the quotations are a part of the pattern the regex is searching for. This would make that pattern above a breaking change for current valid schemas.

FWIW, I don't think the pattern you're looking to match, i.e., this one: ```"^(fix|chore)\\(deps\\):(?!.*(exception1|exception2)).*$```, would match a normal pattern, for [example](https://regex101.com/r/OJPhxa/1).

I played with attempting to create a pattern for a pattern, I think it's a really tall order that's bound to create some unexpected issues. The `pattern` property is really intended to act on a literal, not another Regex, so I think it will be hard to forsee what kinds of strings will pass.

I'd strongly recommend either creating a new rule, or something like:

```
              "oneOf": [
                      { "type": "string", "const": "^(fix|chore)\\(deps\\):" },
                      {"type": "string", "const": "^(fix|chore)\(deps\):(?!.*(exception1|exception2)).*$"}
                    ]
```
with the exceptions filled out.

kweinmeister:
Closing in favor of a v2 schema version (needs to be written).

PR #5002, opened 2023-03-16
feat: add python sample dependency

sofisl:


packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -244,8 +244,30 @@ export function doesDependencyChangeMatchPRTitle(
   return false;
 }
 
+export function doesDependencyAgainstRegexes(
```
chingor13, line 247: Please add a jsdoc to describe what this is doing.

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -244,8 +244,30 @@ export function doesDependencyChangeMatchPRTitle(
...
+  versions: Versions,
+  regexToInclude: RegExp[] | undefined,
+  regexToExclude: RegExp[] | undefined
```
chingor13, line 250: These could be optional
```suggestion
  regexToInclude?: RegExp[],
  regexToExclude?: RegExp[]
```

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -244,8 +244,30 @@ export function doesDependencyChangeMatchPRTitle(
...
+): boolean {
+  let doesDepIncludeRegexToInclude = true;
```
chingor13, line 252: Is this correct? If there are no regexes, then this would return true?

packages/auto-approve/src/utils-for-pr-checking.ts
chingor13: Also, this could use a test if it's a re-usable helper

chingor13 (changes requested):

packages/auto-approve/src/utils-for-pr-checking.ts
sofisl: Exactly! If no regexes are provided, then there's no criteria to check the dependency against; so, it's trivially valid.

chingor13 (approved):
Note: we are meeting today to discuss refactoring to make these much simpler in the future.

PR #5003, opened 2023-03-17
feat: add auto label multiple config setting

### LINKED ISSUE #4992: Support for multiple labels based on path (auto-label bot)
For the purposes of improved labels management and auto tagging for DEE team I need a way for a given issue to have two labels (with two different prefixes) attached to it based on the path.

Ideally configuration shown in this [example](https://github.com/GoogleCloudPlatform/python-docs-samples/commit/8db954f01ca85f545b7d3b637cacd4fb0bbb292d) to work correctly by:
 - assigning label "api: recaptchaenterprise" to files located in `recaptcha_enterprise` folder
 - assigning labels "api: recaptchaenterprise" **AND** "asset: flagship" to files located in `recaptcha_enterprise/demosite` folder

Not considered, but ready to discuss.

 **Additional context**
Discussed via email with @sofisl. Sofia's comment: *"From a cursory glance it seems like we could iterate through several paths [here](https://github.com/googleapis/repo-automation-bots/blob/d55d03992c006bf200bd294477b320810010ad1f/packages/auto-label/src/auto-label.ts#L242), and then attach more than one label to the issue. There seems to be some logic inhibiting against attaching more than one label to the issue, but perhaps we could circumvent it by specifically adding a new "paths" section[...]"*

sofisl:
Fixes #4992 

cc @rsamborski


packages/auto-label/src/auto-label.ts
```diff
@@ -246,7 +246,7 @@ handler.autoLabelOnPR = async function autoLabelOnPR(
       logger.info(
         `Path label added to PR #${pull_number} in ${owner}/${repo} is ${path_label}`
       );
-      new_labels.push(path_label);
+      new_labels = path_label;
```
SurferJeffAtGoogle, line 249: I like const, because it's easier to reason about the code.  I'd prefer to leave line 238 as const, and change line 249 to 
```new_labels.push(...path_label);```

File: packages/auto-label/src/auto-label.ts [hunk 2]
```diff
@@ -265,7 +265,7 @@ handler.autoLabelOnPR = async function autoLabelOnPR(
       logger.info(
         `Language label added to PR #${pull_number} in ${owner}/${repo} is ${language_label}`
       );
-      new_labels.push(language_label);
+      new_labels.concat(language_label);
```
SurferJeffAtGoogle, line 268: This doesn't change either array.  I don't imagine that was intended because the return value containing the concatenated array is dropped.

I think this should be
```
new_leabels.push(...language_label);
```

SurferJeffAtGoogle (changes requested):

packages/auto-label/src/auto-label.ts
sofisl: I think, since we're definitely changing getLabel to return a string array, this is safe to do. The concern is that, with the old behavior, we could potentially return a string, and then adding a spread operator would add each character as an array value. But since we're switching over to returning an array, this is safe to do!

SurferJeffAtGoogle (approved):

Issue #5010, opened 2023-03-22
Possible race condition in OwlBot force pushes


codyoss:
The team observed that our owlbot PR was missing the commit description for one of the changes. Yesterday

The owlbot PR: https://github.com/googleapis/google-cloud-go/pull/7586
The missing commit description: https://github.com/googleapis/googleapis-gen/commit/b4ca7c760dbba430acf62b022ebe20693bb0aebe
The files were changed in the above PR, but without a trail of the commit: https://github.com/googleapis/google-cloud-go/pull/7586/files#diff-94ed2340518c4bc41382ce47d95309b8844cdc9feb69e0ef163230ceb2860808L20

In the PR I can see entries for 1e999b9 and 760597f, but not the missing commit in between.

## Other stuff

- The PR was force pushed to 7 times. It is possible that this force pushing
- The regenerate PR box was ticked multiple times
- There was a short period where our post processor and owlbot config both had issues that required us to have the PR re-generate.

PR #5011, opened 2023-03-24
fix: races while updating pull request branches

SurferJeffAtGoogle:
Fixed one race entirely.

It's not possible to fix another race without a
nonexistant variant of "git push -f" that first confirms the existing branch matches a particular commit hash.  So for that race, I minimized the time window.

Fixes https://github.com/googleapis/repo-automation-bots/issues/5010


SurferJeffAtGoogle:
Ready for review.  I added the do-not-merge label because I don't want to merge 'til Monday.

PR #5015, opened 2023-03-28
feat: update mono-repo-publish to leave a tarball for archiving

SurferJeffAtGoogle:
So we can generate SBOMs and attestations for the tarball.


chingor13 (approved):
Note that this is a releasable library, so merge it as a `fix` or `feat`

PR #5037, opened 2023-04-21
feat: add manual list of extra allowed api shortnames

### LINKED ISSUE #5036: [repo-metadata-lint] does not support non-gRPC APIs
The linter is using the [API index](https://github.com/googleapis/googleapis/blob/master/api-index-v1.json) to gather the list of valid api_ids (from scanning the googleapis/googleapis service configs).

https://github.com/googleapis/googleapis/commit/208d6b53fecb1278b74fc04bee98acc82369d915 removed the unavailable bigquery v2 gRPC endpoint which removed it from the API index.

chingor13:
Going with the manually curated list of extra allowed api_shortname entries because this scenario is so rare. If we allow every product entry, we could get false negatives where we happen to match a "product" but we don't have protos for it.

Fixes #5036

### DIFF
--- a/packages/repo-metadata-lint/src/validate.ts
+++ b/packages/repo-metadata-lint/src/validate.ts
@@ -31,6 +31,11 @@ const API_LIBRARY_TYPES = [
   'GAPIC_COMBO',
 ];
 
+// Manually curated list of allowed api_shortname entries
+const EXTRA_ALLOWED_API_SHORTNAMES = [
+  'bigquery', // handwritten client that has no protos
+];
+
 interface ApiIndex {
   apis: {hostName: string}[];
 }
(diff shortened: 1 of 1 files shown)


packages/repo-metadata-lint/src/validate.ts
```diff
@@ -102,6 +107,9 @@ export class Validate {
   async validApiShortNames() {
     const apiIndex = await this.getApiIndex();
     const apiShortNames = new Set<string>();
+    for (const shortname of EXTRA_ALLOWED_API_SHORTNAMES) {
```
SurferJeffAtGoogle, line 110: Pass EXTRA_ALLOWED_API_SHORTNAMES to the Set constructor, then no loop is necessary.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/Set

chingor13: Done.

SurferJeffAtGoogle (approved):

PR #5014, opened 2023-03-28
feat(snippet-bot): read products from service configs in googleapis/googleapis

chingor13:
Short term solution to merge new products from service_configs in googleapis/googleapis into the exported product list from DKP.

To improve performance, we use a daily cron job to scan googleapis/googleapis for service configs and merge the result with the exported list from DKP.

Fixes #4994

feat: add daily cron to reload products cache
refactor: pass ApiLabels list into violations logic to avoid implicit API call

PR #5038, opened 2023-04-21
feat: use combined DRIFT/api index product list

### LINKED ISSUE #5036: [repo-metadata-lint] does not support non-gRPC APIs

chingor13:
This implementation uses the combined product list which is maintained by snippet-bot (from #5014)

Fixes #5036


packages/repo-metadata-lint/src/validate.ts
```diff
@@ -31,20 +33,25 @@ const API_LIBRARY_TYPES = [
...
 export class Validate {
-  octokit: Octokit;
-  cachedApiIndex?: ApiIndex;
-  constructor(octokit: Octokit) {
-    this.octokit = octokit;
+  allowedApiShortnames: Set<string>;
+  constructor(allowedApiShortnames: Set<string>) {
+    this.allowedApiShortnames = allowedApiShortnames;
+  }
+  static async build(
```
SurferJeffAtGoogle, line 42: Why not add these arguments to the constructor?

chingor13: I'm not a huge fan of this class reaching out and doing stuff implicitly given an octokit or GCS connection/bucket.

The class really only cares about the allowed shortnames and the schema. Therefore, I opted to put the API calls into a helper method and pass only the data necessary for the validator into the class.

chingor13:
Closing in favor of #5037

PR #5047, opened 2023-05-03
refactor(auto-approve): create BaseLanguageRule, pass PR as an argument

sofisl:
- [x] Pass pull request as checkPR argument
- [x] Implements base language rule


chingor13 (changes requested):
Looking great!

Can you add some docs on `checkPR` for its inputs/outputs for posterity? (clarify that returning `true` means that the PR is ok to auto-approve) Also, can you add class docs for each of the process check classes to describe what they're used for (for future reviewers/maintainers).

chingor13 (approved):

Issue #5043, opened 2023-04-29
Blunderbuss: Multiples assignees for PRs


amanda-tarafa:
Use case: in sample repos, sample PRs should have two assignees one from the product reviewer pool and one from the language reviewer pool.

According to [documentation](https://github.com/googleapis/repo-automation-bots/tree/main/packages/blunderbuss#usage), Blunderbuss does not support this use case. `assign_prs_by` works well for assigning to a product reviewer, but if present, has precedence over `assign_prs` which means that can't be used for assigning to a second reviewer (language), but only as a fallback in case there are no matching product reviewers .

Possible solution (but anything that works best implementation wise and supports this use case is fine): introduce `assign_prs_always` so that a reviewer from this set is always assigned regardless of the presence of other assignees. This would be non-breaking in behaviour as `assign_prs` would continue to behave as a fallback with respect to `assign_prs_by`.

(It might be that I'm missreading docs and this is already supported.)

amanda-tarafa:
Amazing Jeff, thanks! Quick question since I'm not familiar with how the bots are released. When will this feature become usable? Can you ping this issue when that happens? Thanks again!

PR #5051, opened 2023-05-09
feat(blunderbuss): assign_prs_always in config

SurferJeffAtGoogle:
Fixes https://github.com/googleapis/repo-automation-bots/issues/5043

### DIFF
--- a/packages/blunderbuss/test/fixtures/config/pr_on_always.yml
+++ b/packages/blunderbuss/test/fixtures/config/pr_on_always.yml
@@ -1,0 +1,9 @@
+assign_prs_by:
+- labels:
+  - 'samples'
+  to:
+  - java-samples-reviewers
+assign_prs:
+- prs1
+assign_prs_always:
+- prs-always
--- a/packages/blunderbuss/__snapshots__/blunderbuss.js
+++ b/packages/blunderbuss/__snapshots__/blunderbuss.js
@@ -1,24 +1,25 @@
+exports['Blunderbuss pr tests assigns pr by label and always 1'] = {
+  "assignees": [
+    "java-samples-reviewers",
+    "prs-always"
+  ]
+}
+
 exports['Blunderbuss issue tests assigns opened issues with no assignees 1'] = {
   "assignees": [
     "issues1"
   ]
 }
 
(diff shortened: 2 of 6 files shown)


packages/blunderbuss/test/fixtures/config/pr_on_always.yml
```diff
@@ -0,0 +1,9 @@
+assign_prs_by:
+- labels:
+  - 'samples'
+  to:
+  - java-samples-reviewers
+assign_prs:
+- prs1
+assign_prs_always:
```
bcoe, line 8: Rather than introducing an additional `assign_prs_always` which seems a little redundant given that we have `assign_prs`, what if you instead introduce:

```
assignment_config:
  apply_all_pr_rules: true
```

Or something like this.

bcoe: If `apply_all_pr_rules` was set to `true`, it implies we should apply both `assign_prs_by` and `assign_prs`.

bcoe (changes requested):
Left a suggestion, let me know what you think.

Also, whatever we decide on, mind adding to `packages/blunderbuss/README`.

packages/blunderbuss/test/fixtures/config/pr_on_always.yml
SurferJeffAtGoogle: I prefer @amanda-tarafa's proposal.  What does @amanda-tarafa say?

amanda-tarafa: I also prefer my own proposal (haha) because it means that assign_prs can act as a fallback for assign_prs_by regardless of whether assign_prs_always is set. For instance, for products whose samples don't have an owner (i.e. we don't have an assign_prs_by), we still need two reviews, so we still need two assignees. We would get one from assign_prs (as a fallback for assign_prs_by) and one from assign_prs_always.

bcoe: I prefer my proposal, but am outvoted 😝 . Approved!

PR #5052, opened 2023-05-10
feat: triggerWithoutPullRequest

SurferJeffAtGoogle:


packages/release-trigger/src/bot.ts
```diff
@@ -44,12 +44,44 @@ import {
...
 const TRIGGER_LOCK_ACQUIRE_TIMEOUT_MS = 120 * 1000;
 
+/**
+ * Execute a function while holding a lock.
+ *
+ * @param targetUrl The lock's identifier.  Example: the pull request url.
+ * @param f The function to execute while holding the lock.
+ * @returns the value returned by f().
+ */
+async function withTriggerLock<R>(
```
chingor13, line 61: We could consider adding this to the DatastoreLock class as it seems pretty common (execute function under lock)

chingor13 (approved):

PR #5053, opened 2023-05-12
feat(auto-approve): add sample app repo processes

### LINKED ISSUE #5006: [auto-approve] feature requests for compatibility with sample applications
The DevRel team leverages Renovate to keep up-to-date with sample application dependencies but this can quickly become unwieldy. Therefore, I'm looking into leveraging auto-approval (couple with merge-on-green), but a few features are missing for our use cases.

Examples of sample application repositories that we'd like auto-approve help with:
- [Online Boutique](https://github.com/GoogleCloudPlatform/microservices-demo) (13,900+ stars ☆)
- [Bank of Anthos](https://github.com/GoogleCloudPlatform/bank-of-anthos) (680+ stars ☆)
- Potentially more sample application repositories down the line

I am happy helping out with creating PRs for these, but wanted to check in first.

### 1) DockerDependency non-existant
Currently, there are no type for Docker PRs.

**Suggestion**: Add a new type for this.

**Example PR**: https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1356

- no major bumps
- renovate
- no max files
- always a file named Dockerfile
- all dependencies match all files and the title of the PR

### 2) GolangDependency non-existant
Currently, there are no type for Golang PRs.

**Suggestion**: Add a new type for this.

**Example PR**: https://github.com/GoogleCloudPlatform/microservices-demo/pull/1597

- no major bumps
- renovate
- no max files
- always a file in go.sum or go.mod (go.mod always changes in the same ay, go.sum doesn't)
- all dependencies match all files and the title of the PR

-- Modifications of existing types ---
### 3) Maximum number of files on all deps too small
The current maximum number of files for each dependency type is set to 3. For sample applications composed of multiple services (and multiple services of the same language) this means that a majority of Renovate PRs could never get auto-approved.

**Suggestion**: Removing that constraint, or bumping it higher.

**Example PR**: https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1355
- go, python, node, java
- not merging major,
- no max files

### 4) PythonDependency doesn't support .in files
There is currently a constraint that only `requirements.txt` files are supported for Python dependencies. In some of our repositories, we also use a `requirements.in` file for package management (see [pip-tools](https://pypi.org/project/pip-tools/)).

**Suggestion**: Adding `requirements.in` as a valid file matching.

**Example PR**: https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1358

### 5) PythonDependency filepath matching too strict
There is currently a constraint that all filepath must match `/^samples/snippets/requirements.txt$/`. This is too strict for our repositories, whose Python's dependencies files live in more arbitrary locations.

**Suggestion**: Removing that constraint.

**Example PR**: https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1361

-ends in requirements.txt, or .in

sofisl:
Fixes #5006 

cc @bourgeoisor 

- Adds a Go Dependency process
- Adds a Docker Dependency process
- Adds a Python Sample App (specific to the requirements in the bug) process
- Updates the language for Java Dependency (it was 50, documentation still said 3)

Will remove the Node Dependency requirement in a F/U PR (the Node team does not need max files)


packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -413,6 +527,25 @@ export function runVersioningValidation(versions: Versions): boolean {
...
+/**
+ * Runs additional validation checks when a version is upgraded to ensure that the
+ * version is only upgraded, not downgraded, and that the major version is not bumped.
+ * If the PR could have a rev or sha (as in Docker or Go PRs), this checks that
+ * the minor version is bumped, and/or the rev or sha is upgraded
+ *
+ * @param versions the versions object returned from getVersions, getVersionsV2, and getJavaVersions
+ * @returns true if the version is only bumped a minor, not a major
+ */
+export function runVersioningValidationWithShaOrRev(
```
chingor13, line 539: The naming of this method does not help readability. A better common naming scheme IMO for something returning a true/false value would be `isValidSomething`.

packages/auto-approve/__snapshots__/auto-approve.test.js
chingor13, line 1: Can we revert these changes or fix the formatting?

packages/auto-approve/src/interfaces.ts
```diff
@@ -55,6 +55,8 @@ export interface Versions {
   oldMinorVersion: string;
   newMajorVersion: string;
   newMinorVersion: string;
+  oldShaOrRevTag?: string;
```
chingor13, line 58: Is there a need to overload this class? Can we have a separate class for a SHA diff. The function that parses the diff could return a `Versions | ShaDiff` type.

chingor13 (changes requested):
Can we break this into smaller PRs for unrelated things? It makes the reviews easier to follow and rollback if necessary.

packages/auto-approve/src/interfaces.ts
sofisl: Sounds good!

packages/auto-approve/src/utils-for-pr-checking.ts
```diff
@@ -280,12 +365,41 @@ export function doesDependencyChangeMatchPRTitleV2(
...
+ * This function checks whether the dependency stated in a given title was the one that was changed for Go
+ * This function is different from doesDependencyChangeMatchPRTitle as it is specific to Go, since those titles
+ * vary in their regex.
+ *
+ * @param versions the Versions object that contains the old dependency name and new dependency name and versions
+ * @param dependencyRegex the regular exp to find the dependency within the title of the PR
+ * @param title the title of the PR
+ * @returns whether the old dependency, new dependency, and dependency in the title all match
+ */
+export function doesDependencyChangeMatchPRTitleGo(
```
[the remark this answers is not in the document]
sofisl: Done. I still needed to add a dependency regex parameter because there is a "file matching" process. If we ever want to add more specific rules to a different file (not a go.mod), then we might need to pass in a different regex.

sofisl:
> Can we break this into smaller PRs for unrelated things? It makes the reviews easier to follow and rollback if necessary.

I removed the change to the Node dependency max files, since that is the only backwards-looking fix (everything else is a feature). Let me know if you want me to do more for readability.

chingor13 (changes requested):

chingor13 (approved):

PR #5054, opened 2023-05-15
feat: add LOCAL_MODE option to bazel-bot's generate script

chingor13:
For locally unsticking bazel-bot, you can run docker locally. By setting the `LOCAL_MODE` env variable to `true`, we skip pulling/pushing to git. It will generate locally and you can then fix the commit message.

Run the following command from a directory that has both googleapis (at the commit you want to build from) and googleapis-gen (at head) repos checked out. This will run bazel-bot's build and commit mechanism without attempting to push changes. You can then go to the googleapis-gen directory and amend the commit to your liking.
```
docker run --rm -it \
  -v $(pwd)/googleapis:/workspace/googleapis \
  -v $(pwd)/googleapis-gen:/workspace/googleapis-gen \
  --workdir /workspace
  --env BROKEN_SHAS="<list-of-shas>" \
  --env LOCAL_MODE=true \
  gcr.io/repo-automation-bots/bazel-bot
```

### DIFF
--- a/packages/bazel-bot/docker-image/generate-googleapis-gen.sh
+++ b/packages/bazel-bot/docker-image/generate-googleapis-gen.sh
@@ -26,6 +26,9 @@
 #
 # BAZEL_FLAGS: additional flags to pass to 'bazel query' and 'bazel build'.
 # Useful for setting a remote cache, coping with different versions of bazel, etc.
+#
+# LOCAL_MODE: if set to "true", then skip git pull and git push. Useful for generating
+# locally in docker with mounted directories.
 
 # Fail immediately.
 set -e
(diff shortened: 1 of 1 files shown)


SurferJeffAtGoogle:
To solve this same problem, I ran bazel locally.  And I ran it with a single build target so it revealed the error in a minute or two the first time, and about 10 seconds for each commit hash I tried.  To me, that was a simpler process than using a local mode of the docker image.

chingor13:
This is also for the untarring and copying to the right spot in googleapis-gen. Is there an easy workaround for that part too?

The language maintainers preferred having the correct context on a commit (e.g. for a valid commit that happened while bazel was broken)

chingor13:
Closing in favor of #5056

PR #5057, opened 2023-05-16
feat: withDatastoreLock()

SurferJeffAtGoogle:
Makes it easier to acquire and release a datastore lock.


packages/datastore-lock/src/datastore-lock.ts
```diff
@@ -207,3 +207,56 @@ export class DatastoreLock {
     else return false;
   }
 }
+
+/**
+ * Parameters for creating a DatastoreLock.
+ */
+export interface DatastoreLockDetails {
+  lockId: string;
```
SurferJeffAtGoogle, line 215: @chingor13 What doc comment should I add for `lockId` and `target`?  There are no existing comments describing them.  I'm not sure of their purpose.

SurferJeffAtGoogle: I think I figured it out.

chingor13 (approved):

PR #5060, opened 2023-05-17
fix(auto-approve): update node dependency for mono repo migration

sofisl:
In moving to the mono-repo, we no longer have package.json in one fixed spot; this PR updates the rules to reflect that. Also, we may have more than 3 files being changed, so this updates that rule.

### DIFF
--- a/packages/auto-approve/README.md
+++ b/packages/auto-approve/README.md
@@ -83,8 +83,7 @@ Below is what each process checks for:
     - /package\.json$/
   - All files must: 
     - Match either these regexp:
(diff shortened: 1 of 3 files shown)


bcoe (commented):
We still have some split repos like Storage, will this configuration work for both split repos and the monorepo?

sofisl:
> We still have some split repos like Storage, will this configuration work for both split repos and the monorepo?

Yes! Technically it is less restrictive so it will work with split repos. Also, I don't think any split repos are currently using it.

bcoe (approved):

PR #5064, opened 2023-05-18
fix(auto-approve): update regex for Node

sofisl:
Some strange PRs have prompted looking into the pattern being used.

### DIFF
--- a/packages/auto-approve/src/process-checks/node/dependency.ts
+++ b/packages/auto-approve/src/process-checks/node/dependency.ts
@@ -44,16 +44,19 @@ import {BaseLanguageRule} from '../base';
 export class NodeDependency extends BaseLanguageRule {
   classRule = {
     author: 'renovate-bot',
-    titleRegex: /^(fix|chore)\(deps\): update dependency (@?\S*) to v(\S*)$/,
+    titleRegex:
+      // This would match: fix(deps): update dependency @octokit/rest to v19.0.8 or ^0.23.0 or ~0.23.0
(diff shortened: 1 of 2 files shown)


packages/auto-approve/src/process-checks/node/dependency.ts
```diff
@@ -44,16 +44,19 @@ import {BaseLanguageRule} from '../base';
 export class NodeDependency extends BaseLanguageRule {
   classRule = {
     author: 'renovate-bot',
-    titleRegex: /^(fix|chore)\(deps\): update dependency (@?\S*) to v(\S*)$/,
+    titleRegex:
+      // This would match: fix(deps): update dependency @octokit/rest to v19.0.8 or ^0.23.0 or ~0.23.0
+      /^(fix|chore)\(deps\): update dependency (@?\S*) to v?\^?~?(\S*)$/,
```
chingor13, line 49: If we know there's at least one of `v`, `^`, or `~`, put it in a character group:
```suggestion
      /^(fix|chore)\(deps\): update dependency (@?\S*) to [v\^~](\S*)$/,
```

chingor13 (changes requested):

packages/auto-approve/src/process-checks/node/dependency.ts
sofisl: I don't think we know for sure, although I haven't seen it without one of those yet. But I think it would be safer the way I have it, since then we could just accept the number, for example:

`fix(deps): update dependency recast to 0.23.0`

PR #5081, opened 2023-06-02
feat: store validated .repo-metadata in table

bcoe:
Store validated fields from .repo-metadata.json in a BigQuery table, so that it can feed into team analytics, e.g., library coverage.

### DIFF
--- a/packages/repo-metadata-lint/src/store-metadata.ts
+++ b/packages/repo-metadata-lint/src/store-metadata.ts
@@ -1,0 +1,51 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


packages/repo-metadata-lint/src/store-metadata.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export async function storeMetadata(
+  metadata: Metadata,
+  logger: GCFLogger = defaultLogger
+) {
+  try {
+    await bigquery
+      .dataset(datasetId)
+      .table(tableId)
+      .insert([{timestamp: new Date(), ...metadata}]);
```
chingor13: Would a date partitioned dataset work?

bcoe: Okay, let's try a partitioned table.

packages/repo-metadata-lint/src/validate.ts
```diff
@@ -99,6 +100,17 @@ export class Validate {
...
 
+    // On success, store an entry in a metadata table. This data is
+    // used to answer questions such as API coverage:
+    if (result.status === 'success') {
+      await StoreMetadata.storeMetadata({
+        release_level: repoMetadata.release_level,
+        language: repoMetadata.language,
+        repository: repoMetadata.repository,
+        api_service: `${repoMetadata.api_shortname}.googleapis.com`,
+      });
```
chingor13, line 111: Consider also storing the raw data as a string in a field. BQ has helpers to parse JSON and read fields to avoid the JSON not having a strict schema

chingor13 (approved):

Issue #5098, opened 2023-07-05
[auto-approve] Java dependencies PRs not always auto-approved


bourgeoisor:
It seems that Java dependencies that should be auto-approved are not always.

e.g. https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1579 (can provide more examples if needed)

Perhaps it's because of this: "Change the dependency that was there previously, and that is on the title of the PR, **and is a Google dependency**". (emphasis mine) What does "Google dependency" mean in this context?

PR #5153, opened 2023-07-18
feat: add Java sample app dependency

sofisl:
Fixes https://github.com/googleapis/repo-automation-bots/issues/5098


bcoe:
@chingor13 is a better reviewer, given my calendar this week. Thanks for this awesome work supporting @bourgeoisor!

packages/auto-approve/src/process-checks/sample-application-repos/java-dependency.ts
```diff
@@ -0,0 +1,171 @@
...
+      }
+
+      const doesDependencyMatch = doesDependencyChangeMatchPRTitleV2(
+        versions,
+        // We can assert this exists since we're in the class rule that contains it
+        fileMatch.dependencyTitle!,
+        incomingPR.title
+      );
+
+      const isVersionValid = runVersioningValidation(versions);
```
chingor13, line 137: It would still be nice to do the refactor we had planned or at least make this more readable. It's hard to tell at a glance what this is validating.

packages/auto-approve/src/process-checks/sample-application-repos/java-dependency.ts
```diff
@@ -0,0 +1,171 @@
+// Copyright 2023 Google LLC
```
chingor13, line 1: nit: the file name should match the class name better - it would help make it easier to discover where the class is defined.

chingor13 (approved):

Issue #5034, opened 2023-04-18
owlbot: stage PR as draft intially


codyoss:
**Is your feature request related to a problem? Please describe.**

Too many people initial get tagged into OwlBot PRs. This is due to extra diffs existing that get filtered out with post processing. One example we have in Go is generally we format all code in our repo. The proto libraries, not gapic, are not formatted to common standards and we don't control the bazel rule for this target as it is the common protoc go rule. Because of all of these diffs in a monorepo when the PR open everyone gets tagged into review the PR. After post-processing most people would not have actually needed to be tagged in as reviews.

 **Describe the solution you'd like**

Maybe owlbot could stage PR as a draft initially and then open for review after a successful post-processor run.

 **Describe alternatives you've considered**

- I am not sure it has been considered in the past but I also thought about framing this request as owl-bot and post processor changes should happen in the same commit. This imo would be the best option, but I assume there are reasons things they are right now. This would reduce the noise on these PRs massively though and be better for infra costs as things don't get started running on one commit and then start another job on the post-processed commit. Also may help save quotas on constrained projects
- Keep things they way they are now and tell most people to just ignore this PR always

 **Additional context**

jskeet:
We've just seen these in .NET - the "create as draft" part seems to be working, but they're not being converted to "normal" PRs after the successful post-processor run. Looks like it's happening for other languages, too...

SurferJeffAtGoogle:
Ouch.  I have to use the graphql API to remove the draft tag from a PR.

SurferJeffAtGoogle:
I'm afraid that implementing this feature request is not possible with the current GitHub API.

Even the graphql API doesn't provide away to clear the draft flag from a pull request. 

updatePullRequest  doesn't have a draft flag:
https://docs.github.com/en/graphql/reference/objects#pullrequest

There is a convertPullRequestToDraft:
https://docs.github.com/en/graphql/reference/mutations#convertpullrequesttodraft
but there is no convertPullRequestFromDraft.

codyoss:
@SurferJeffAtGoogle This is possible, here is the Go code I used to use to do something similar: https://github.com/googleapis/google-cloud-go/blob/e039e2b1c0b4062706948fbadf424e94d88b5f57/internal/gapicgen/git/github.go#L370-L385

SurferJeffAtGoogle:
From the GitHub documentation:
clientMutationId (String) | A unique identifier for the client performing the mutation.

[...]

PR #5156, opened 2023-07-19
fix: flag to enable creating draft pull requests

SurferJeffAtGoogle:
Step 2 of 3 towards fixing #5034

### DIFF
--- a/packages/owl-bot/src/copy-code.ts
+++ b/packages/owl-bot/src/copy-code.ts
@@ -426,13 +426,15 @@ interface WorkingCopyParams extends CopyParams {
  * googleapis-gen.  Creates pull request if there's none matching and open.
  *
  * @param yamlPaths .OwlBot.yaml paths that triggered this copy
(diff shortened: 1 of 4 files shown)


packages/owl-bot/src/copy-code.ts
```diff
@@ -580,7 +582,9 @@ command in a local clone of this repo:
       WithRegenerateCheckbox.Yes,
       apiList,
       Force.Yes,
-      logger
+      logger,
+      undefined,
+      draftPulledRequests
```
chingor13, line 587: Consider refactoring to an options hash with defaults to avoid needing to specify N optional fields positionally.

chingor13 (approved):

packages/owl-bot/src/copy-code.ts
SurferJeffAtGoogle: that sounds appealing

PR #5172, opened 2023-08-01
deps: update wordwrap dependency

chingor13:
Ran `npm audit fix`


SurferJeffAtGoogle:
I worry that @chingor13 is not enjoying his vacation as much as he could because he's still merging code!  😊

Issue #5179, opened 2023-08-07
owlbot processing not starting, PR stuck as draft with staging directory


alexander-fenster:
PR: https://github.com/googleapis/google-cloud-node/pull/4498

Created 2+ hours ago with `owl-bot-staging` directory; the processing has never started and there is no note from Owlbot left in the CI section.

I know for sure that if I add `owlbot:run` to that PR it will get processed, but leaving it as is for now so you could investigate.

SurferJeffAtGoogle:
I expect https://github.com/googleapis/repo-automation-bots/pull/5181 to have fixed it.  If not, I'll flip a flag so Owl Bot generates pull requests that are not drafts.

SurferJeffAtGoogle:
Yay!  Owl Bot ran the post processor for this draft PR:
https://github.com/googleapis/google-cloud-node/pull/4536

PR #5184, opened 2023-08-10
fix: trigger post processor when owl-bot-run label was added to a dra…

SurferJeffAtGoogle:
…ft PR

Fixes https://github.com/googleapis/repo-automation-bots/issues/5179

### DIFF
--- a/packages/owl-bot/src/owl-bot.ts
+++ b/packages/owl-bot/src/owl-bot.ts
@@ -382,39 +382,28 @@ async function handlePullRequestLabeled(
     return;
   }
 
-  const command_labels = [OWLBOT_RUN_LABEL, OWL_BOT_COPY_COMMAND_LABEL];
+  const commandLabels = [OWLBOT_RUN_LABEL, OWL_BOT_COPY_COMMAND_LABEL];
 
-  // Only run on label event if the label added is the owlbot label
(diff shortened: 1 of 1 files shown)


bcoe (approved):

packages/owl-bot/src/owl-bot.ts
```diff
@@ -382,40 +382,29 @@ async function handlePullRequestLabeled(
...
 
-  const command_labels = [OWLBOT_RUN_LABEL, OWL_BOT_COPY_COMMAND_LABEL];
-
-  // Only run on label event if the label added is the owlbot label
-  if (!command_labels.includes(payload.label.name)) {
-    logger.info(
-      `skipping non-owlbot label: ${payload.label.name} ${head} for ${base}`
+  const commandLabels = [OWLBOT_RUN_LABEL, OWL_BOT_COPY_COMMAND_LABEL];
+
+  if (payload.pull_request.draft && payload.label.name === OWL_BOT_COPY) {
```
bcoe, line 387: nit, the `noop` is a bit confusing, I might change this logic around a bit so the "non-owlbot" case breaks, rather than the happy path doing nothing.

bcoe (approved):
LGTM. A test would be nice, but not a blocker (if you're itching to see this work end to end).

PR #5195, opened 2023-09-06
refactor: use sinon to mock octokit methods

chingor13:
feat!: require node 18
feat!: remove `addOrUpdateIssueComment` which is moved to issue-utils.

### DIFF
--- a/packages/gcf-utils/src/configuration.ts
+++ b/packages/gcf-utils/src/configuration.ts
@@ -1,0 +1,36 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+export const RUNNING_IN_TEST = process.env.NODE_ENV === 'test';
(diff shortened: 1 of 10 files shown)


packages/gcf-utils/src/configuration.ts
```diff
@@ -0,0 +1,36 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+export const RUNNING_IN_TEST = process.env.NODE_ENV === 'test';
+// Adding 30 second delay for each batch with 30 tasks
+export const DEFAULT_FLOW_CONTROL_DELAY_IN_SECOND = 30;
+
+export interface WrapConfig {
```
SurferJeffAtGoogle, line 20: Please add a comment for this interface.

packages/gcf-utils/src/installations.ts
```diff
@@ -0,0 +1,89 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import {WrapConfig} from './configuration';
+
+export interface AppInstallation {
+  id: number;
+  targetType: string;
```
SurferJeffAtGoogle, line 20: Please add a comment for the targetType and login members.  I can guess what the other two mean.

SurferJeffAtGoogle (changes requested):

packages/gcf-utils/src/configuration.ts
chingor13: Done

packages/gcf-utils/src/installations.ts
chingor13: Done

SurferJeffAtGoogle (approved):

PR #5201, opened 2023-09-12
build: owlbot templating should ignore release-please config

chingor13:


chingor13:
Looks like updating release-please-config.json is not a template

PR #5217, opened 2023-09-18
feat: migrate addOrUpdateIssueComment to issue-utils

chingor13:
This function was removed from gcf-utils v15 as it's not a core method needed by each bot.

### DIFF
--- a/packages/issue-utils/src/issue-comments.ts
+++ b/packages/issue-utils/src/issue-comments.ts
@@ -1,0 +1,104 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


packages/issue-utils/src/issue-comments.ts
```diff
@@ -0,0 +1,104 @@
...
+// limitations under the License.
+
+// eslint-disable-next-line node/no-extraneous-import
+import {Octokit} from '@octokit/rest';
+
+/**
+ * It creates a comment string used for `addOrUpdateissuecomment`.
+ */
+export const getCommentMark = (installationId: number): string => {
+  return `<!-- probot comment [${installationId}]-->`;
```
sofisl, line 22: I wonder if it's worth including a timestamp or updating a counter for how many times it's commented, to make sure it's run recently.

sofisl (approved):

Issue #5193, opened 2023-09-06
MoG not merging


codyoss:
I noticed the eyeballs emoji was added to track a PR but it sat over 3 hours approved before I merged it by hand. Saw a similar report from another Googler as well.

The PR: https://github.com/googleapis/google-cloud-go/pull/8508

SurferJeffAtGoogle:
Here's the final block of comments left by merge on green for this PR.  @codyoss, does this agree with your understanding of the state of the PR at 11:30am (18:30Z)?

```
INFO 2023-09-06T18:30:12.396926Z [resource.labels.functionName: merge_on_green] [labels.executionId: xpo2sf2t6f1g] === checking required reviews googleapis/google-cloud-go/8508 ===
INFO 2023-09-06T18:30:12.611116Z [resource.labels.functionName: merge_on_green] [labels.executionId: xpo2sf2t6f1g] fetched statusesForRef in 214ms googleapis/google-cloud-go/8508
INFO 2023-09-06T18:30:12.611146Z [resource.labels.functionName: merge_on_green] [labels.executionId: xpo2sf2t6f1g] === checking required checks for googleapis/google-cloud-go/8508 ===
INFO 2023-09-06T18:30:12.611173Z [resource.labels.functionName: merge_on_green] [labels.executionId: xpo2sf2t6f1g] Looking for required checks in status checks for googleapis/google-cloud-go/8508.
INFO 2023-09-06T18:30:12.845588Z [resource.labels.functionName: merge_on_green] [labels.executionId: xpo2sf2t6f1g] fetched completed reviews in 449ms googleapis/google-cloud-go/8508
INFO 2023-09-06T18:30:12.845631Z [resource.labels.functionName: merge_on_green] [labels.executionId: xpo2sf2t6f1g] One of your reviewers did not approve the PR googleapis/google-cloud-go/8508 state = COMMENTED
INFO 2023-09-06T18:30:12.994199Z [resource.labels.functionName: merge_on_green] [labels.executionId: xpo2sf2t6f1g] Looking for required checks in status checks for googleapis/google-cloud-go/8508.
INFO 2023-09-06T18:30:12.994231Z [resource.labels.functionName: merge_on_green] [labels.executionId: xpo2sf2t6f1g] Looking for required checks in status checks for googleapis/google-cloud-go/8508.
[...]

PR #5231, opened 2023-09-19
fix(merge-on-green)!: remove restrictions for checking reviews for merge on green

sofisl:
When we initially created merge-on-green, we weren't sure whether repos had set up branch protection adequately, so we wrote additional logic to make sure PRs had reviews. However, we've since instituted org-wide protection, this seems like an additional piece of logic now that causes a lot of confusion, see: https://github.com/googleapis/repo-automation-bots/issues/5193


packages/merge-on-green/test/merge-on-green.ts
```diff
@@ -422,7 +422,7 @@ describe('merge-on-green wrapper logic', () => {
 
         const [reducedPRs] = handler.maybeReducePRList(prs);
 
-        assert.strictEqual(reducedPRs.length, 25);
+        assert.ok(reducedPRs.length === 24 || reducedPRs.length === 25);
```
chingor13, line 425: We should be confident in this value.

chingor13 (changes requested):
Main change seems ok, but a nit about the test

packages/merge-on-green/test/merge-on-green.ts
sofisl: I think it's OK if we're not confident. It's produced by a random value, and the only requirement is that it has to be under 25 (the chunk size): https://github.com/googleapis/repo-automation-bots/blob/ce561fe11471717eaa93aea7cdf2fe2922140b23/packages/merge-on-green/src/merge-on-green.ts#L87. I've updated the test to reflect that.

sofisl:
Friendly ping @chingor13

Iben1993:
> Amikor kezdetben az egyesületet--zöldre hoztuk létre, nem voltunk biztosak abban, hogy a repos megfelelően létrehozta-e a fióktelepek védelmét, ezért további logikát írtunk annak érdekében, hogy a PRok legyenek felülvizsgálatok. Azonban, azóta, hogy bevezettük az org-wide védelmet, ez úgy tűnik, mint egy további logika most, amely sok zűrzavart okoz, lásd: #5193

Iben1993:
LICENSE

chingor13 (approved):

PR #5236, opened 2023-09-25
feat: introduce new bot-framework library

chingor13:
Reimplements gcf-utils as bot-framework without probot as a dependency. `@octokit/webhooks` already handles webhook routing and signature verification.


packages/bot-framework/Dockerfile
chingor13, line 1: This was for local testing and can be removed later.

packages/bot-framework/src/appserver.ts
chingor13, line 1: This is for local testing and can be removed later.

packages/bot-framework/README.md
```diff
@@ -0,0 +1,18 @@
+# foo
```
SurferJeffAtGoogle, line 1: Imagine I want to write a new bot.*  Please describe the list of services provided by this library and why I would want to use them.

* I know we said we never want to write a new bot, but I think imagining we might will result in a README that better helps maintainers.

SurferJeffAtGoogle (changes requested):

PR #5263, opened 2023-11-20
fix(auto-approve): update Docker dependency regex to include another regex use case

### LINKED ISSUE #5245: [auto-approve] Docker deps not matching built-in regex
I've noticed over the past week that a handful of Docker deps PR from Renovate don't get auto-approved.

From a bit of investigation, I wonder if they're just not matching the auto-approve regex correctly (or rather, the regex is not capturing all Docker deps PRs).

From the [README docs](https://github.com/googleapis/repo-automation-bots/blob/main/packages/auto-approve/README.md), the current regex is:
```
/^(fix|chore)(deps): update (\D[^:?]). docker (digest|tag) to (.*)$/
```

But the title of Renovate PRs don't match it (example PR: https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/1855)

I've managed to match the example PR above by using this slightly modified regex:
```
/^(fix|chore)\(deps\): update (.*) docker (digest|tag) to (.*)$/
```

Changes:
- `(deps)` -> `\(deps\)` since the brackets are expected to be literals, not a capture group
- `(\D[^:?]).` -> `(.*)`; not sure why e.g. the `:` was explicitly disallow, since it makes titles with `package:version` fail the regex

sofisl:
Fixes #5245 

@bourgeoisor, I added the PR you linked as a test case and found out that it's not the title regex that fails, but rather the regex in the file validation. It's expecting two decimal points instead of just one, so I updated the regex appropriately.


bourgeoisor (approved):
Yay! LGTM, thank you!

PR #5286, opened 2024-02-07
fix(owl-bot): ignore non-cloud php repos

chingor13:
owl-bot is opening some PRs against sharded, non-php repos. We should not be opening copy-code PRs against these repos.

example: https://github.com/googleapis/php-shopping-merchant-reports/pull/5

These repos `googleapis/php-<something>`, are non-cloud and split of the primary PHP monorepo `googleapis/google-cloud-php`.

### DIFF
--- a/packages/owl-bot/src/should-ignore-repo.ts
+++ b/packages/owl-bot/src/should-ignore-repo.ts
@@ -32,6 +32,6 @@ export function shouldIgnoreRepo(ownerSlashRepo: string): boolean {
   // subdirectories of google-cloud-php and never accept pull requests.
   const regexp = process.env.IGNORE_REPO_REGEXP
     ? new RegExp(process.env.IGNORE_REPO_REGEXP)
-    : /googleapis\/(googleapis(-gen)?|google-cloud-php-.+)/;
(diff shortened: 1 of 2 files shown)


chingor13:
Note that this won't close existing open PRs, but should prevent new PRs from being opened (after the scan GitHub configs cron task runs)

bshaffer (approved):
LGTM, I have verified all existing `googleapis/php-*` repos are subrepos and should be ignored. The only non-monorepo` owlbot repo is `googleapis/common-protos-php` (which at some point we _should_ move under `googleapis/google-cloud-php`).

packages/owl-bot/src/should-ignore-repo.ts
```diff
@@ -32,6 +32,6 @@ export function shouldIgnoreRepo(ownerSlashRepo: string): boolean {
   // subdirectories of google-cloud-php and never accept pull requests.
   const regexp = process.env.IGNORE_REPO_REGEXP
```
sofisl, line 33: Seems like these should always be ignored, not only if they're not explicitly specified.

chingor13: I think is an override for tests/testing environments and we use the default in production

sofisl (approved):

PR #5303, opened 2024-03-21
feat: add ability to load GitHub credentials via provided private key

chingor13:
feat: add `install-credentials` command

This will allow us to not pass in credentials as part of the build steps, but rather load them from secrets injected from secret manager.


chingor13:
This method is blocked on access issues from granting the Cloud Build job secret manager access

PR #5351, opened 2024-06-18
fix: log bazel-bot runs only to Cloud Logging

chingor13:


chingor13:
Closing in favor of #5352

Issue #5355, opened 2024-06-20
consider adding '@octokit/plugin-retry'


codyoss:
Noticed some recent failures that I assume were transient 500s. It would be nice if we had retires built into octokit calls. Found the above module when searching around and it seemed like a good fit.

PR #5356, opened 2024-06-20
feat(owl-bot): add default retry policy for octokit usage

codyoss:
Fixes: #5355


codyoss:
I just manually looked up versions of the deps the seemed compatible with our octokit-rest usage and added them with `npm install @dep@version` -- not sure there is a better way.

Also I saw this in the readme for the retry project: "As we use conditional exports, you will need to adapt your tsconfig.json by setting "moduleResolution": "node16", "module": "node16"."

Tests seemed to pass without this so I did not add, but I wan't to highlight that as it is gibberish to me

chingor13 (approved):

PR #5364, opened 2024-07-08
build(release-trigger): build from official node image

chingor13:


packages/release-trigger/Dockerfile
```diff
@@ -14,9 +14,33 @@
 
 # Use a multi-stage docker build to limit production dependencies.
 
+FROM node:18.20.3-bullseye AS BASE
```
meltsufin, line 17: Can we setup Renovate to suggest updates to this version?

packages/release-trigger/Dockerfile
```diff
@@ -14,9 +14,33 @@
...
+RUN git clone https://github.com/pyenv/pyenv.git ~/.pyenv && \
+    echo 'export PATH="/home/node/.pyenv/bin:$PATH"' >> ~/.profile && \
+    echo 'export PATH="/home/node/.pyenv/shims:$PATH"' >> ~/.profile && \
+    echo 'eval "$(pyenv init -)"' >> ~/.profile && \
+    echo 'eval "$(pyenv virtualenv-init -)"' >> ~/.profile
+
+ENV PATH="/home/node/.pyenv/bin:/home/node/.pyenv/shims:${PATH}"
+
+# Install python
+RUN pyenv install 3.10.13 && \
```
meltsufin, line 37: Also here, maybe Renovate can be configured to manage this version.

meltsufin (approved):

packages/release-trigger/Dockerfile
chingor13: Can we do it separately? Also updating a bunch of images in #5365

packages/release-trigger/Dockerfile
meltsufin: Yes, let's track it as a separate issue.

chingor13: #5366 5366

sofisl (approved):

PR #5365, opened 2024-07-08
build: upgrade to 18.20.3 base image

chingor13:


meltsufin (approved):
LGTM, but we should have a way to get Renovate to manage this version.

chingor13:
> LGTM, but we should have a way to get Renovate to manage this version.

#5366

sofisl (approved):

PR #5369, opened 2024-07-09
build: enable renovate

chingor13:
Use renovate to update versioned Dockerfile base images.

Note: using `.json5` for a more readable config format

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -1,0 +1,7 @@
+{
+  "extends": [
+    // update base images in Dockerfiles
+    ":docker",
+    ":disableDependencyDashboard",
+  ]
+}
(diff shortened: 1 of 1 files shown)


renovate.json5
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": [
+    // update base images in Dockerfiles
+    ":docker",
+    ":disableDependencyDashboard",
```
meltsufin, line 5: Why disable it? It's a good way to check if Renovate is discovering the places to update correctly.

meltsufin (approved):

renovate.json5
chingor13: Done, put it back

sofisl (approved):

PR #5385, opened 2024-07-10
chore: remove CODEOWNERS to avoid tagging everyone for review

meltsufin:


codyoss:
This does more than tag people though. This says who the required reviewers are for a directory. This this desirable?

meltsufin:
Do we really want to block PRs on individuals or can any of the committers approve the PRs across the repo?

codyoss:
The codeowners line are an or. So it says the group or the person can approve. That said it seems like this CO could be simplified to `* @googleapis/github-automation` for this repo. In my mind there is a difference between write access to the the repo and being a reviewer for the project. I think we may require the existence of this file also, but @chingor13 may know better than I.

chingor13:
The GitHub automation team should be required review for most code in this repository. If other teams/maintainers own a subfolder, that will override the root ownership.

PR #5392, opened 2024-08-04
fix: trigger header-checker-lint bot on PR labeling

### LINKED ISSUE #5391: license-header-lint: Support forced re-triggering functionality to enable re-launch bot if it gets stuck
In some pull requests (PRs) the bot isn't triggered during a long period of time or forever, effectively locking PR in the unmergable state. The only working method is to close PR and re-open it. In this event the license-header-lint check is re-triggered.

Introduce a custom label (e.g. `license-header-lint:force-run`) that an authorized user can add to PR that will cause to the bot to relaunch on this PR.

While it is possible to search for the reason why the bot get stuck on certain PRs this investigation is time consuming and its result is uncertain.

 **Additional context**
None

minherz:
### Description

Adds triggering upon adding a label to [pull_request][1] `pull_request.labeled`.
Runs the validation if the label `'header-checker-lint:force-run'` is added to PR.
Updates README to add the description of the new behavior.
Adds explicit `octokit` dependencies to `package.json`.

Fixes #5391

[1]: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

### DIFF
--- a/.github/workflows/audit.yaml
+++ b/.github/workflows/audit.yaml
@@ -10,12 +10,12 @@ jobs:
     outputs:
       nodePaths: ${{ steps.interrogate.outputs.nodePaths }}
     steps:
-      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
         with:
           fetch-depth: 0
       - uses: actions/setup-node@1a4442cacd436585916779262731d5b162bc6ec7 # v3
         with:
-          node-version: 14
+          node-version: 18
       - id: interrogate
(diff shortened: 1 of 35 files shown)


minherz:
@chingor13 it looks like the failing tests do not fail due to the code I added. The tests seem outdated to me. Please, advise about which line of actions you prefer:

* Ignore the error `TS2322` that is the common error for all failures
* Mark the failing tests to be ignored
* Rewrite all tests with a focus on validating the logic of header verification
* Suggest a fix for these errors

Thank you

chingor13 (changes requested):
This is a compilation error because of octokit types not matching with probot. It is non-trivial to update octokit packages

chingor13 (approved):

PR #5557, opened 2024-11-07
fix: restrict snippet bot to only be allowed to run for google org

ldetmer:


packages/snippet-bot/cron.yaml
```diff
@@ -4,6 +4,9 @@ cron:
     description: snippet-bot syncing labels
     params:
       cron_type: "repository"
+      allowed_organizations:
```
chingor13, line 7: note: I don't think the cron helper will overwrite the existing scheduler job

chingor13 (approved):
I don't think both checks are necessary (if this were fresh code).

In theory, the `schedule.repository` check in `gcf-utils` uses the URL parameter for allowed_organizations and doesn't even dispatch the events to repositories that are not in those orgs. But, the cron installer won't force overwrite the existing cron (we should probably migrate these to terraform instead).

PR #5563, opened 2024-11-19
build: simplify deploy scripts

chingor13:
service configuration including task queues are being managed by terraform instead

In a follow-up, we should be able to directly run the gcloud deploy commands to simplify the release process


chingor13:
I will remove these files in a follow-up PR that runs the deploy gcloud commands directly from cloudbuild.yaml

PR #5564, opened 2024-11-19
chore(owl-bot): add updated cross-spawn dependency

diegomarquezp:
Fixes the cross-spawn [CVE](https://github.com/advisories/GHSA-3xgq-45jj-v275) by directly adding it to package + package-lock.

### DIFF
--- a/packages/owl-bot/package-lock.json
+++ b/packages/owl-bot/package-lock.json
@@ -20,6 +20,7 @@
         "adm-zip": "^0.5.9",
         "ajv": "^8.11.0",
         "code-suggester": "^4.1.0",
+        "cross-spawn": "^7.0.6",
         "firebase-admin": "^12.4.0",
         "follow-redirects": "^1.15.2",
         "fs-extra": "^11.0.0",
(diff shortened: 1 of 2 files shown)


chingor13 (commented):
I think this dependency was already updated in #5562

diegomarquezp:
> I think this dependency was already updated in #5562

The lock file still left a few versions under `7.0.5`
https://github.com/googleapis/repo-automation-bots/blob/c64c6a64491ab48d12c3d4391a0d5f1aa806eb4e/packages/owl-bot/package-lock.json#L3486

Not sure if it downloads one copy of the dependency per parent dependency?

chingor13:
> > I think this dependency was already updated in #5562
> 
> @chingor13 The lock file still left a few versions under `7.0.5`
> 
> https://github.com/googleapis/repo-automation-bots/blob/c64c6a64491ab48d12c3d4391a0d5f1aa806eb4e/packages/owl-bot/package-lock.json#L3486
> 
> Not sure if it downloads one copy of the dependency per parent dependency?

This is the declared dependency from a direct dependency, not the version resolved.

PR #5567, opened 2024-11-19
chore: retag canary-bot-cloud-run as canary-bot

meltsufin:
Fixes b/379504250.


chingor13:
We'll actually want another `gcloud run deploy` for the GCF version's frontend, but rather than adding another invocation of the `publish-run.sh`, I can add it as I clean up usage of `publish-run.sh`

PR #5568, opened 2024-11-20
build: disable cron deployments via GCB

chingor13:
Cloud Scheduler jobs are now managed via terraform.


chingor13:
The bash lint is in a different file.

PR #5580, opened 2024-11-22
feat: allow overriding the service account for background tasks

chingor13:


meltsufin (commented):
How does adding support for additional env vars help "allow overriding the service account for background tasks"?

chingor13:
> How does adding support for additional env vars help "allow overriding the service account for background tasks"?

There is a `taskCaller` instance variable, configurable now by the `TASK_CALLER_SERVICE_ACCOUNT` env var, which is used when queuing background tasks. Adding this env var allows us to set which account is used for the background tasks via the deployment config (in terraform)

meltsufin (approved):

PR #5587, opened 2024-11-28
feat(owlbot): skip running owlbot on PRs from forks

meltsufin:
Disable running OwlBot on fork PRs even if the `owlbot:run` label is added.
`yoshi-code-bot` is the only exception.

b/379713607

**Potential future enhancements**:
- Add a comment to the PR that OwlBot did not execute
- Do not remove the `owlbot:run` label
- Consider removing `yoshi-code-bot` from allowlist

### DIFF
--- a/packages/owl-bot/src/owl-bot.ts
+++ b/packages/owl-bot/src/owl-bot.ts
@@ -616,7 +616,9 @@ const runPostProcessor = async (
 ) => {
   // If the pull request is from a fork, skip.
   if (opts.head !== opts.base) {
-    logger.info(`head ${opts.head} does not match base ${opts.base} skipping PR from fork`);
+    logger.info(
+      `head ${opts.head} does not match base ${opts.base}, skipping PR from fork`
+    );
     return;
   }
   // Fetch the .Owlbot.lock.yaml from head of PR:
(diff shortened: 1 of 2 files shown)


packages/owl-bot/test/owl-bot.ts
```diff
@@ -940,7 +939,7 @@ describe('OwlBot', () => {
         ],
         head: {
           repo: {
-            full_name: 'rennie/owl-bot-testing',
+            full_name: 'yoshi-code-bot/owl-bot-testing',
```
chingor13, line 942: Isn't this one in the allowlist and so it should trigger the build?

packages/owl-bot/src/owl-bot.ts
```diff
@@ -59,6 +59,7 @@ import {shouldIgnoreRepo} from './should-ignore-repo';
 // We use lower case organization names here, so we need to always
 // check against lower cased owner.
 const ALLOWED_ORGANIZATIONS = ['googleapis', 'googlecloudplatform'];
+const ALLOWED_FORK_OWNERS = ['yoshi-code-bot'];
```
chingor13, line 62: super minor nit but also acknowledging the prior usage of a list: we could use a `Set` here.

chingor13 (changes requested):

packages/owl-bot/src/owl-bot.ts
meltsufin: For a small list like this, it probably doesn't matter in terms of performance.

chingor13 (approved):

PR #5610, opened 2024-12-04
fix: release-trigger: upgrade from debian bullseye to bookworm

burkedavison:
See also b/382229925


meltsufin (approved):

burkedavison:
Choosing https://github.com/googleapis/repo-automation-bots/pull/5611 instead

PR #5612, opened 2024-12-04
fix: release-trigger: remove base python3 installation

burkedavison:
See also b/382229925


burkedavison:
Choosing https://github.com/googleapis/repo-automation-bots/pull/5611 instead

PR #5613, opened 2024-12-05
feat(release-trigger): Add GoogleCloudDataproc to allowlist

bsidhom:
Adding the Dataproc organization to the release-trigger allowlist.

### DIFF
--- a/packages/release-trigger/src/release-trigger.ts
+++ b/packages/release-trigger/src/release-trigger.ts
@@ -63,7 +63,11 @@ export const execFile = function (
   });
 };
 
-export const ALLOWED_ORGANIZATIONS = ['googleapis', 'GoogleCloudPlatform'];
+export const ALLOWED_ORGANIZATIONS = [
+  'googleapis',
+  'GoogleCloudDataproc',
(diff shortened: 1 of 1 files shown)


packages/release-trigger/src/release-trigger.ts
```diff
@@ -63,7 +63,7 @@ export const execFile = function (
   });
 };
 
-export const ALLOWED_ORGANIZATIONS = ['googleapis', 'GoogleCloudPlatform'];
+export const ALLOWED_ORGANIZATIONS = ['googleapis', 'GoogleCloudDataproc', 'GoogleCloudPlatform'];
```
chingor13, line 66: Looks like lint is failing.

Can you make the suggested change from below? (or run `npm run fix`)

chingor13 (changes requested):

packages/release-trigger/src/release-trigger.ts
bsidhom: Done.

chingor13 (approved):

PR #5639, opened 2024-12-19
fix(deps): update golang/x/net to 0.33.0

chingor13:
Ran:

```
go get -u
go mody tidy
```


packages/flakybot/go.mod
```diff
@@ -1,30 +1,44 @@
 module github.com/googleapis/repo-automation-bots/packages/flakybot
 
-go 1.19
+go 1.23.2
```
codyoss, line 3: Not sure how this bot is built but note this is bumping the min version of Go to 1.23.2. Do we need to update a docker file to? I personally never use the -u flag as it literally bumps everything include all indirect deps. Usually I would just `go get some/dep@latest && go mod tidy.`

codyoss (approved):

meltsufin (approved):

packages/flakybot/go.mod
chingor13: This is already using go 1.23 in the Dockerfile (https://github.com/googleapis/repo-automation-bots/blob/1aab308d37bda94e51e08956915af3be465f4453/serverless-scheduler-proxy/Dockerfile)

chingor13: We will monitor the app and be ready to revert if it breaks

PR #5669, opened 2025-02-19
deps: update gcf-utils oktokit dependencies to fix ReDoS vulnerability

ldetmer:
This is to fix ReDoS vulnerability found in octokit dependencies.  This was logic in choosing version:

Latest Probot version (13.4.3) is using @octokit/core 5.0.2 and octokit/Plugin-paginate-rest: 9.2.2, which was a limiting factor in upgrades:

@octokit/request:  upgrade to 8.4.1 for vulnerability fix + compatibility with  octokit/core v5
@octokit/request-error: 5.1.1 for vulnerability fix
@octokit/plugin-paginate-rest: 9.2.2 for vulnerability fix 

Could not upgrade @octokit/rest > 19 because of incompatibility with @octokit/plugin-paginate-rest > v9 , so need to manually set version of @octokit/core dependency in package-lock.json to 5.0.2

Additional changes had to be made:

Webhook verify throws error if payload is not include, so added logic to catch and propigate as a 500.

Tested via canary-bot: https://github.com/googleapis/repo-automation-bots/issues/5672

### DIFF
--- a/packages/gcf-utils/package.json
+++ b/packages/gcf-utils/package.json
@@ -21,11 +21,15 @@
   "dependencies": {
     "@google-cloud/run": "^1.4.0",
     "@google-cloud/secret-manager": "^5.6.0",
-    "@google-cloud/storage": "^7.12.1",
+    "@google-cloud/storage": "^7.15.1",
     "@google-cloud/tasks": "^5.5.0",
     "@octokit/auth-app": "^4.0.5",
+    "@octokit/core": "^5.0.2",
     "@octokit/graphql": "^5.0.0",
     "@octokit/plugin-enterprise-compatibility": "^2.0.3",
+    "@octokit/plugin-paginate-rest": "9.2.2",
+    "@octokit/request": "^8.4.1",
+    "@octokit/request-error": "^5.1.1",
     "@octokit/rest": "^19.0.4",
     "@octokit/types": "^9.0.0",
     "@probot/octokit-plugin-config": "^1.1.6",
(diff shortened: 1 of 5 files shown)


meltsufin (approved):

packages/gcf-utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "gcf-utils",
-  "version": "16.2.1",
+  "version": "16.2.2",
```
suztomo, line 3: Release Please maintains this line. I think this manual edit will be overridden by it (with the same value).

image

suztomo (approved):

packages/gcf-utils/test/server.ts
```diff
@@ -273,6 +277,23 @@ describe('GCFBootstrapper', () => {
...
+          headers: {
+            'x-github-delivery': '123',
+            'x-github-event': 'issues',
+          },
+          // don't throw on non-success error codes
+          validateStatus: () => {
+            return true;
+          },
+        });
+        assert.deepStrictEqual(response.status, 500);
```
suztomo, line 293: Missing payload shouldn't throw 500 error.

ldetmer: ah ok i'll put up a fix. It should log error but throw 200?

suztomo: 400 for "[Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400)" looks appropriate https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

suztomo: Just to be sure, I'm not saying all errors should be returned as "400 Bad Requests". You may have to differentiate error code based on the problems.

PR #5678, opened 2025-02-24
deps: update all octokit dependencies + transitive dependencies in probot for vulnerability fixes

ldetmer:
Probot does not have the latest vulnerability fixes for octokit, so adding transitive dependencies to override.  Additionally bumping octokit dependencies that has underlying dependencies with vulnerabilities.

Tested with npm audit and verified vulnerabilities are resolved.

Fixes [#2080](https://github.com/googleapis/repo-automation-bots/security/dependabot/2080)


packages/gcf-utils/package.json
```diff
@@ -23,14 +23,14 @@
...
-    "@octokit/plugin-paginate-rest": "9.2.2",
+    "@octokit/auth-app": "^7.1.5",
+    "@octokit/core": "^5.2.0",
+    "@octokit/graphql": "^8.2.1",
+    "@octokit/plugin-enterprise-compatibility": "^5.0.1",
+    "@octokit/plugin-paginate-rest": "^9.2.2",
     "@octokit/request": "^8.4.1",
     "@octokit/request-error": "^5.1.1",
-    "@octokit/rest": "^19.0.4",
+    "@octokit/rest": "^21.1.1",
```
chingor13, line 33: I think other libraries of ours are using v20 because newer deps that use v21 are requiring ESM. We should probably pick v20 and related major versions

PR #5679, opened 2025-02-24
feat: Release Please bot's option to tag a release commit

suztomo:
tagPullRequestNumber configuration to enable tagging on release commits.

TODO items before making this out of draft:

* [ ] Add a test to read the new configuration from a file.
* [ ] Run this bot locally (how?) to confirm it actually works to tag the release commit.

### DIFF
--- a/packages/release-please/src/release-please.ts
+++ b/packages/release-please/src/release-please.ts
@@ -498,6 +498,8 @@ const handler = (app: Probot) => {
       logger
     );
 
+    // Using a set to de-duplicate the commits
+    const commits: Set<string> = new Set();
     for (const branchConfiguration of branchConfigurations) {
       // if branch is configured for on-demand releases, then skip the push event
       // unless it looks like a release (we)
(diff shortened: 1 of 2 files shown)


packages/release-please/src/release-please.ts
```diff
@@ -520,6 +522,49 @@ const handler = (app: Probot) => {
...
+      }
+    }
+    const commitsAsList = [...commits];
+    if (commitsAsList.length !== 1) {
+      logger.info(
+        'The number of commits is not 1. Did the PR get squash-merged properly? Commits:',
+        commitsAsList
+      );
+      return;
+    } else {
```
chingor13, line 536: This `else` shouldn't be necessary as the previous conditional is a guard that stops execution.

packages/release-please/src/release-please.ts
```diff
@@ -520,6 +522,49 @@ const handler = (app: Probot) => {
...
+          pullRequestNumber = pullRequest.number;
+          break;
+        }
+      }
+      if (pullRequestNumber === undefined) {
+        logger.info(
+          `Could not find a pull request that created commit ${releaseSha}`
+        );
+        return;
+      }
```
chingor13, line 560: For readability and testability, consider breaking into a separate function that can be more easily tested outside of the context of a webhook handler.

packages/release-please/src/release-please.ts
```diff
@@ -520,6 +522,49 @@ const handler = (app: Probot) => {
         octokit,
         {logger, skipPullRequest: branchConfiguration.onDemand}
       );
+      if (branchConfiguration.tagPullRequestNumber) {
+        context.payload.commits.forEach(c => commits.add(c.id));
```
chingor13, line 526: If we don't actually need all of the commits, consider just grabbing a count (or boolean for > 1)

PR #5682, opened 2025-02-28
feat: runBranchConfiguration* functions to return CreatedReleases

suztomo:
Now the release-pleaes library's createReleases returns the list
of `CreatedRelease`s. However when we make the method call,
runner.ts does not have `octokit` instance or `owner`/`repo` names.

We may have to pass the `CreateRelease`s. However this idea gets
stuck at

```
async function runBranchConfigurationWithConfigurationHandling(
  github: GitHub,
  repoLanguage: string | null,
  repoUrl: string,
  branchConfiguration: BranchConfiguration,
  octokit: Octokit,
  options: RunBranchOptions
): Promise<CreatedRelease[] | undefined> {
  const target = `${repoUrl}---${branchConfiguration.branch}`;
  await withDatastoreLock(
    {
      lockId: RP_LOCK_ID,
      target,
      lockExpiry: RP_LOCK_DURATION_MS,
      lockAcquireTimeout: RP_LOCK_ACQUIRE_TIMEOUT_MS,
    },
    async () => {
      return await runBranchConfigurationWithConfigurationHandlingWithoutLock(
        github,
        repoLanguage,
```

PR #5683, opened 2025-02-28
feat: tagPullRequestNumber option to tag release SHA

suztomo:
(Compared to the previous attempt https://github.com/googleapis/repo-automation-bots/pull/5682, this version creates the ref in runBranchConfiguration.)

With the following lines in release-please.yml file, the bot now can tag the release 

```
handleGHRelease: true
tagPullRequestNumber: true
```

Test case is in 'should tag pull request number if configured'.

### DIFF
--- a/packages/release-please/src/release-please.ts
+++ b/packages/release-please/src/release-please.ts
@@ -403,13 +403,12 @@ async function runBranchConfiguration(
         for (const release of releases) {
           prNumberToSha.set(release.prNumber, release.sha);
         }
-        for (const prNumber of prNumberToSha.keys()) {
+        for (const [prNumber, sha] of prNumberToSha.entries()) {
           // A Git tag is a ref. A lightweight tag only requires this
           // create references API call, rather than a tag objec
           // (/repos/{owner}/{repo}/git/tags).
           // https://docs.github.com/en/rest/git/tags?apiVersion=2022-11-28#create-a-tag-object
           // https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#create-a-reference
-          const sha = prNumberToSha.get(prNumber)!;
           const tagRefName = `refs/tags/release-please-${prNumber}`;
           logger.info(`Creating ${tagRefName} pointing to ${sha}`);
           const tagResponse = await octokit.git.createRef({
--- a/packages/release-please/src/release-please.ts
+++ b/packages/release-please/src/release-please.ts
@@ -399,11 +399,11 @@ async function runBranchConfiguration(
         // It's possible for manifest.createReleases() to create
         // releases for multiple pull requests. Find the unique
         // pair of pull request numbers and their commit SHAs.
-        var prNumberToSha = new Map<number, string>();
+        const prNumberToSha = new Map<number, string>();
         for (const release of releases) {
           prNumberToSha.set(release.prNumber, release.sha);
         }
-        for (const prNumber of (prNumberToSha.keys())) {
+        for (const prNumber of prNumberToSha.keys()) {
           const sha = prNumberToSha.get(prNumber)!;
           const tagRefName = `refs/tags/release-please-${prNumber}`;
           logger.info(`Creating ${tagRefName} pointing to ${sha}`);
--- a/packages/release-please/src/release-please.ts
+++ b/packages/release-please/src/release-please.ts
@@ -392,19 +392,29 @@ async function runBranchConfiguration(
         // version
         manifest = null;
       }
-      if (releases.length > 0 && branchConfiguration.tagPullRequestNumber) {
-        // All releases from one Git push event share the same pull request
-        // number and commit SHA. We just use the first one.
-        const firstRelease = releases[0];
-        const tagRefName = `refs/tags/release-please-${firstRelease.prNumber}`;
-        logger.info(`Creating ${tagRefName} pointing to ${firstRelease.sha}`);
-        const tagResponse = await octokit.git.createRef({
-          owner: github.repository.owner,
-          repo: github.repository.repo,
-          ref: tagRefName,
-          sha: firstRelease.sha,
-        });
-        logger.info('Got tag response: ', tagResponse);
+      if (branchConfiguration.tagPullRequestNumber) {
+        // Record the pull request number to the commit as
+        // a tag.
+
+        // It's possible for manifest.createReleases() to create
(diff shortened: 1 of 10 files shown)


suztomo:
I need to fix tests.

packages/release-please/src/release-please.ts
```diff
@@ -383,13 +385,27 @@ async function runBranchConfiguration(
...
+      const releases = await Runner.createReleases(manifest);
+      logger.info(`Created ${releases.length} releases`);
+      if (releases.length > 0) {
         // we created a release, reload config which may include the latest
         // version
         manifest = null;
       }
+      if (releases.length > 0 && branchConfiguration.tagPullRequestNumber) {
+        // All releases from one Git push event share the same pull request
+        // number and commit SHA. We just use the first one.
```
chingor13, line 397: This may not be true.

Release-please looks for all pending releases and will try and tag them all (not necessarily only the last one). We probably want to find all the unique release PR/commit combinations and tag each of those

suztomo: Thank you.

packages/release-please/test/release-please.ts
```diff
@@ -696,7 +708,7 @@ describe('ReleasePleaseBot', () => {
 
         sinon.assert.calledOnce(createPullRequestsStub);
         sinon.assert.calledOnce(createReleasesStub);
-        sinon.assert.calledOnce(fromManifestStub);
+        sinon.assert.calledTwice(fromManifestStub);
```
suztomo, line 711: @chingor13 This is now called twice. Can you help me understand why?

suztomo:
The tests succeeded in GitHub but it fails in my Linux:

```
  2) release-please bot
       config schema check on PRs
         should create a failing status check for broken custom manifest configs:
     Uncaught Error: Different value of snapshot "release-please bot config schema check on PRs should create a failing status check for broken custom manifest configs"
{
  output: {
    text: "the given config is not valid JSON 😱 \nUnexpected token , in JSON at position 59" => "the given config is not valid JSON 😱 \nUnexpected token ',', ...\"ing-comma\",\n  }\n}\n\" is not valid JSON"
  }
}
      at /usr/local/google/home/suztomo/repo-automation-bots/packages/release-please/node_modules/snap-shot-core/src/file-system.js:263:11
      at InternalConstructor.orElse (node_modules/folktale/result/result.js:187:14)
      at raiseIfDifferent (node_modules/snap-shot-core/src/file-system.js:254:10)
      at setOrCheckValue (node_modules/snap-shot-core/src/index.js:367:7)
      at core (node_modules/snap-shot-core/src/index.js:388:12)
      at snapshot (node_modules/snap-shot-it/src/index.js:293:18)
      at Object.<anonymous> (build/test/compatibility-test.js:221:44)
      at matchBody (node_modules/nock/lib/match_body.js:42:17)
```

suztomo:
> The tests succeeded in GitHub but it fails in my Linux:

It's the behavior change in `JSON.parse` in bot-config-utils.js:

Screenshot 2025-03-02 at 5 26 28 PM

The CI is set up with Node version v18. I have v22.

packages/release-please/test/release-please.ts
suztomo: 1st call:

Screenshot 2025-03-02 at 5 46 32 PM

2nd call:

Screenshot 2025-03-02 at 5 46 43 PM

suztomo: [The source code comment](https://github.com/suztomo/repo-automation-bots/blob/8624a2d4d429c204b006603d376cb03f6489387b/packages/release-please/src/release-please.ts#L391) says when we create a release, we reload the manifest configuration.

packages/release-please/src/release-please.ts
suztomo: Fixed. Added a test case 'should tag each SHA for pull requests number if configured'.

suztomo:
All tests passed.

packages/release-please/src/release-please.ts
```diff
@@ -383,13 +385,42 @@ async function runBranchConfiguration(
...
+        // a tag.
+
+        // It's possible for manifest.createReleases() to create
+        // releases for multiple pull requests. Find the unique
+        // pair of pull request numbers and their commit SHAs.
+        const prNumberToSha = new Map<number, string>();
+        for (const release of releases) {
+          prNumberToSha.set(release.prNumber, release.sha);
+        }
+        for (const prNumber of prNumberToSha.keys()) {
```
chingor13, line 406: For iterating over a map, you can do something like: 
`for (let [key, value] of  map.entries()) {for (let [key, value] of  map.entries()) {`

packages/release-please/src/release-please.ts
```diff
@@ -383,13 +385,42 @@ async function runBranchConfiguration(
...
+          // A Git tag is a ref. A lightweight tag only requires this
+          // create references API call, rather than a tag objec
+          // (/repos/{owner}/{repo}/git/tags).
+          // https://docs.github.com/en/rest/git/tags?apiVersion=2022-11-28#create-a-tag-object
+          // https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#create-a-reference
+          const sha = prNumberToSha.get(prNumber)!;
+          const tagRefName = `refs/tags/release-please-${prNumber}`;
+          logger.info(`Creating ${tagRefName} pointing to ${sha}`);
+          const tagResponse = await octokit.git.createRef({
```
chingor13, line 415: minor nit that could make the tests cleaner would be to move this into a function like `createTag` where we wrap the octokit call.

Rather than testing the HTTP mock in the event handler test, you can assert that `createTag` is called N times.

Then separately, you could test the implementation of the `createTag` where we mock the HTTP call.

chingor13 (approved):

packages/release-please/src/release-please.ts
suztomo: Updated. Thanks.

packages/release-please/src/release-please.ts
suztomo: (Thinking about which class should have `createTag` method) Is that the `Runner` for? Now it makes sense.

chingor13 (approved):

packages/release-please/src/release-please.ts
suztomo: Done.

PR #2079, opened 2021-06-16
feat(merge-on-green): start syncing labels

tmatsuo:
part of #1673

PR #5689, opened 2025-03-10
fix: octokit request/request-error sec vuln fixes

ldetmer:
update all probot/octokit dependencies to use request 8.4.1 and request-error 5.1.1 

We can not upgrade to latest octokit version because of breaking migration to ESM that probot does not support.  

Tested via canary bot: https://github.com/googleapis/repo-automation-bots/issues/5691

Fixes #2079


packages/gcf-utils/package.json
```diff
@@ -76,6 +71,25 @@
...
+    "@octokit/core": {
+      "@octokit/graphql": "^7.1.1",
+      "@octokit/request": "8.4.1",
+      "@octokit/request-error": "5.1.1"
+    },
+    "@octokit/oauth-methods": {
+      "@octokit/request": "8.4.1",
+      "@octokit/request-error": "5.1.1"
+    }
+  },
```
chingor13, line 92: > We can not upgrade to latest octokit version because of breaking migration to ESM that probot does not support. So for each octokit library, we need to override to use 8.4.1/5.1.1 because these older versions were not upgraded with vuln fix

We should only need these if they are overriding major versions. Otherwise, the user of this library will be resolving dependencies (and should be picking up the patched version).

ldetmer: ok I removed, but package-lock is still showing older versions.  However if I run npm ls @octokit/request its only showing the correct version.  Is that good enough?

suztomo: > Is that good enough

Try "npm audit".

chingor13 (approved):

packages/gcf-utils/package.json
ldetmer: confirmed npm audit does not show issues

PR #5707, opened 2025-03-12
deps(header-checker-lint): Updating gcf-utils and bot-config-utils

suztomo:
b/402728578

### DIFF
--- a/packages/header-checker-lint/package-lock.json
+++ b/packages/header-checker-lint/package-lock.json
@@ -9,8 +9,8 @@
       "version": "1.2.0",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


chingor13:
Octokit v20 now has an option for setting the `fetch` transport which is native on node18. We do not use the native fetch (experimental in node 18) because nock cannot mock those requests.

We'll need to update the tests to tell Octokit to use node-fetch (a polyfill library) like in https://github.com/googleapis/repo-automation-bots/pull/5699/files#diff-e91e39ef5e2362b5d4afe8ed8405bb9554ef07b12f718316e316404fc7c4da5a

suztomo:
@chingor13 Thank you

chingor13 (approved):

PR #5763, opened 2025-05-13
feat: use secure image published via AR Exit Gate

jinseopkim0:
b/407529057#comment37


meltsufin (commented):
Is there any way to test the image before we cutover?

PR #5781, opened 2025-06-05
build: Use MOSS-approved base image for owl-bot-backend

meltsufin:
b/420030444

### DIFF
--- a/packages/owl-bot/Dockerfile.backend
+++ b/packages/owl-bot/Dockerfile.backend
@@ -14,8 +14,8 @@
 
 # Use a multi-stage docker build to limit production dependencies.
 
-# Stage 1: Build
-FROM marketplace.gcr.io/google/debian12:latest AS BUILD
+# Stage 0: Node.js Base Image
+FROM marketplace.gcr.io/google/debian12:latest AS NODE_BASE
 
 # Install Node.js v18 and npm.
 RUN apt-get update && \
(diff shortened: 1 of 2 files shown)


packages/owl-bot/Dockerfile.backend
```diff
@@ -14,9 +14,14 @@
...
 # Use a multi-stage docker build to limit production dependencies.
 
-# Use the official lightweight Node.js 14 image.
-# https://hub.docker.com/_/node
-FROM node:18.20.5-slim AS BUILD
+# Stage 1: Build
+FROM marketplace.gcr.io/google/debian12:latest AS BUILD
+
+# Install Node.js v18 and npm.
+RUN apt-get update && \
```
chingor13, line 21: To speed this up, we should be able to create an intermediate image so we only have to install nodejs once.

meltsufin: Do you mean publishing a custom base image to AR and using it everywhere? We might want to do it, but probably in a separate PR.

chingor13:
Just building the image is a start. Then we know it will at least run.

We can also add container tests like we do for our testing images: https://github.com/googleapis/testing-infra-docker/blob/b7c76e8d3d7a5112c27c0281be6aeb7416bdbe0b/php/cloudbuild-test.yaml#L30-L34

packages/owl-bot/Dockerfile.backend
chingor13: This can be done with the same Dockerfile using a multi-stage build.

```
FROM marketplace.gcr.io/google/debian12:latest AS NODE_BASE
RUN <install node>

FROM NODE_BASE AS BUILD
RUN npm i && npm run compile

# Final image
FROM NODE_BASE
RUN <do stuff>
```

chingor13 (approved):

packages/owl-bot/Dockerfile.backend
meltsufin: I like that idea. I'll update.

chingor13 (approved):

PR #5934, opened 2025-10-07
fix(deps): upgrade pino to fix fast-redact vulnerabilty

meltsufin:
https://github.com/googleapis/repo-automation-bots/security/dependabot/2366


chingor13 (changes requested):
The change LGTM, but this should be a `fix` type so that we can release a new version of gcf-utils and update the other bots

chingor13 (approved):

PR #6115, opened 2026-04-20
feat: add retry plugin to octokit instance passed to release-please

chingor13:
Attempt to fix a seemingly transient errors

### DIFF
--- a/packages/release-please/src/auth.ts
+++ b/packages/release-please/src/auth.ts
@@ -1,0 +1,60 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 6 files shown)


packages/release-please/src/auth.ts
```diff
@@ -0,0 +1,60 @@
...
+import {retry} from '@octokit/plugin-retry';
+
+const OctokitWithRetry = Octokit.plugin(retry);
+/**
+ * Build Octokit instance with the given installation ID.
+ * If installation ID is not provided, authenticate as a bot.
+ * @param installationId installation ID of a GitHub App.
+ * @returns Octokit instance.
+ */
+export async function getAuthenticatedOctokit(
```
chingor13, line 29: If this works well, we can upstream this in gcf-utils. However, we want to more quickly test this.

suztomo (approved):

PR #6135, opened 2026-05-07
feat(bazelbot): remove building Go and Python targets

codyoss:
Both Go and Python have moved to Librarian for the generation so we no longer need to build the libraries for googleapis-gen, mostly. The exception here is [python-aiplatform](https://github.com/googleapis/python-aiplatform/blob/main/.github/.OwlBot.yaml) which still uses owlbot.

This should significantly speed up the builds of bazelbot and make it less likely to flake.

Internal Bug: b/510765130


JoeWang1127:
Could we remove `java` from this build as well? Java uses hermetic build several years ago.

noahdietz (approved):

codyoss:
@JoeWang1127 Sure, I will create a followup PR and assign it to you.

PR #6191, opened 2026-05-14
fix: increase lock timeouts for large monorepos

codyoss:
Tasks on large monorepos (e.g., google-cloud-node) can take 4+ minutes due to size and the number of GitHub API requests made. The previous 60s limit caused active locks to prematurely auto-expire, resulting in concurrent worker overlap and jobs running unlocked for a period of time.

### DIFF
--- a/packages/release-please/src/release-please.ts
+++ b/packages/release-please/src/release-please.ts
@@ -286,7 +286,7 @@ async function buildManifest(
 }
 
 const RP_LOCK_ID = 'release-please';
-const RP_LOCK_DURATION_MS = 60 * 1000;
+const RP_LOCK_DURATION_MS = 5 * 60 * 1000;
 const RP_LOCK_ACQUIRE_TIMEOUT_MS = 120 * 1000;
 interface RunBranchOptions {
   logger?: GCFLogger;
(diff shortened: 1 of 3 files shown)


packages/release-please/src/release-please.ts
```diff
@@ -286,7 +286,7 @@ async function buildManifest(
 }
 
 const RP_LOCK_ID = 'release-please';
-const RP_LOCK_DURATION_MS = 60 * 1000;
+const RP_LOCK_DURATION_MS = 5 * 60 * 1000;
```
suztomo: Can you use 10 minutes?

suztomo (approved):
Would you fix the check failure?

packages/release-please/src/release-please.ts
codyoss: Sure.

suztomo:
Release Please needs to use the new version of datastore-lock; otherwise the old datastore-lock may throw an error for an invalid duration.

suztomo:
> the old datastore-lock may throw an error for an invalid duration.

Added a temporary workaround in 521458652#comment7.

PR #6245, opened 2026-06-11
deps: bump @grpc/grpc-js to 1.14.4

sofisl:
Bump @grpc/grpc-js to version 1.14.4


quirogas (approved):
Nit: if there is a tracking issue could you include it in the description please?

PR #6248, opened 2026-06-12
fix(bazel-bot): replace gh and jq dependencies

sofisl:
Fixes 
- https://b.corp.google.com/issues/521406587
- https://b.corp.google.com/issues/521484096
- https://b.corp.google.com/issues/521395073

### DIFF
--- a/packages/bazel-bot/docker-image/docker-main.sh
+++ b/packages/bazel-bot/docker-image/docker-main.sh
@@ -39,7 +39,7 @@ RESPONSE=$(curl -X POST \
     -H "Authorization: Bearer $JWT" \
     -H "Accept: application/vnd.github.v3+json" \
     https://api.github.com/app/installations/$GITHUB_APP_INSTALLATION_ID/access_tokens)
-GITHUB_TOKEN=$(echo "$RESPONSE" | jq -r .token)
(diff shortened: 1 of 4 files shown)


packages/bazel-bot/docker-image/docker-main.sh
```diff
@@ -39,7 +39,7 @@ RESPONSE=$(curl -X POST \
     -H "Authorization: Bearer $JWT" \
     -H "Accept: application/vnd.github.v3+json" \
     https://api.github.com/app/installations/$GITHUB_APP_INSTALLATION_ID/access_tokens)
-GITHUB_TOKEN=$(echo "$RESPONSE" | jq -r .token)
+GITHUB_TOKEN=$(echo "$RESPONSE" | python3 -c "import sys, json; print(json.load(sys.stdin)['token'])")
```
codyoss, line 42: This is clever. It looks like we are using a base image that must already include python so this seems fine. One alternative would be do so this completely with bash instead of an inlined script: `GITHUB_TOKEN=$(echo "$RESPONSE" | grep -o '"token": *"[^"]*"' | cut -d'"' -f4)`

codyoss (approved):
LGTM, gave an alternative but I am good with either approach.

sofisl:
Actually, I don't think this would solve most of the current vulnerabilities... I can close if you want, or we can give it a try?

sofisl:
actually yeah, going to close for the time being but we can always reopen if it's jq-specific

PR #6287, opened 2026-06-30
revert: "pin node version (#6254)"

codyoss:
This reverts commit ca1e124d9c867896357de4010a2f35b6e0f2d5cb. A new node version was published with a patch for the bug we were running into.

Internal Bug: b/525453379

### DIFF
--- a/packages/gcf-utils/Dockerfile
+++ b/packages/gcf-utils/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM node:22.22.3
+FROM node:22
 
 USER root
 
--- a/packages/owl-bot/Dockerfile.backend
+++ b/packages/owl-bot/Dockerfile.backend
@@ -15,20 +15,21 @@
 # Use a multi-stage docker build to limit production dependencies.
 
 # Stage 0: Node.js Base Image
-FROM marketplace.gcr.io/google/debian12:latest AS node_base
+FROM marketplace.gcr.io/google/debian12:latest AS NODE_BASE
 
-# Install Node.js v22.22.3 and npm.
+# Install Node.js v18 and npm.
 RUN apt-get update && \
     apt-get upgrade -y && \
-    apt-get install -y curl xz-utils && \
-    curl -fsSL https://nodejs.org/dist/v22.22.3/node-v22.22.3-linux-x64.tar.xz | tar -xJf - --strip-components=1 -C /usr/local && \
+    apt-get install -y curl && \
+    curl -fsSL https://deb.nodesource.com/setup_22.x | bash - && \
+    apt-get install -y nodejs && \
     rm -rf /var/lib/apt/lists/*
 
 # Remove unnecessary cross-spawn from npm to resolve CVE-2024-21538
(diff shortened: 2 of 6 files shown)


packages/gcf-utils/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM node:22.22.3
+FROM node:22
```
g-husam, line 15: curious why we don't want to pin specific versions? It would prevent issues like this from popping up in the future. Is it considered too toilsome to maintain?

packages/owl-bot/Dockerfile.backend
```diff
@@ -15,20 +15,21 @@
 # Use a multi-stage docker build to limit production dependencies.
 
 # Stage 0: Node.js Base Image
-FROM marketplace.gcr.io/google/debian12:latest AS node_base
+FROM marketplace.gcr.io/google/debian12:latest AS NODE_BASE
 
-# Install Node.js v22.22.3 and npm.
+# Install Node.js v18 and npm.
```
g-husam, line 20: nit: this comment is outdated (can be addressed separately to keep this revert clean - same for other Dockerfiles)

packages/owl-bot/Dockerfile.frontend
```diff
@@ -15,20 +15,21 @@
 # Use a multi-stage docker build to limit production dependencies.
 
 # Stage 0: Node.js Base Image
-FROM marketplace.gcr.io/google/debian12:latest AS node_base
+FROM marketplace.gcr.io/google/debian12:latest AS NODE_BASE
 
-# Install Node.js v22.22.3 and npm.
+# Install Node.js v18 and npm.
```
g-husam, line 20: same here and in the other Dockerfiles

g-husam (commented):
nit: would be nice if the version of node could be centralized across Dockerfiles. Not sure how these dockerfiles are used, but if only through CI can maybe make the version an arg to each dockerfile, and define it once in the CI script

packages/gcf-utils/Dockerfile
codyoss: We could but like you hint at it would make more dependency PRs here as well as internal bugs when things like CVEs get patched and we have not updated yet. This node issue that we worked around was the first issue I have seen on the team where we were outright broken by the language ecosystem. In general it feels pretty safe/stable to float the major here.

packages/owl-bot/Dockerfile.backend
codyoss: You are correct. Yeah if possible I would like to keep this a clean revert and followup in a PR to address this feedback.

packages/owl-bot/Dockerfile.frontend
codyoss: same as above, will address in a followup.

codyoss:
> nit: would be nice if the version of node could be centralized across Dockerfiles. Not sure how these dockerfiles are used, but if only through CI can maybe make the version an arg to each dockerfile, and define it once in the CI script

I agree. Although, I do see most things in this repo getting turned-down and deleted by EoY -- at least that is my hope. A lot of toil in this repo :)

zhumin8 (approved):

PR #6332, opened 2026-07-27
deps: upgrade brace-expansion to 5.0.8 (CVE-2026-14257)

zhumin8:
Upgrades brace-expansion dependency to the secure version 5.0.8 across all packages in the monorepo to address CVE-2026-14257.

To resolve compatibility issues with brace-expansion 5.0.8 (which uses named exports instead of default function export):
    - Upgraded c8 to 12.0.0 in all packages to pull in compatible minimatch 10.x.
    - Upgraded glob to 13.0.6 in owl-bot.
    - Updated glob usage in owl-bot to use globSync and filtered out '.' from results.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1074,19 +1074,6 @@
         "url": "https://github.com/sponsors/isaacs"
       }
     },
-    "node_modules/glob/node_modules/minimatch": {
-      "version": "3.1.5",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.5.tgz",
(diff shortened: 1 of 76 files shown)


zhumin8:
The upgrade of glob to 13.0.6 in owl-bot is necessary because of a chain of dependency incompatibilities triggered by the security fix for brace-expansion.
  We need to force brace-expansion to 5.0.8 to fix the vulnerability. However, brace-expansion 5.0.8 changed its packaging format to use named exports instead of a default function export. We need use a version of minimatch that is compatible with brace-expansion 5.x (minimatch@10.x).  Since glob manages its own dependency on minimatch, we had to upgrade glob to a version that uses minimatch@10.x.

 c8 is used to measure code coverage when running tests. We need use a version of c8 that uses test-exclude and minimatch versions compatible with brace-expansion 5.x.

sofisl (approved):

PR #6354, opened 2026-08-24
feat(git-file-utils): fallback to non-recursive tree traversal on HTTP 422 and ignore submodules

codyoss:
### Problem

When traversing Git trees in repositories containing Git submodules (gitlink entries with `mode 160000` / `type commit`), GitHub's Git Trees API rejects recursive queries (`?recursive=true`) with **HTTP 422 Unprocessable Entity** because submodule commit objects reside in external repositories.

### Solution

1. **HTTP 422 Fallback**: In `BranchFileCache.getTree(sha)`, catch HTTP 422 from recursive tree queries and fall back to non-recursive tree traversal (`fetchTree(sha, false)`).
2. **Skip Submodules in Tree Traversal**: In `BranchFileCache.treeEntryIterator()`, ignore entries with `mode === '160000'` or `type === 'commit'` so they are not yielded or enqueued for child tree traversal.
3. **Guard Direct File Reads**: In `BranchFileCache.fetchFileContents()`, throw `FileNotFoundError` if a path points to a submodule entry.

I verified locally that this should fix the issue we saw last Friday by running release please unpatched and also by locally patching in these changes, to make sure we no longer fail with 422s when dealing with git submodules.

### DIFF
--- a/packages/git-file-utils/src/git-file-utils.ts
+++ b/packages/git-file-utils/src/git-file-utils.ts
@@ -325,7 +325,7 @@ export class BranchFileCache {
     const treeEntries = await this.getFullTree();
     if (treeEntries) {
       const found = treeEntries.find(entry => entry.path === path);
-      if (found?.sha) {
+      if (found?.sha && !isGitSubmodule(found)) {
         return await this.fetchContents(found.sha, found);
       }
       throw new FileNotFoundError(path);
@@ -338,7 +338,7 @@ export class BranchFileCache {
     for (const part of parts) {
       const {tree} = await this.getTree(treeSha);
       found = tree.find(item => item.path === part);
-      if (!found?.sha) {
+      if (!found?.sha || isGitSubmodule(found)) {
         throw new FileNotFoundError(path);
       }
       treeSha = found.sha;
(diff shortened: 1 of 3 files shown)


codyoss:
This change will need to be merged and released, and then release please will need to depend on this to fully fix the issue we saw last week in google-cloud-python.

packages/git-file-utils/src/git-file-utils.ts
```diff
@@ -325,7 +325,7 @@ export class BranchFileCache {
     const treeEntries = await this.getFullTree();
     if (treeEntries) {
       const found = treeEntries.find(entry => entry.path === path);
-      if (found?.sha) {
+      if (found?.sha && found.mode !== '160000' && found.type !== 'commit') {
```
hj690: nit: maybe create a helper function for this? something like `isGitSubmodule(treeEntry: TreeEntry): boolean`

hj690 (approved):

packages/git-file-utils/src/git-file-utils.ts
codyoss: Done! Added `isGitSubmodule` helper function.

PR #6359, opened 2026-08-27
fix(release-please): inherit top-level config options in branch overrides

codyoss:
Merge top-level configuration defaults into each branch configuration so options like `handleGHRelease`, `manifest`, and `releaseType` are not dropped when a branch defines specific overrides.

Also prevents root-only properties from leaking into branch configs and ensures inherited monorepo paths do not trigger duplicate executions.

Fixes: b/553522648


codyoss:
cc @shollyman

hj690 (approved):

